PR #2, opened 2015-04-23
Added possibility to send status 'opened'  for a message back to onesignal backend from client source code

polok:
Hi,

We override our notification handler for GCM  messages because of our requirements. However we would like to have a possibility to send a callback to your server from our code that the message was opened. We can't invoke this method "public static void handleNotificationOpened(Context inContext, Bundle data)" as it forces us to open the app what we don't want to do here".

If you don't want to mark this method as public maybe you would like to change visibility of this method:
"private static String getSavedUserId(Context inContext)". Then we could do a PUT callback on our own but we have to know the player_id.

Thanks in advance.

### DIFF
--- a/OneSignalSDK/src/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/src/com/onesignal/OneSignal.java
@@ -630,7 +630,7 @@ public class OneSignal {
          runNotificationOpenedCallback(data, false, false);
    }
 
-   private static void sendNotificationOpened(Context inContext, Bundle data) {
+   public static void sendNotificationOpened(Context inContext, Bundle data) {
       try {
          JSONObject customJson = new JSONObject(data.getString("custom"));
          String notificationId = customJson.getString("i");


jkasten2:
Hello,

Could you explain in more detail about the overriding your doing? Are you still using our GcmBroadcastReceiver to show an Android notification? Are you using our background data option currently?

I don't think either method is something we want to directly make public. Let us know more about your use cases and we can make sure our SDK can work with them in a future update.

Thanks

polok:
Hi,

Yes, sure.

We are overriding your GcmBroadcastReceiver where we build our custom notifications look based on data from the push notification. Also, we put this message into our local storage what is critical for us and we want to notify your server that the message was 'opened' what in our case will mean that the message was saved in our local storage - this is very, very critical for our use case as we want to be sure that the message was saved in user device. Going further, this was a reason why we have overrided your implementation of IntentService. As it contains this part

 if (isGcmMessage(intent))
         NotificationBundleProcessor.Process(this, extras, com.onesignal.NotificationOpenedActivity.class);

And we want to have our own way to handle notifications what I explained at the beginning. We always show pushes in Android notifcation's center.
This is a reason why we can't use your implementaiton of showing/handling notifications as we do here more custom things.

To sum up:
We are going to use your tool as a critical process in our workflow and it's for us super important to make sure that the message was saved under user device.

Please, let us know what do you think.

jkasten2:
Thanks for your explanation, let us know what custom things your app needs to change on the notification before showing it to the notification area on Android. We can look into adding the features your app requires in a future update.

We use run:
```java
if (isGcmMessage(intent))
  NotificationBundleProcessor.Process(this, extras, com.onesignal.NotificationOpenedActivity.class);
```

to make sure it is a valid GCM message, it is from our server, and it isn't a duplicated (from canonical ids). As long as this passes the notification is always displayed if it has a message body (alert key).

If the look of the notification is ok and you just need to do custom things with the data on the notification you can use our [Background Data](http://documentation.onesignal.com/v2.0/docs/android-notification-customizations#background-data-silent-notifications) receiver.  Make sure to also the enable background data on the notification when you send it. You may also leave the body of the notification blank and it will be a silent notification.

We would like to support the customization you need over opening up the calling of `handleNotificationOpened(...)` to be public. However in the mean time you can call this with reflection if you need too. Let us know if you need more details on using reflection to call a private method.

Lastly you can also use [IdsAvailable](http://documentation.onesignal.com/v2.0/docs/android-sdk-api#IdsAvailable) to get userId to send your own PUT call.

Thanks.

polok:
Hi,

Regarding notification changes we send some json data under 'custom' field. We extract this data and store it locally. Some of it goes to body of our message. Yes, we could send this body data under 'alert' key but we don't want to change our backend side. 

I'm glad to hear that you are going to open this method. This is brilliant news for us. For now we will go with reflection or with IdsAvailable method. Btw. Thanks for pointing it out.

The last thing, are you able to say when you are going to release new version of the SDK ?

jkasten2:
We would like to understand the customization your doing to the notification instead of opening up the `private static void sendNotificationOpened(Context inContext, Bundle data)` method. The method is designed to be internal to the SDK and there might be future change to it that could break your app's code. You can use reflection to call this method but you should review the code changes each time you update to a new version of our SDK.

Are you using or REST API to send notifications or are you using GCM directly? Can you change your backend to use our content field from our API? That will put the message data into the alert bundle field.

Thanks.

PR #31, opened 2016-03-10
Add a JSON attribute to send the OneSignal ping and attributes w/o notification

Fleker:
I want to generate my own notifications by heavily taking advantage of Android's notification APIs without having to receive two OneSignal notifications. However, I don't want to remove features that would allow iOS users to receive notifications as expected.

So, I created the optional attribute `generateAndroidNotification`. If this is set to false, OneSignal will still receive a GCM message and call the custom background receiver. However, no Android notification will be automatically generated. At the same time, this will not have any effect for current users. If `generateAndroidNotification` is set to true or is not set at all, OneSignal will work as currently implemented.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/one-signal/onesignal-android-sdk/31)


Fleker:
It looks like some of the local build files caused the build to fail. I'm going to remove those and fix the PR.

jkasten2:
@Fleker Thanks for your pull request, we are considering a new receiver pattern to consolidate the current SDK options which will cover the use case of your pull request as well. You can add feedback to the following pull request.
https://github.com/one-signal/OneSignal-Android-SDK/pull/32

Closing this PR so discussion isn't split up.

PR #32, opened 2016-03-15
Added method to disable showing notifications

natangr:
I want to prevent a notification that is sent with content from being displayed because I want to handle it in the background receiver. I've added a method that allows disabling the notifications that have content from being displayed. This is done via a flag in shared preferences.

---
This change is [Reviewable](https://reviewable.io/reviews/one-signal/onesignal-android-sdk/32)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
@@ -56,6 +56,8 @@ public class NotificationBundleProcessor {
 
          BackgroundBroadcaster.Invoke(context, bundle, isActive);
 
+         if (OneSignal.getPreventDisplayingNotificationEnabled(context)) return;
+
          if (!bundle.containsKey("alert") || bundle.getString("alert") == null || bundle.getString("alert").equals(""))
             return;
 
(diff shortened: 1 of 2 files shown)


dlazaro66:
:+1: Please merge this changes since it will add more flexibility to the SDK

dlazaro66:
Is there any update on this issue?

jkasten2:
@dlazaro66 @natangr We are looking at going down the route of using a overrider `BroadcastReceiver` which will allow the app developer to add logic to modify the notification properties before , decide if some notification should be skipped, or just handle the notification as a background event. This will help consolidate and simplify the current options into a single pattern as well as allow more customization that we have been getting requests for.

I'll look into rolling out a simple check to omit displaying a notification when presents of a new overrider receiver is in the `AndroidManifest.xml` to start in the next minor release. In a later release we will provide options to display the SDK notification if needed through a method with options to override parts of it.

Fleker:
Yes this would be a great addition to the API so developers can take advantage of niche platform-specific features.

Fleker:
I ended up doing the same thing

jkasten2:
Just added a `NotificationExtenderService` to allow this.
See [Background Data and Notification Overriding](https://documentation.onesignal.com/v2.0/docs/android-notification-customizations#background-data-and-notification-overriding) on how to set this up and the features it contains.

Feel free to reply and reopen the issue if you run into any problems with it.

PR #350, opened 2017-10-26
cleaning up the example app to remove v11 references

sul279:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/350)


jkasten2:
@sul279 Thanks for the pull. We don't need the v11 folders any more however the icons themselves need to be moved into the standard drawable folders. Since `-v11` in the folder name means for API 11+ devices

jkasten2:
Correctly removing deprecated icons in PR #463

Issue #375, opened 2017-11-20
Wrong intent type using OS deeplinks in launch URL on Android platforms


pmaroco:
Hello! 

I was creating a notification containing an OS level deeplink using the Launch URL field as described [here](https://documentation.onesignal.com/v3.0/docs/links#section-os-level-deep-links), but it's creating a VIEW intent instead of a MAIN intent after opening the notification.

That said, if the app is opened in the background and a notification containing a deeplink is opened, a second instance of the app is opened and killed after performing some actions. Further investigation, this behavior only happens on Android, since deeplinks in iOS are working fine.

jkasten2:
@pmaroco I'll follow up with you on the PR you linked to above.

PR #376, opened 2017-11-20
Issue #375 - Added logic to create propper intent based on URL type

pmaroco:
Solution for to Issue #375

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/376)


jkasten2:
@pmaroco I believe after this change this would launch the main Activity of the target app for app deep links. This might not be the correct Activity for a number of apps.

Is this change to fix an issue where you have multiple instances of your main activity? I believe this would be due to the `Intent.FLAG_ACTIVITY_NEW_TASK` flag. However this is needed as `startActivity` requires this when using an application content.

PR #418, opened 2018-01-25
Modified OneSignal.java to fit the marmalade SDK

cookksss:
- method stubs for onPause/onResume
- implemented workarounds for inFocusDisplaying
- refactored OpenedHandler to fit the marmalade method signature

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/418)


jkasten2:
@cookksss We can't merge this commit directly into master as is. It changes public methods on the main target which would break all 3.X releases.

It is possible to add another gradle target and a copy of this class could be added without this limitation. However we don't plan to continuously maintain this 2.x to 3.x bridge for the Marmalade SDK.

I would be open to an update to [OneSignal-Marmalade-SDK](https://github.com/OneSignal/OneSignal-Marmalade-SDK) instead to use the new 3.X version of this SDK however.

PR #479, opened 2018-03-21
Fixes OneSignalPrefs reads that may use old values

jkasten2:
* OneSignalPrefs uses a different thread to write,
   if a read is done before the write thread finishes the read would use old value
  - This may have resulted in some vary rare race condition issues
  - Reads are now done from the pending write cache now to prevent any possible issue
  - This should improve test stability as we did not wait for writes between tests before
* OneSignalPrefs now uses a HandlerThread instead of an overkill ScheduledThreadPoolExecutor
  - This results in only using 1 thread instead of up to 10 as it was configured
  - Writes are now buffer with a short 200ms delay to prevent extra write I/O

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/479)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
@@ -1,26 +1,47 @@
+/**
+ * Modified MIT License
+ *
+ * Copyright 2018 OneSignal
+ *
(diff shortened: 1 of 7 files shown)


Nightsd01:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/onesignal/onesignal-android-sdk/479)*

PR #523, opened 2018-05-04
Added support for the FCM library

### LINKED ISSUE #507: Google removing GCM in favor of FCM
I received an email saying GCM will no longer work after **April 11, 2019** so apps need to be upgraded to FCM before then. I see that this library uses/depends on GCM. In the [docs](https://documentation.onesignal.com/docs/firebase-cloud-messaging-fcm) it says that it can work with FCM. 

- Do you have docs for how to migrate to FCM using OneSignal? 
- And do you plan to upgrade this library soon to depend on FCM?

jkasten2:
* Added support to use the Firebase FCM library
* GCM can still be used if the app developer hasn't updated yet
* Preventing multiple FCM/GCM registration threads from running
* Added warning to the logcat if GCM is still included in the app project
* Refactored library checks and registration code.
  - Most of the PushRegistratorGPS.java logic was moved into PushRegistratorAbstractGoogle.java and refactored there.
      - Also UI prompt moved into GooglePlayServicesUpgradePrompt.java
   - Some of OSUntil.java internals were refactored
   - Made constants for each notification_types error codes and device_types  
* Resolves #507

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/523)


jkasten2:
Review status: 0 of 20 files reviewed at latest revision, 1 unresolved discussion.

---

*[OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java, line 17 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/523#-LBgnzHqCDDbf2wJGsn0:-LBhl7q8Appv7c1yl0-j:b-se2mis) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/d56717569d7219eac207aa1a2c7cef8921917595/OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java#L17)):*
<details><summary><i>Previously, Nightsd01 (Brad Hesse) wrote…</i></summary><blockquote>

Rather than hard coding these strings, wouldn't it be a bit easier to use `GoogleApiAvailability.GOOGLE_PLAY_SERVICES_PACKAGE` to get the Google Play Services package name, and `GooglePlayServiceUtils.GOOGLE_PLAY_STORE_PACKAGE` to get the store package name? Or are these definitions only available in some API levels or something like that?
</blockquote></details>

Good catch, we use `GooglePlayServiceUtils` in this file already. This bring up that we need to include `com.google.android.gms:play-services-base` as a dependency for this. `com.google.android.gms:play-services-location` include it as a dependency but not everyone will have this.

---


*Comments from [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/523)*

PR #554, opened 2018-05-25
Misuse of multiple SQL statements in execSQL().

lordpyre:
Hello,
I've spotted a misuse of multiple SQL statements in an execSQL(). In fact, after the first index, the others do not get created by SQLite like this. Please consider reviewing and fixing it.
Best,
Csaba

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/554)


jkasten2:
@lordpyre Thanks for the catch! For reference this the multiple statements per `execSQL` is noted below.
https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase#execsql_3

We will review our SQL statements for any other cases where this could be happening. Well will leave this PR open until so so feel free to add to this PR if you find any other until then

Thanks!

jkasten2:
I checked our code base and was not able to find any other cases of this issue, just the index create which you fixed @lordpyre. Merging this PR in. Thanks again!

PR #563, opened 2018-06-07
Fix idsAvailable not firing a 2nd time

### LINKED ISSUE #558: `OneSignal.idsAvailable` callback has null registrationId while `OneSignal.getPermissionSubscriptionState().getSubscriptionStatus().getPushToken()` is not null.
We're seeing an issue in production where after upgrading our client apps to OneSignal 3.9.1 and including FCM in our project, the `OneSignal.idsAvailable` callback sends a null `registrationId` (related to https://github.com/OneSignal/OneSignal-Android-SDK/issues/348).

We've added some investigative logging to track down this issue. We are seeing that at the same time that the `OneSignal.idsAvailable` callback is fired, `OneSignal.getPermissionSubscriptionState().getSubscriptionStatus().toString()` returns a json object with a userId and pushToken like so:

`{"userId":"<redacted>","pushToken":"<redacted>","userSubscriptionSetting":true,"subscribed":true}`

We have asked users to reinstall the app and they are still getting a null registrationId. We cannot reproduce this issue, but we have seen it happen to many clients.

jkasten2:
* Fixed issue where idsAvailable would not fire a 2nd time if registrationId the first time
* resolves #558

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/563)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -1683,7 +1683,7 @@ public class OneSignal {
       runIdsAvailable.run();
    }
 
-   private static void fireIdsAvailableCallback() {
+   static void fireIdsAvailableCallback() {
       if (idsAvailableHandler != null) {
          OSUtils.runOnMainUIThread(new Runnable() {
             @Override
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/UserStatePushSynchronizer.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/UserStatePushSynchronizer.java
@@ -157,5 +157,7 @@ class UserStatePushSynchronizer extends UserStateSynchronizer {
     protected void onSuccessfulSync(JSONObject jsonFields) {
         if (jsonFields.has("email"))
             OneSignal.fireEmailUpdateSuccess();
+        if (jsonFields.has("identifier"))
+            OneSignal.fireIdsAvailableCallback();
     }
 }
--- a/OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
+++ b/OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
@@ -591,13 +591,19 @@ public class MainOneSignalClassRunner {
       //   and is testing the same logic.
       ShadowPushRegistratorGCM.fail = true;
       OneSignalInitWithBadProjectNum();
-
       threadAndTaskWait();
       Robolectric.getForegroundThreadScheduler().runOneTask();
-      assertEquals(-7, ShadowOneSignalRestClient.lastPost.getInt("notification_types"));
 
+      assertEquals(-7, ShadowOneSignalRestClient.lastPost.getInt("notification_types"));
       // Test that idsAvailable still fires
       assertEquals(ShadowOneSignalRestClient.pushUserId, callBackUseId);
+      assertNull(getCallBackRegId); // Since GCM registration failed, this should be null
+
+      // We now get a push token after the device registers with Onesignal,
+      //    the idsAvailable callback should fire a 2nd time with a registrationId automatically
+      ShadowPushRegistratorGCM.manualFireRegisterForPush();
+      threadAndTaskWait();
+      assertEquals(ShadowPushRegistratorGCM.regId, getCallBackRegId);
    }
 
    @Test


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1683,7 +1683,7 @@ public void run() {
       runIdsAvailable.run();
    }
 
-   private static void fireIdsAvailableCallback() {
+   static void fireIdsAvailableCallback() {
```
runningcode: i commented in the github issue, but just curious. has this callback always been expected to fire twice?

runningcode:
Also, do you have a repro case, so we can test on our side as well?

runningcode:
Actually looking at this a second time. This doesn't seem to resolve the underlying issue.
If I understand correctly, `registrationId` and `pushToken` are the same thing.
This doesn't solve the problem that `registrationId` parameter in the idsavailable callback is null, even though `OneSignal.getPermissionSubscriptionState().getSubscriptionStatus().getPushToken()` is not null.

jkasten2:
@runningcode Thanks for reviewing the PR! I was not able to find any other case where the `registrationId` would be `null` when it shouldn't be.

**This PR fixes the following case:**
1. First time the app start, `OneSignal.startInit(...).endInit();` is called.
2. Then app calls `OneSignal.idsAvailable(idsAvailableCallback);`
. SDK tries to register for a FCM / GCM token.
4. Device runs into an error registering the for a pushToken or times out.
   - SDK will keep trying but continues in the background while going to step 4.
5. SDK then registers with onesignal.com to get a OneSignal playerId / userId
6. After getting back a OneSignal `userId` the SDK fires callback;
   -  `idsAvailableCallback(userId: "UUID...", registrationId : null)`
7. Device finally gets a FCM / GCM `pushToken`, SDK fires callback a 2nd time;'
   -  `idsAvailableCallback(userId: "UUID...", registrationId : "fcmToken...")`

This PR fixes the issue where step 7 was not happening.

The behavior of the 2nd fire in the case above has been in place for a while and has aways been the intended behavior. In most cases however, `idsAvailableCallback` will fire once with both values set when there are no timeout issues with FCM / GCM.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
jkasten2: Yes, but only if there was an issue with the device getting a `registrationId`. More details in my comment on the #563 PR.

runningcode:
We saw an issue in production where the callback is fired with a `userId` and a null `registrationId` `idsAvailableCallback(userId: "UUID...", registrationId : null)`.

Inside of the callback we added some logging to track down the issue. `userSubscription=true` and `subscribed=true`.
 We called `OneSignal.getPermissionSubscriptionState().getSubscriptionStatus().getPushToken()` and got a non-null `pushToken`.

Repeated app starts with excellent network conditions would lead to the exact same problem. We also asked our clients to reinstall the app and it led to the same problem. The OneSignal SDK was stuck in a state where the idsAvailableCallback would always fire with a null registrationId even though `userSubscription=true` and `subscribed=true` and getPushToken() returned a valid `pushToken`.

When would it make sense for the `pushToken` to return a value while the `registrationId` is `null`?
This workaround doesn't solve the issue we were seeing.

jkasten2:
@runningcode `getPushToken()` shouldn't be returning a non-null value while `registrationId` is. I'll take another pass at looking at the source and trying to reproduce the issue you described. 

Going to move our conversation back to issue #558 since I want to keep this PR scoped to the title I opened this under.

MacDegger:
Just to mention that we also got hit with this issue (after an update to OneSignal?) Anyway, since we didn't really use the pushToken/registrationid and were only looking for the userid, we replaced the call with OneSignal.getPermissionSubscriptionState().getSubscriptionStatus().getUserID(), which we could use instead ...

MacDegger:
@runningcode we were on 3.8.4 and went to 3.9.3. Might it have to do with using a lambda version of the expression, ie: OneSignal.idsAvailable((userId, registrationId) -> {});?

runningcode:
Ah, good to know. Would you mind commenting on the github issue? https://github.com/OneSignal/OneSignal-Android-SDK/issues/558
It looks like you're seeing the same thing I saw.

3.9.3 should have the fix for the "ids not firing a second time" which was not related to the github issue I filed.
We kept the github issue open since I had a feeling this wouldn't resolve the underlying cause.

PR #611, opened 2018-08-15
Fix invoking interface method on a null object reference

letranloc:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/611)


jkasten2:
@letranloc Thanks for finding the issue and making this PR!

PR #626, opened 2018-09-11
Fix FirebaseInstanceIdService Missing Exception

### LINKED ISSUE #625: Missing FirebaseInstanceIdService Exception
***NOTE*** This issue was originally ([#623](https://github.com/geektimecoil/react-native-onesignal/issues/623)) created for our react-native-onesignal SDK, however it appears to be an issue with recent changes we've made to our Android native SDK so I have moved the issue here. I have copied-pasted the issue below:

**Description:**
Integrated OneSignal with the official docs. iOS is working fine but Android after build came with this error when init called:
screen

**Environment**

Latest React Native (0.56)
Latest OneSignal (3.2.6)
NPM (6.2.0)
MacOS

**Steps to Reproduce Issue:**

1. Install the OneSignal SDK using npm into your project
2. Link and implement using official docs
3. react-native run-android

Nightsd01:
• It appears that in a previous commit ([f614e8b](https://github.com/OneSignal/OneSignal-Android-SDK/commit/f614e8b28bc1cb3b8d1dcf3427db06728464f757#diff-a58cf150e49c38041a5ede3d3b12f3c1)) we added code to disable the FirebaseInstanceIdService if there isn't a default Firebase app

According to [Google's documentation](https://developer.android.com/reference/android/os/Build.VERSION_CODES): `Calls to PackageManager.setComponentEnabledSetting will now throw an IllegalArgumentException if the given component class name does not exist in the application's manifest.`

• However it appears we were only catching `NoClassDefFound` exceptions, even though the package manager would throw an `IllegalArgumentException` in this circumstance.

• Intended to fix #625

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/626)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorFCM.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorFCM.java
@@ -66,6 +66,8 @@ class PushRegistratorFCM extends PushRegistratorAbstractGoogle {
       } catch (NoClassDefFoundError ignored) {
          // Will throw if missing FirebaseInstanceIdService class, ignore in this case.
          // We already print a logcat error in another spot
+      } catch (IllegalArgumentException ignored) {
+         // also not handled
       }
    }
 


jkasten2:
This will throw when FCM is not added to the project. As a work around you can use FCM in your project instead of GCM until this is included in the next release.
```gradle
compile 'com.google.firebase:firebase-messaging:12.0.1'
```

PR #637, opened 2018-09-29
Add ability to set a proxy on requests

tbt-hasan:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/637)


jkasten2:
@tbt-hasan Thanks for the PR however we do not plan on supporting different endpoints for the SDK.

PR #704, opened 2019-01-17
WIP - Update badge count when restoring notifications

jkasten2:
* Also omit trying to restore notifications if notifications are disabled.
* Misc code cleanup around this


OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationLimitManager.java
```diff
@@ -0,0 +1,126 @@
...
+         );
+
+         int notifsToClear = (cursor.getCount() - getMaxNumberOfNotificationsInt()) + notifsToMakeRoomFor;
+         // We have enough room in the notification shade, no need to clear any notifications
+         if (notifsToClear < 1)
+            return;
+
+         while (cursor.moveToNext()) {
+            notifsToClear--;
+            if (notifsToClear < 0)
```
Nightsd01: Small syntax comment, on line 78 you are both decrementing and checking against zero on the same line, but in this loop you are doing these on separate lines. Maybe we could make it consistent in both places

PR #756, opened 2019-05-15
Fix on getTags nested call

Jeasmine:
* Calling nested getTags cause a ConcurrentModificationException
  * Add new thread to getTags and Vector to handler list

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/756)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -38,6 +38,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.TimeZone;
+import java.util.Vector;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -38,6 +38,7 @@
 import java.util.Iterator;
 import java.util.List;
 import java.util.TimeZone;
+import java.util.Vector;
```
jkasten2: Looks like this didn't end up being used, please remove.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1629,61 +1630,56 @@ public static void getTags(final GetTagsHandler getTagsHandler) {
       if (shouldLogUserPrivacyConsentErrorMessageForMethodName("getTags()"))
          return;
 
-      synchronized (pendingGetTagsHandlers) {
```
jkasten2: Glad we could remove these 3 `synchronized` calls! 🎉

PR #798, opened 2019-07-10
Added app id pre-check for changing app id during runtime

mikechoch:
* remoteParams is set null when a app id change occurs
* Unit test written to test app id change during runtime
* Fixed a spelling mistake in OneSignalRemoteParams

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/798)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -627,6 +627,11 @@ public class OneSignal {
       if (isSubscriptionStatusUninitializable())
          return;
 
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -638,8 +643,6 @@ public static void init(Context context, String googleProjectNumber, String oneS
 
       saveFilterOtherGCMReceivers(mInitBuilder.mFilterOtherGCMReceivers);
 
-      // NOTE: This must be called here, something above conflicts with the handlers internals
```
jkasten2: Has this note been addressed in this PR?

PR #801, opened 2019-07-15
Orientation change listener fix for config change flags

mikechoch:
When configChanges for an activity has the orientation flag the view is not recreated, which causes issues with the in app message layout. We need to manually track orientation change and recreate the IAM so that it is shown correctly in portrait and landscape on a device (fix applies to native and wrapper devices).

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/801)


OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java
```diff
@@ -151,6 +161,41 @@ static private void logCurActivity() {
...
+        }
+
+        // Show view
+        for (Map.Entry<String, ActivityAvailableListener> entry : sActivityAvailableListeners.entrySet()) {
+            entry.getValue().available(curActivity);
+        }
+
+        ViewTreeObserver treeObserver = curActivity.getWindow().getDecorView().getViewTreeObserver();
+        for (Map.Entry<String, OSSystemConditionController.OSSystemConditionObserver> entry : sSystemConditionObservers.entrySet()) {
+            KeyboardListener keyboardListener = new KeyboardListener(entry.getValue(), entry.getKey());
```
jkasten2: Do we need to re setup the keyboard listener in the rotation case? Shouldn't the decorView be the same?

mikechoch: The height issue with center modal occurs without resetting up the keyboard listeners

jkasten2: ah gotcha, just want to make sure it was needed and didn't create issues with extra events

PR #802, opened 2019-07-16
Added new OneSignal.addTriggersFromJsonString

jkasten2:
* Added this addTriggersFromJsonString method to make it easy for triggers to be added from any wrapper SDK via a JSON String.
   - This is a common serialization for most OneSignal's wrappers SDKs.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/802)


rgomezp:
unfortunately just caught this, but the function should be named `addTriggersFromJsonArrayString` (just to be consistent) @jkasten2

PR #803, opened 2019-07-16
Added removeTriggersForKeysFromJsonArrayString

jkasten2:
* Added removeTriggersForKeysFromJsonArrayString method to make it easy for triggers to be removed from any wrapper SDK via a JSONArray String.
   - JSON is a common serialization for most OneSignal's wrappers SDKs.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/803)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
@@ -43,6 +43,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.v4.app.NotificationManagerCompat;
 import android.telephony.TelephonyManager;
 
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
```diff
@@ -439,4 +443,16 @@ static boolean hasConfigChangeFlag(Activity activity, int configChangeFlag) {
...
       return hasFlag;
    }
+
+   static @NonNull Collection<String> extractStringsFromCollection(@Nullable Collection<Object> collection) {
+      Collection<String> result = new ArrayList<>();
+      if (collection == null)
+         return result;
+
+      for (Object value : collection) {
+         if (value instanceof String)
```
mikechoch: maybe we could add an else and log that "This is an invalid type" and it only excepts Strings

jkasten2: Done, added a warning to `removeTriggersForKeysFromJsonArrayString`.
Kept `extractStringsFromCollection` the same as wanted to keep this a pure utility method.

jkasten2:
Test failures are due to unrelated flaky tests.

PR #804, opened 2019-07-16
Fix app error App may not schedule more than 100 distinct jobs

Jeasmine:
* Some devices on API 26 and upper cant schedule more than 100 distincts jobs
  * Replace JobService with JobIntentService

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/804)


OneSignalSDK/onesignal/src/main/AndroidManifest.xml
```diff
@@ -64,8 +64,9 @@
         <receiver android:name="com.onesignal.NotificationOpenedReceiver" />
 
         <service android:name="com.onesignal.GcmIntentService" />
-        <service android:name="com.onesignal.GcmIntentJobService"
-                 android:permission="android.permission.BIND_JOB_SERVICE" />
+
+        <service android:name="com.onesignal.GcmJobIntentService"
```
jkasten2: I do like the renaming here but unfortunately for those who use our .jar instead of our .aar this would be a breaking change. Since they would need to update their `AndroidManifest.xml` manually in the .jar case. Can we put back the name?

OneSignalSDK/onesignal/src/main/java/com/onesignal/GcmJobIntentService.java
```diff
@@ -0,0 +1,32 @@
...
+@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+public class GcmJobIntentService extends JobIntentService {
+
+    public static final String BUNDLE_EXTRA = "Bundle:Parcelable:Extras";
+    private static final int JOB_ID = 123890;
+
+    @Override
+    protected void onHandleWork(@NonNull Intent intent) {
+        Parcelable parcelable = intent.getExtras().getParcelable(BUNDLE_EXTRA);
+        BundleCompat bundle = parcelable instanceof PersistableBundle && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1?
```
jkasten2: Small nit, I think we can clean this up a bit by using `BundleCompatFactory`. Feel free to add a new method to `BundleCompat` or  `BundleCompatFactory` if needed.
Or if it really doesn't help it is ok to leave this code as is.

OneSignalSDK/onesignal/src/main/AndroidManifest.xml
Jeasmine: Oh good catch!

OneSignalSDK/onesignal/src/main/java/com/onesignal/GcmJobIntentService.java
jkasten2: 👍 Looks cleaner now! Thanks.

PR #812, opened 2019-07-23
Update proguard configuration

JulienDev:
If R8 is disabled, the build fails because of proguard errors :

```
Note: the configuration refers to the unknown class 'com.onesignal.shortcutbadger.impl.AsusHomeLauncher'
Note: the configuration refers to the unknown class 'com.onesignal.OneSignalUnityProxy'
```

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/812)

### DIFF
--- a/OneSignalSDK/onesignal/consumer-proguard-rules.pro
+++ b/OneSignalSDK/onesignal/consumer-proguard-rules.pro
@@ -33,7 +33,7 @@
 
 -keep class com.onesignal.shortcutbadger.impl.AdwHomeBadger { <init>(...); }
 -keep class com.onesignal.shortcutbadger.impl.ApexHomeBadger { <init>(...); }
--keep class com.onesignal.shortcutbadger.impl.AsusHomeLauncher { <init>(...); }
+-keep class com.onesignal.shortcutbadger.impl.AsusHomeBadger { <init>(...); }
 -keep class com.onesignal.shortcutbadger.impl.DefaultBadger { <init>(...); }
 -keep class com.onesignal.shortcutbadger.impl.EverythingMeHomeBadger { <init>(...); }
 -keep class com.onesignal.shortcutbadger.impl.HuaweiHomeBadger { <init>(...); }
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/consumer-proguard-rules.pro
```diff
@@ -33,7 +33,7 @@
 
 -keep class com.onesignal.shortcutbadger.impl.AdwHomeBadger { <init>(...); }
 -keep class com.onesignal.shortcutbadger.impl.ApexHomeBadger { <init>(...); }
--keep class com.onesignal.shortcutbadger.impl.AsusHomeLauncher { <init>(...); }
+-keep class com.onesignal.shortcutbadger.impl.AsusHomeBadger { <init>(...); }
```
jkasten2: Good catch, this was renamed and missed in this file

OneSignalSDK/onesignal/consumer-proguard-rules.pro
```diff
@@ -54,5 +54,3 @@
 -keep public class com.onesignal.ADMMessageHandler {*;}
 
 -keep class com.onesignal.JobIntentService$* {*;}
-
--keep class com.onesignal.OneSignalUnityProxy {*;}
```
jkasten2: There is a specific Unity target where this file will exist so we need to keep this.

It might be possible to make a 2nd `consumer-proguard-rules.pro` for this target. Not sure if it will merge the main target with the unity target though.

A 3rd option is this proxy should be its own project completely or moved to the OneSignal-Unity-SDK repo.

jkasten2:
@JulienDev It seems the log is just treating it as informational, as it just says "Note: ". Can you provide steps on reproducing?

Left details on the `OneSignalUnityProxy` in the PR review itself.

mikechoch:
@JulienDev Good find!
#1111 is merged and covers the `AsusHomeLauncher` refactor to `AsusHomeBadger`

We will close this one now as the `OneSignalUnityProxy` is required for our Unity SDK and in the future we will decouple this file form our native SDK and remove the keep line from the `consumer-proguard-rules.pro`

PR #816, opened 2019-07-30
Handling some nullPointer exceptions regarding privacy consent

mikechoch:
* mInitBuilder now initialized above privacy consent check inside init
* User privacy consent check will force onAppFocus to return early

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/816)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,5 @@ after_failure:
     - pwd
     - ls -la $HOME
     - ls -la $HOME/android-sdk
+
+dist: trusty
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -28,4 +28,6 @@ after_failure:
     - cd ../..
     - pwd
     - ls -la $HOME
-    - ls -la $HOME/android-sdk
\ No newline at end of file
+    - ls -la $HOME/android-sdk
+
+dist: trusty
```
jkasten2: nit, would be easier to read the environment if `dist: trusty` was one of the first few lines

PR #825, opened 2019-08-20
Fixed NPE with the Builder on app focus

jkasten2:
* Fixed issue where postNotification could be called before an app_id is set.
* Fixed a bug where player created could have been called before an app_id is set.
   - Would have only happened in the specific focus case without initializing the SDK.
* Fixed case where the builder would be null when the app comes into focus if OneSignal.init was not called but a Context was set.
* This could happen if a push was received and later the app is opened but OneSignal was not initialized in Application.onCreate.
* Added validation for app_id to all REST API calls durning tests via a new RestClientValidator class.
* Added new RestClientAsserts to make checking if REST API calls were made much simpler.

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
@@ -62,6 +62,8 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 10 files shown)


EmilAlipiev:
was that released in stable or pre? if not, any ETA please?

HBiSoft:
@EmilAlipiev I believe it was released as a stable release in 13.11.3 - https://github.com/OneSignal/OneSignal-Android-SDK/releases

PR #856, opened 2019-10-08
Updated dependencies and unittest for robolectric

mikechoch:
* Robolectric is now on major release 4.0 (we want to move up to latest soon, but getting to 4.0 is a good starting point)
* Fixed all unittests to allow Robolectric to work correctly
* Removed all AndroidManifest.xml files because the @Config(manifest=...) does not work anymore
* Added some extra shadows to handle AndroidManifest.xml configs

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/856)


OneSignalSDK/onesignal/build.gradle
```diff
@@ -1,24 +1,23 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 27
-    buildToolsVersion '27.0.3'
-
+    compileSdkVersion 28
     defaultConfig {
+        manifestPlaceholders = [onesignal_app_id: 'b2f7f966-d8cc-11e4-bed1-df8f05be55ba',
```
jkasten2: Let's check to make sure by adding the setting of these `manifestPlaceholders` that it doesn't hardcode them into the generated .aar

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
```diff
@@ -893,7 +884,7 @@ public void badgeCountShouldNotIncludeOldNotifications() {
       NotificationBundleProcessor_ProcessFromGCMIntentService(blankActivity, getBaseNotifBundle(), null);
 
       // Go forward 1 week
-      ShadowSystemClock.setCurrentTimeMillis(System.currentTimeMillis() + 604_801L * 1_000L);
+      advanceSystemTimeBy(604_801 * 1_000L);
```
jkasten2: Lets include the units here. Maybe call it `advanceSystemTimeByMs`

OneSignalSDK/onesignal/build.gradle
mikechoch: didnt mean to push this honestly, will take it out

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
mikechoch: Only reason I wasnt adding units in the name is because the parameter is `long millis`

jkasten2: Gotcha, I guess the param hint will end up showing in Android Studio with `millis` so it won't be an issue.

OneSignalSDK/onesignal/build.gradle
jkasten2: You are right, it gets sync errors from Android Studio if we remove these. Odd thing is that `gradlew onesignal:assembleRelease` works just fine.

However if we keep these static values generated `.aar` also includes them instead of keeping the placeholders.

I found a work around though, if we redefine the placeholders it fixes the sync error & correctly keeps the placeholders in the build .aar file. See https://github.com/OneSignal/OneSignal-Android-SDK/pull/856/commits/1e93337ad05342233f59c5b3f49dffa04ccaa1cb

PR #857, opened 2019-10-08
Fix for Lenovo Yoga crash due to NullPointerException inside RootToolsInternalMethods.isRooted()

vlad-roid:
Fixes the crash below, affecting all Lenovo Yoga devices running Android 6. Crash hasn't occurred anymore since deploying this fix.
image

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/857)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -1303,7 +1303,9 @@ public class OneSignal {
 
       deviceInfo.put("net_type", osUtils.getNetType());
       deviceInfo.put("carrier", osUtils.getCarrierName());
-      deviceInfo.put("rooted", RootToolsInternalMethods.isRooted());
+      try {
+         deviceInfo.put("rooted", RootToolsInternalMethods.isRooted());
+      } catch (Throwable t) {}
 
       OneSignalStateSynchronizer.updateDeviceInfo(deviceInfo);
 


jkasten2:
@vlad-roid Thanks for the PR however I don't how this could be an NPE. [`RootToolsInternalMethods` line 33](https://github.com/OneSignal/OneSignal-Android-SDK/blob/c33b7a943fb8e08e174ecae8d0a63c67c0a5140c/OneSignalSDK/onesignal/src/main/java/com/onesignal/RootToolsInternalMethods.java#L33) is a String array that is defined in the method.

The only way this could be possible is if this class was modified in the APK by the user. Modification like this is something we won't be able to prevent.

vlad-roid:
@jkasten2 Baseless assumptions.
1. This NPE and crash affects around 100 users across our apps _and is the most occurring crash in each of them_, so there's no way each user modified the APK in the same way, nor is there any point or advantage to any of the users in doing that.
2. There is also no way they had access to an APK that was modified and distributed elsewhere because our apps are distributed as Android App Bundles, not as APKs.
3. **Even if** theoretically it's still due to an APK modification, how come it affects only 3 specific Lenovo Yoga devices, every single user of these 3 devices, and not a single other device?
4. **Even if** for whatever absolutely impossible reason you still think it's an APK modification, which would mean these users have a specific build they wouldn't update, how come this crash had been happening for 6 different releases _across months_ and stopped occurring for 100% of all affected users after deploying this fix? It's been 3 weeks since deployment, and the NPE hasn't occurred ever since, so each affected user has either uninstalled the app or updated it.

I'm also fully aware that it doesn't make sense this specific array access would throw an NPE, but still, it's not healthy to simply assume it's impossible, especially because RootToolsInternalMethods was not written by OneSignal. Many manufacturers build in such a huge amount of weird errors in their custom firmwares (Samsung above all), and in this case it's absolutely Lenovo who's at fault. But manufacturers won't fix things like that, considering they normally don't even fix huge issues which affect millions of devices (Samsung again).

It's still a very good question how accessing an internally declared array, in a custom method that a manufacturer could never access or modify, could throw an NPE. I would say Lenovo is using some weird modified JRE where the garbage collector has serious bugs and for whatever reason in this specific case leads to that array being garbage collected right after initialization, effectively setting it back to null.

Finally, having said all that, what can possibly be the harm in merging this fix? At least this way you would do us the favor of not having to constantly merge and republish the fork whenever OneSignal publishes a new SDK release, saving hours of time wasted on our side for maintaining the fork. We don't have the resources to do it here from Australia (none of the affected devices are sold here anymore), but if OneSignal absolutely requires a way to reproduce the crash, please order one of the 3 affected devices, e.g. https://www.amazon.com/Lenovo-Yoga-Tab-Pro-ZA0F0099US/dp/B01N3USBZT and run the OneSignal sample, and I'm sure it will crash straight on app start.

nlgtuankiet:
Hey guys, we also get the same crashes, I wonder if we have any way to prevent this?
image

vlad-roid:
Hey @nlgtuankiet, happy to see someone else stumbled upon this issue ;)
Until this is merged (which most probably won't happen, because it's not possible to reproduce the issue without buying a Lenovo Yoga tablet), to fix it, feel free to add my repository via Gradle:
```
repositories {
    maven { url "https://jitpack.io" }
}

dependencies {
    implementation 'com.github.vlad-roid:OneSignal-Android-SDK:3.12.7@aar'
}
```
I always update it to the latest OneSignal SDK version. The only code change is a try/catch around the call that causes the crash, otherwise everything is exactly the same.

khang-tiki:
> @vlad-roid Thanks for the PR however I don't how this could be an NPE. [`RootToolsInternalMethods` line 33](https://github.com/OneSignal/OneSignal-Android-SDK/blob/c33b7a943fb8e08e174ecae8d0a63c67c0a5140c/OneSignalSDK/onesignal/src/main/java/com/onesignal/RootToolsInternalMethods.java#L33) is a String array that is defined in the method.
> 
> The only way this could be possible is if this class was modified in the APK by the user. Modification like this is something we won't be able to prevent.

@jkasten2 I agree that it's impossible to happen, but with the data we got from crash devices, these users hadn't modified the APK, it's installed correctly from PlayStore. I think we should treat it as a "device-specific issue", and provide a workaround to avoid it (try-catch is a not bad workaround).

khang-tiki:
I would like to reopen it in different PR, @jkasten2 please take a look:  #978

nlgtuankiet:
@jkasten2 FYI
image

PR #861, opened 2019-10-08
Outcome events release

mikechoch:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/861)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
```diff
@@ -0,0 +1,243 @@
...
+            OneSignal.Log(OneSignal.LOG_LEVEL.ERROR, "Generating addNotificationId:JSON Failed.", e);
+        }
+    }
+
+    void cleanSession() {
+        this.session = null;
+    }
+
+    void restartSessionIfNeeded() {
+        long resetTime = sessionTime != null ? System.currentTimeMillis() - sessionTime : HALF_MIN_IN_MILLIS;
```
jkasten2: This is duplicated code, we check this in `OneSignal.doSessionInit` where this is being called from

mikechoch: @Jeasmine is there a reason we check session time twice? Can we remove this if statement since we check this in the previous method call?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -3034,4 +3155,124 @@ static void fireEmailUpdateFailure() {
...
       }
    }
+
+   /*
+   * Start OneSignalOutcome module
+   */
+
+   private static OutcomeSettings outcomeSettings = null;
+
+   public static void changeOutcomeSettings(OutcomeSettings settings){
```
jkasten2: Does look like this is used anywhere. Also if we do need it it should not be public

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
```diff
@@ -0,0 +1,295 @@
...
+     */
+    void attemptSessionUpgrade() {
+        // We will try to override any session with DIRECT
+        if (OneSignal.getAppEntryState().isNotificationClick()) {
+            setSession(Session.DIRECT, directNotificationId, null);
+            return;
+        }
+
+        // We will try to override the UNATTRIBUTED session with INDIRECT
+        if (session.isUnattributed()) {
```
Jeasmine: sorry just want to check this case, if i have my app with an INDIRECT session with indirect notifications [id: "abcd"], close the app, and receive an Notification with notification id ["efgh"], if a open the app from the app icon no new session was triggered, the session will still de indirect for the previous session and continue to have the notificationsId as [id: "abcd"], not [id: "abcd", "efgh"] thats the desire flow?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
```diff
@@ -0,0 +1,295 @@
...
+     * Attempt to override the current session before the 30 second session minimum
+     * This should only be done in a upward direction:
+     *   * UNATTRIBUTED can become INDIRECT or DIRECT
+     *   * INDIRECT can become DIRECT
+     *   * DIRECT can become DIRECT
+     */
+    void attemptSessionUpgrade() {
+        // We will try to override any session with DIRECT
+        if (OneSignal.getAppEntryState().isNotificationClick()) {
+            setSession(Session.DIRECT, directNotificationId, null);
```
Jeasmine: cant we depend on the directNotificationId to be set? and to pass as param OneSignal.getAppEntryState().isNotificationClick() boolean? to remove this static dependency and keep only composition relation from OneSignal to SessionManager

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2064,18 +2070,38 @@ public static void handleNotificationOpen(Context inContext, JSONArray data, boo
...
       runNotificationOpenedCallback(data, true, fromAlert);
 
-      // Open/Resume app when opening the notification.
-      if (!fromAlert && !urlOpened && !defaultOpenActionDisabled)
-         startOrResumeApp(inContext);
+      // Check if the notification click should lead to a DIRECT session
+      if (sessionManager != null && shouldInitDirectSessionFromNotificationOpen(inContext, fromAlert, urlOpened, defaultOpenActionDisabled)) {
+         // We want to set the app entry state to NOTIFICATION_CLICK when coming from background
+         appEntryState = AppEntryAction.NOTIFICATION_CLICK;
+         sessionManager.onDirectSessionFromNotificationOpen(notificationId);
```
Jeasmine: i think this check if the application was on the foreground right? so in theory onDirectSessionFromNotificationOpen wont be set if the app was on foreground and the notification was clicked

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
```diff
@@ -0,0 +1,192 @@
...
+
+    static class OutcomeException extends Exception {
+
+        OutcomeException(String message) {
+            super(message);
+        }
+
+    }
+
+    private Set<String> eventsSent = OSUtils.newConcurrentSet();
```
Jeasmine: looking to the session stuff, this we need eventsSent to live as the session lives right? while the same session is up we should not let the user sent twice the same event. So if the user close and re open the app on the 30 secs of the session in theory i think that wont work?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
```diff
@@ -0,0 +1,295 @@
...
+                jsonObject.put(OutcomesUtils.NOTIFICATIONS_IDS, indirectNotificationIds);
+            }
+        } catch (JSONException e) {
+            OneSignal.Log(OneSignal.LOG_LEVEL.ERROR, "Generating addNotificationId:JSON Failed.", e);
+        }
+    }
+
+    void restartSessionIfNeeded() {
+        // Avoid reset session if app was focused due to a notification click (direct session recently set)
+        if (OneSignal.getAppEntryState().isNotificationClick())
```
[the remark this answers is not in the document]
jkasten2: Good point, I think should revisit it later however. We have a ton of code today looking at OneSignal.java to get global state and need to come up with a good pattern for it. OneSignal.java should really be just incharge of holding onto managers / controllers and have other classes for states where they fit best. When we add these other class we will do need to prevent circular references which is what we have here though.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
jkasten2: Right we should just have one notification id, the old one of id `abcd` from your example.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
jkasten2: For unique outcomes they should not be sent more than once per notification id. @mikechoch Is making the changes to handle this right now.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
```diff
@@ -31,26 +37,46 @@
...
+
+        initUniqueOutcomeEventsSentSets();
+    }
+
+    /**
+     * Init the sets used for tracking attributed and unattributed unique outcome events
+     */
+    private void initUniqueOutcomeEventsSentSets() {
+        attributedUniqueOutcomeEventsSentSet = OSUtils.newConcurrentSet();
+        unattributedUniqueOutcomeEventsSentSet = OSUtils.newConcurrentSet();
```
Jeasmine: if i close and re open on less that 30 seconds in theory unattributedUniqueOutcomeEventsSentSet should still keep the last uniqueOutcomes, but well we can agree on ignoring this case

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
```diff
@@ -150,8 +208,6 @@ public void run() {
                                     new OutcomeEvent(session, notificationIds, name, timestampSeconds, params));
                         }
                     }, OS_SAVE_OUTCOMES).start();
-                } else {
```
Jeasmine: shoudnt the outcome being remove from unattributedUniqueOutcomeEventsSentSet  if it was the case?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
```diff
@@ -100,32 +126,56 @@ void onSuccess(String response) {
...
                 .newInstance()
                 .setWeight(value)
                 .build();
-        sendOutcomeEvent(name, params, callback);
+        sendOutcomeEvent(name, notificationIds, params, callback);
     }
 
-    private void sendOutcomeEvent(@NonNull final String name, @Nullable final OutcomeParams params, @Nullable final OneSignal.OutcomeCallback callback) {
+    private void sendOutcomeEvent(@NonNull final String name, @Nullable final JSONArray notificationIds, @Nullable final OutcomeParams params, @Nullable final OneSignal.OutcomeCallback callback) {
         OSSessionManager.SessionResult sessionResult = osSessionManager.getSessionResult();
```
Jeasmine: no need to call this again right? you can pass a new OSSessionManager.SessionResult

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
```diff
@@ -175,15 +231,64 @@ public void run() {
...
+        return getUniqueOutcomeNotificationIds(name, notificationIds);
     }
 
-    void sendOutcomeEvent(@NonNull String name, @NonNull String value) throws OutcomeException {
-        if (value.isEmpty())
-            throw new OutcomeException("Value must not be empty");
+    /**
+     * Validate whether or not the JSONArray of notificationIds has ids that have not been sent with the specific unique outcome name
+     */
+    private boolean hasNoUniqueOutcomeNotificationIds(String uniqueName, JSONArray notificationIds) {
```
Jeasmine: is this being used on other place? maybe it can return the JSONArray and avoid two fors

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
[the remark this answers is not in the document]
mikechoch: Eventually we agree we need to create independent classes from OneSignal that handle these states

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
mikechoch: If `getAppEntryState().isNotificationClick()` is `true` we would have already had to set the `directNotificationId`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
mikechoch: But if `sessionManager` was null it would crash anyway. Maybe we should check if it is null and then initialize it if it is?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
mikechoch: You're right, I will fix this for unattributedUniqueOutcomeEventsSentSet

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
mikechoch: You are caching on failure though so when ever your cached outcomes are reattempted it will be handled

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
mikechoch: I have to modify `notificationIds` in the ATTRIBUTED for unique outcome events

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
Jeasmine: yeah but cache can be disable, so in that case you should be able to send the outcome again

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -3111,17 +3109,13 @@ public static void sendOutcomeWithValue(@NonNull String name, float value) {
...
    public static void sendOutcomeWithValue(@NonNull String name, float value, OutcomeCallback callback) {
-      if (!validateOutcomeEntry(name))
+      if (!isValidOutcomeEntry(name))
          return;
 
       outcomeEventsController.sendOutcomeEventWithValue(name, value, callback);
    }
 
-   private static boolean validateOutcomeEntry(String name) {
-      if (outcomeEventsController == null) {
```
jkasten2: We still want to have the `outcomeEventsController` `null` check here. As init or any of our services / broadcastReceiver  may not have been called to setup a context.

In the future we can have it go through `addTaskToQueue` like we have for `sendTags` so the call isn't drop if the dev calls things out of order.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -3111,17 +3109,13 @@ public static void sendOutcomeWithValue(@NonNull String name, float value) {
    }
 
    public static void sendOutcomeWithValue(@NonNull String name, float value, OutcomeCallback callback) {
-      if (!validateOutcomeEntry(name))
+      if (!isValidOutcomeEntry(name))
```
jkasten2: 👍 on `is`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
Jeasmine: yeah i mean intead of passing JSONArray notificationIds as param,  pass an OSSessionManager.SessionResult in that way you are not calling again osSessionManager.getSessionResult()

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: that can be an option

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
mikechoch: We cant modify the `notificationIds` inside of the `SessionResult` so we would have to pass another param then

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
mikechoch: Why would we disable cache though? Just wondering

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
mikechoch: Should we pass the extra param? It will be a 5 param method then

Jeasmine: but you can build a new SessionResult. In that way getSessionResult is not being called twice

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
Jeasmine: maybe the user/the host application may want to disable it? OutcomeSetting is there for the user to be able to disable it

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
jkasten2: It looks like this was used to mock out the settings for tests in the past but they were changed to not use it any more. Android Studio now says it isn't called any where any more

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
jkasten2: Running `getSessionResult()` more than once won't have any side effects. I think we might be able to clean this up with some kind of internal `OutcomeEvent` class in the future instead of adding more params. But I don't think it is necessary to make this change now.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
jkasten2: `cacheActive` should always be enabled so outcomes can always be set offline. Don't see a case where it would be different from app to app.

jkasten2: We can leave it as package-private for now so the code is there, we can open the API up later if it is really needed.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
jkasten2: Let's leave this as is for now. We can refactor how the dependency on `appEntryState` is structured later

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
jkasten2: We don't want a notification click when the app is in the foreground to change the session, only background clicks

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
jkasten2: @mikechoch Fixed this.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OutcomeEventsController.java
jkasten2: Let's leave this as is for now. We can revisit if we want to support disabling caching in the future.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSessionManager.java
jkasten2: This isn't relevant after the recent changes we have made

PR #873, opened 2019-10-22
Outcome event clean up

mikechoch:
* Removed OutcomeParams class and instead only use a float weight
* Fixed all tests that broke after removing OutcomeParams
* Cleaned up requests in OutcomeEventsRepository so now there is only 1 of each session (DIRECT, INDIRECT, UNATTRIBUTED)
* Reordered expected outcome unit tests because of outcome event changes to JSON creation
* Added checking in two places for weight of 0 being sent with an OutcomeEvent, one is when we call the public method (logs an error) and the other is in the toJSON of the OutcomeEvent preventing a weight of 0 from being added into the JSON

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/873)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalDbContract.java
```diff
@@ -66,7 +66,7 @@
       public static final String COLUMN_NAME_TIMESTAMP = "timestamp";
       public static final String COLUMN_NAME_NAME = "name";
       public static final String COLUMN_NAME_SESSION = "session";
-      public static final String COLUMN_NAME_PARAMS = "params";
+      public static final String COLUMN_NAME_WEIGHT = "weight";
```
Jeasmine: params was made, because on the future it will be able to save strings, bundles, and other types, so it was a generic param for extension purposes

mikechoch: We decided it over complicates the functionality for now since it only has 1 attribute. We can always re-add it in the future, but for now we should keep it as simple as possible for easier usage and understanding.

The main reason is we now return the `OutcomeEvent` in the `onSuccess` callback and a `getParams` would be very confusing for the end user and they would need to know the keys inside the `JSON` to get the weight of the `OutcomeEvent`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -3120,7 +3120,7 @@ public static void sendOutcomeWithValue(@NonNull String name, float value) {
    }
 
    public static void sendOutcomeWithValue(@NonNull String name, float value, OutcomeCallback callback) {
-      if (!isValidOutcomeEntry(name))
+      if (!isValidOutcomeEntry(name) && !isValidOutcomeValue(value))
```
jkasten2: This should be an OR instead of an AND.

mikechoch: Got it!

PR #911, opened 2019-12-14
Remove BroadcastReceiver notification handler

Jeasmine:
* BroadcastReceiver was used to launch context and intent handler on notification click
  * Change to Activity
  * Avoid singleTop launch mode on host application

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/911)


mikechoch:
@jkasten2 should look at this too, but I tested this on API 25 to API 29 and it works perfect, no weird transitions or anything. Also clears the stack now with the flags since we switched to Activity, solid work.

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
```diff
@@ -69,34 +53,39 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Random;
+
 import static com.onesignal.OSUtils.getResourceString;
 
 class GenerateNotification {
-   private static Context currentContext = null;
```
jkasten2: 👍 Nice clean up on the static Context!

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
```diff
@@ -69,34 +53,39 @@
...
 
    private static void setStatics(Context inContext) {
-      currentContext = inContext;
-      packageName = currentContext.getPackageName();
-      contextResources = currentContext.getResources();
-
-      PackageManager packageManager = currentContext.getPackageManager();
-      Intent intent = new Intent(currentContext, NotificationOpenedReceiver.class);
-      intent.setPackage(currentContext.getPackageName());
-      if (packageManager.queryBroadcastReceivers(intent, 0).size() > 0) {
```
jkasten2: Since this is a breaking change let's rebase this PR off the 4.0.0 branch.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2093,7 +2093,7 @@ static boolean startOrResumeApp(Context inContext) {
       Intent launchIntent = inContext.getPackageManager().getLaunchIntentForPackage(inContext.getPackageName());
       // Make sure we have a launcher intent.
       if (launchIntent != null) {
-         launchIntent.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT | Intent.FLAG_ACTIVITY_NEW_TASK);
+         launchIntent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_NEW_TASK);
```
jkasten2: Will these flag result in a new Activity each time? Or will it be reused if it already exists?
For wrapper SDK we need to the flag to result in using existing instances, since it will create multiple instances of the framework environment and handles. Such as with react native and Unity.

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
mikechoch: I rebased it

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: If it already exist will be reused, for example if i have A->B->C and i launched A then the stack will be the A already there

PR #913, opened 2019-12-17
Fix/remote params exception on app id update

mikechoch:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/913)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSEmailSubscriptionState.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSEmailSubscriptionState.java
@@ -62,7 +62,12 @@ public class OSEmailSubscriptionState implements Cloneable {
     }
 
     void setEmailUserId(@NonNull String id) {
-        boolean changed = id != null && !id.equals(emailUserId);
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSEmailSubscriptionState.java
```diff
@@ -62,7 +62,7 @@ void clearEmailAndId() {
     }
 
     void setEmailUserId(@NonNull String id) {
-        boolean changed = !id.equals(emailUserId);
+        boolean changed = id != null && !id.equals(emailUserId);
```
jkasten2: `changed` should be `true` if setting email to `null` when it was set. This does not cover this. Let's use the same logic as `OSSubscriptionState.setUserId`.

mikechoch: I will need to talk to you about this issue in general and what we should do to prevent a crash here instead.

mikechoch:
Linking issues:
#916

PR #915, opened 2019-12-19
MERGED IN ANOTHER PR - Start of removing OneSignal init

mikechoch:
* Created two new init methods, setAppId and setAppContext, these are required to init OneSignal
* Deleted mInitBuilder and added all params into OneSignal (we should do something like this again, but more organized and not inside of a builder, maybe a simple singleton class?)
* Created a dummy OSNotificationWillShowInForegroundResult.java for now until we create the real thing, but this will replace notification received handler

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/915)


mikechoch:
I almost got extremely OCD (telephone poles) on this PR but I kept it somewhat simple and more of an intro into the major changes we want to release.

First of all I meant for this PR to be only about `init` and changing things to rely on `setAppId` and `setAppContext`. However, after beginning to notice so much dependence, I started refactoring a little more but don't think I pushed past the scope here.

**Major changes include:**
* Removing `mInitBuilder`, I think we should rethink this class and stay away from this builder pattern. I deleted basically everything associated with it.
* Created all new setters for `notificationWillShowInForeground`, `notificationOpened`, and `inAppMessageClicked`
* Added all attributes from `mInitBuilder` into `OneSignal.java` for now
* Created the setAppId and setAppContext methods with the correct modifications to mimic previous functionaility (haven't tested everything)
* Can not confirm how unit tests condition, but it will take a lot of work to refactor them entirely. 
  * I think we should create a PR off of major change branches to separate the things unit test related and SDK related

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -232,171 +230,45 @@ public String getMessage() {
...
 
    private static LOG_LEVEL visualLogLevel = LOG_LEVEL.NONE;
    private static LOG_LEVEL logCatLevel = LOG_LEVEL.WARN;
 
-   private static String userId = null, emailId = null;
+   private static String userId = null;
+   private static String emailId = null;
    private static int subscribableStatus;
 
+   // TODO: These should be cleaned up and managed else where maybe?
```
jkasten2: 👍 Yes, I am ok with doing so in another PR to keep this one smaller.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -572,139 +443,177 @@ private static OSEmailSubscriptionState getLastEmailSubscriptionState(Context co
...
       }
+
+      boolean wasAppContextNull = (appContext == null);
+      appContext = context.getApplicationContext();
+      setupActivityLifecycleListener(wasAppContextNull);
+      setupPrivacyConsent(appContext);
+
+      OneSignal.onesignalLog(LOG_LEVEL.VERBOSE, "setAppContext(context) finished, checking if appId has been set before proceeding...");
+      if (appId == null) {
+         OneSignal.onesignalLog(LOG_LEVEL.WARN, "appContext set, but please call setAppId(appId) with a valid appId to complete OneSignal init!");
```
jkasten2: Attempt to read `appId` from shared perfs when context is set. Wrapper SDK will call setContext first and setAppId won't be called until the wrapper env starts up

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -572,139 +443,177 @@ private static OSEmailSubscriptionState getLastEmailSubscriptionState(Context co
...
-      init(context, googleProjectNumber, oneSignalAppId, null, null);
+   public static void setNotificationWillShowInForegroundHandler(NotificationWillShowInForegroundHandler callback) {
+      notificationWillShowInForegroundHandler = callback;
    }
 
-   public static void init(Context context, String googleProjectNumber, String oneSignalAppId, NotificationOpenedHandler notificationOpenedHandler) {
-      init(context, googleProjectNumber, oneSignalAppId, notificationOpenedHandler, null);
+   public static void setNotificationOpenedHandler(NotificationOpenedHandler callback) {
+      notificationOpenedHandler = callback;
+      if (appContext != null)
```
jkasten2: This seems like a hack as we setting the same value if it exists. We just process the the list of unfired notifications here if there is any and init is done

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -572,139 +443,177 @@ private static OSEmailSubscriptionState getLastEmailSubscriptionState(Context co
...
+         // Set app id null since OneSignal was not init fully
+         appId = null;
          return;
       }
 
-      mInitBuilder = createInitBuilder(notificationOpenedHandler, notificationReceivedHandler);
-
       if (!isGoogleProjectNumberRemote())
-         mGoogleProjectNumber = googleProjectNumber;
+         googleProjectNumber = "";
```
jkasten2: Lets remove these 2 lines, we don't need to set googleProjectNumber to a blank string

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -572,139 +443,177 @@ private static OSEmailSubscriptionState getLastEmailSubscriptionState(Context co
...
          return;
 
-      // Pre-check on app id to make sure init of SDK is performed properly
-      //    Usually when the app id is changed during runtime so that SDK is reinitialized properly
-      if (appId != null && !appId.equals(oneSignalAppId))
-         initDone = false;
-
       if (initDone) {
-         if (mInitBuilder.mNotificationOpenedHandler != null)
+         if (notificationOpenedHandler != null)
```
jkasten2: I think we can clean this up since we don't have an external init that someone can call more than once.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -739,6 +648,22 @@ public static void init(Context context, String googleProjectNumber, String oneS
       startPendingTasks();
    }
 
+   private static void setupActivityLifecycleListener(boolean wasAppContextNull) {
+      // Register the lifecycle listener of the app for state changes in activities with proper context
+      ActivityLifecycleListener.registerActivityLifecycleCallbacks((Application) appContext);
```
jkasten2: Shouldn't this be inside of `if (wasAppContextNull)`? If so let's just remove the boolean param and just guard around where it was called.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -979,8 +900,8 @@ else if (pushStatusRuntimeError(subscribableStatus))
       });
    }
 
-   private static boolean pushStatusRuntimeError(int subscribableStatus) {
-      return subscribableStatus < -6;
+   private static boolean pushStatusRuntimeError(int subscriptionStatus) {
```
jkasten2: `subscribableStatus` was an intended name here. It isn't if the device is subscribed but if it could be.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1049,11 +971,24 @@ public static void provideUserConsent(boolean consent) {
...
+   private static void reassignDelayedInitParams() {
+      appContext = delayedInitParams.context;
+      googleProjectNumber = delayedInitParams.googleProjectNumber;
+      appId = delayedInitParams.appId;
+      notificationWillShowInForegroundHandler = delayedInitParams.notificationWillShowInForegroundHandler;
+      notificationOpenedHandler = delayedInitParams.notificationOpenedHandler;
+
+      delayedInitParams = null;
+
+       if (appContext != null)
```
jkasten2: Seems like a hack of setting the same value. We should call the method we need to here.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
```diff
@@ -1092,7 +1115,10 @@ public void shouldAddDefaultButtonToAlertDialog() throws Exception {
    @Config(shadows = {ShadowGcmBroadcastReceiver.class})
    public void shouldPreventOtherGCMReceiversWhenSettingEnabled() throws Exception {
       OneSignal.setInFocusDisplaying(OneSignal.OSInFocusDisplayOption.InAppAlert);
-      OneSignal.startInit(blankActivity).filterOtherGCMReceivers(true).init();
+//      OneSignal.startInit(blankActivity).filterOtherGCMReceivers(true).init();
```
jkasten2: Delete commented out line

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
```diff
@@ -1077,7 +1097,10 @@ public void shouldNotSetAlertnessFieldsOnLowPriority() {
    @Test
    public void shouldAddDefaultButtonToAlertDialog() throws Exception {
       OneSignal.setInFocusDisplaying(OneSignal.OSInFocusDisplayOption.InAppAlert);
-      OneSignal.startInit(blankActivity).init();
+//      OneSignal.startInit(blankActivity).init();
```
jkasten2: Delete commented out line

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -2448,41 +2473,42 @@ public void ensureNoConcurrentUpdateCallsWithSameData_forPendingActiveTime() thr
...
-      Field OneSignal_Builder_mDisplayOption = builder.getClass().getDeclaredField("mDisplayOption");
-      OneSignal_Builder_mDisplayOption.setAccessible(true);
-      OneSignal.OSInFocusDisplayOption inFocusDisplayOption = (OneSignal.OSInFocusDisplayOption)OneSignal_Builder_mDisplayOption.get(builder);
-      assertEquals(inFocusDisplayOption, OneSignal.OSInFocusDisplayOption.Notification);
-
-      threadAndTaskWait();
-   }
+   // TODO: Builder method should be deleted?
+//   @Test
+//   public void testMethodCallsBeforeInit() throws Exception {
```
jkasten2: We need to keep keep the intent of this test. We should maybe split this into 3 tests.
1. Methods calls before setAppId and setContext
2. Methods calls before setAppId
3. Methods calls before setContext

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -2753,7 +2779,9 @@ public void testSessionTimeTrackingOnPrivacyConsent() throws Exception {
 
       // Provide consent to OneSignal SDK amd forward time 2 minutes
       OneSignal.provideUserConsent(true);
-      threadAndTaskWait();
+      // TODO: This passed without a resume here, why?
```
jkasten2: Did you mean why doesn't this pass without a resume? `OneSignalInit();` already calls resume.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -2753,7 +2779,9 @@ public void testSessionTimeTrackingOnPrivacyConsent() throws Exception {
...
+      blankActivityController.resume();
```
jkasten2: This test shouldn't have needed to be modified to pass.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -3451,6 +3483,7 @@ public void onOSSubscriptionChanged(OSSubscriptionStateChanges stateChanges) {
 
    @Test
    public void shouldFireSubscriptionObserverWhenUserDisablesNotifications() throws Exception {
+//      OneSignal.unsubscribeWhenNotificationsAreDisabled(false);
```
jkasten2: Remove commented out line

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
mikechoch: Yes, sorry. This "doesn't" pass without resume here for some reason

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
mikechoch: Thats what I thought too, gonna dig in a little more

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
mikechoch: Okay sounds good to me

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
mikechoch: Left6 some comments so we knew where it came from but will delete now

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
mikechoch: Okay sounds good

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
mikechoch: I have a better idea, Ill see what i can do

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
mikechoch: Sounds good to me

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
mikechoch: Okay sounds good
Ill just do `appId = getSavedAppId();`
before the null check

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
mikechoch: Got it

PR #920, opened 2020-01-07
When prompting for permissions, we use a PermissionActivity

mikechoch:
* This is not a fully built out tool, but it is working for location currently
  * Eventually we may be able to use this for other permission prompts as well

* The issue was the default transition from current activity to our PermissionActivity
* This is now fixed by creating a 0s duration anim for fading in and fading out of the activity

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/920)


mikechoch:
Here is a demo video to go along showing the before and after.
[Demo Video PermissionActivity Flickering](https://www.dropbox.com/s/pbo43mmmb1s5dzm/Location_Permission_Prompt_Activity.mp4?dl=0)

PR #927, opened 2020-01-21
Make IAM able to re display multiple times

Jeasmine:
* Changes on OSInAppMessage class, add two new params delay and limit
  * Make OSInAppMessageController handle the multiple re displays
  * Save on local database quantity of displays and last display time per IAM
  * Add unit test for new functionality

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/927)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -34,24 +43,42 @@
...
+    private long lastDisplayTime = -1;
+    //Current quantity of displays
+    private int displaysQuantity = 0;
+    //Quantity of displays limit
+    private int displaysLimit = Integer.MAX_VALUE;
+    //Delay between displays in seconds
+    private long delay = 0;
     private double displayDuration;
 
+    private boolean reDisplayEnabled = false;
```
mikechoch: `reDisplayEnabled` should be `redisplayEnabled`

Redisplay is one word, so no capitol D

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -137,22 +189,86 @@ boolean takeActionAsUnique() {
...
+    int getDisplaysQuantity() {
+        return displaysQuantity;
+    }
+
+    void setDisplaysQuantity(int displaysQuantity) {
+        this.displaysQuantity = displaysQuantity;
+    }
+
+    void increaseDisplayQuantity() {
+        this.displaysQuantity++;
```
mikechoch: `increaseDisplayQuantity` should be `incrementDisplayQuantity`

increase could be any quantity, increment means to add 1 to a current value

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -137,22 +189,86 @@ boolean takeActionAsUnique() {
...
+
+    void setLastDisplayTime(long lastDisplayTime) {
+        this.lastDisplayTime = lastDisplayTime;
+    }
+
+    boolean isReDisplayEnabled() {
+        return reDisplayEnabled;
+    }
+
+    boolean isRemainingDisplays() {
```
mikechoch: `isRemainingDisplays` should be `shouldDisplayAgain`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -27,16 +29,20 @@
...
 
     // IAMs loaded remotely from on_session
     //   If on_session won't be called this will be loaded from cache
     @NonNull private ArrayList<OSInAppMessage> messages;
     // IAMs that have had their trigger(s) evaluated to true;
     //   This mean they have been added to the queue to display, or have already displayed
-    @NonNull final private Set<String> triggeredMessages;
+    @NonNull final private Set<String> dismissedAndToDismissMessages;
+    // IAMs displayed with last displayed time and quantity of displays data
+    @NonNull final private List<OSInAppMessage> displayedMessagesData;
```
mikechoch: Why aren't we using the previous `messages` instead?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -27,16 +29,20 @@
...
     private OSSystemConditionController systemConditionController;
+    private OSInAppMessageRepository inAppMessageRepository;
 
     // IAMs loaded remotely from on_session
     //   If on_session won't be called this will be loaded from cache
     @NonNull private ArrayList<OSInAppMessage> messages;
     // IAMs that have had their trigger(s) evaluated to true;
     //   This mean they have been added to the queue to display, or have already displayed
-    @NonNull final private Set<String> triggeredMessages;
+    @NonNull final private Set<String> dismissedAndToDismissMessages;
```
mikechoch: `dismissedAndToDismissMessages` should be `dismissedMessages`, no need to over complicate the name

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -21,7 +30,7 @@
     /**
      * Allows in-app messages to use multiple language variants, or to have variations between
      * different device types (ie. a different image for phones vs. tablets, etc.).
-     *
+     * <p>
```
mikechoch: Why is this `<p>` here?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -51,41 +57,44 @@
 
     @Nullable private static OSInAppMessageController sharedInstance;
     public static OSInAppMessageController getController() {
+        OneSignalDbHelper dbHelper = OneSignal.getDBHelperInstance();
+
         // Make sure only Android 4.4 devices and higher can use IAMs
         if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.JELLY_BEAN_MR2) {
-            sharedInstance = new OSInAppMessageDummyController();
+            sharedInstance = new OSInAppMessageDummyController(dbHelper);
```
mikechoch: Pass `null` into here instead, if possible

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
Jeasmine: reformat code stuff, i will remove it

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
Jeasmine: do you mean ArrayList<OSInAppMessage> messages?
i think is more clear to have separate responsibilities. But i can mix the logic to whenever the message is set add the data for redisplay,  although it will trigger another for loop. We can discuss this if you want

mikechoch: I am wondering why we have
`@NonNull private ArrayList<OSInAppMessage> messages;`
and
`@NonNull final private List<OSInAppMessage> displayedMessagesData;`

Shouldn't we just have one list of messages that we obtain from the `on_session` or cached IAMs?

Jeasmine: yeah i can mix the logics

mikechoch: I think we should have a single `@NonNull private ArrayList<OSInAppMessage> messages;` that contains on IAMs pulled down for the current session

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -34,24 +43,42 @@
     @NonNull
     public ArrayList<ArrayList<OSTrigger>> triggers;
 
+    //Last IAM display time in seconds
+    private long lastDisplayTime = -1;
+    //Current quantity of displays
+    private int displaysQuantity = 0;
+    //Quantity of displays limit
+    private int displaysLimit = Integer.MAX_VALUE;
```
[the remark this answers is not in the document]
mikechoch: Let's make display limit default to 1 since normal IAMs only show once?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -94,6 +102,14 @@ protected OSInAppMessageController() {
         );
         if (tempClickedMessageIdsSet != null)
             clickedClickIds.addAll(tempClickedMessageIdsSet);
+
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.JELLY_BEAN_MR2) {
```
jkasten2: This if would never be true as we new up a `OSInAppMessageDummyController` instead for per 4.4 devices.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
```diff
@@ -153,6 +157,97 @@ public void testBuiltMessageVariants() {
...
+
+    @Test
+    public void testBuiltMessageRedisplayLimit() throws JSONException {
+        OSTestInAppMessage message = InAppMessagingHelpers.buildTestMessageWitRedisplay(
+                LIMIT,
+                DELAY
+        );
+
+        for (int i = 0; i < LIMIT; i++) {
+            assertTrue( message.shouldDisplayAgain());
```
jkasten2: nit, remove space after (

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -303,17 +323,48 @@ void onFailure(int statusCode, String response, Throwable throwable) {
...
+
+        int index = displayedMessagesData.indexOf(message);
+        if (index != -1) {
+            OneSignal.onesignalLog(OneSignal.LOG_LEVEL.DEBUG, "setDataForReDisplay: " + message.messageId);
+
+            OSInAppMessage savedIAM = displayedMessagesData.get(index);
+            message.setLastDisplayTime(savedIAM.getLastDisplayTime());
+            message.setDisplayQuantity(savedIAM.getDisplayQuantity());
+
+            if (message.isDelayTimeSatisfied() && message.shouldDisplayAgain()) {
```
jkasten2: Here we need to add a check to only re-display if a trigger was updated that is in this IAM's trigger set.
This prevents the IAM from reshowing over and over until it hits it's display limit since `evaluateInAppMessages` is called each time an IAM is dismissed.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
jkasten2: @mikechoch and I talked about this. Our `ArrayList<OSInAppMessage> messages` is a list of all IAMs from our server so we should keep this list untouched.

To read and write our two new properties I think we should create a new class to store them under. We can call this something like `OSInAppMessageDisplayStats` which will hold our `displayQuantity` and `lastTimeDisplayed`. This can still be referenced by `OSInAppMessage` as a property.

Jeasmine: ok that make sense. I agree with that the message list should kept untouched.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
Jeasmine: true!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -303,17 +325,47 @@ void onFailure(int statusCode, String response, Throwable throwable) {
...
     }
 
+    /**
+     * In order to an IAM to be re display, two condition need to be satisfied
+     *  - Gap between displays
+     *  - Quantity of displays
+     *
+     * For re display, the message need to be removed from the arrays that track the display
+     * */
+    private void setDataForReDisplay(OSInAppMessage message) {
```
mikechoch: `ReDisplay` should be `Redisplay`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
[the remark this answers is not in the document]
Jeasmine: taking https://github.com/OneSignal/engineering/pull/92/files as reference, it says that if limit is missing make it MAX_INT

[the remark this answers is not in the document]
mikechoch: @jkasten2 Did we decide to keep this `MAX_INT` as default?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -105,8 +159,16 @@ JSONObject toJSONObject() {
...
             }
 
-            json.put("variants", variants);
-            json.put("display_duration", this.displayDuration);
+            json.put(IAM_VARIANTS, variants);
+            json.put(DISPLAY_DURATION, this.displayDuration);
+            json.put(DISPLAY_LIMIT, this.displayLimit);
+            json.put(DISPLAY_DELAY, this.displayDelay);
+
+            if (redisplayEnabled)
```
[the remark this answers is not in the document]
mikechoch: Agree with this

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -553,6 +587,7 @@ public void messageTriggerConditionChanged() {
      */
     void addTriggers(Map<String, Object> newTriggers) {
         triggerController.addTriggers(newTriggers);
+        checkTriggerChanges(newTriggers);
```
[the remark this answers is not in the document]
Jeasmine: ok, i had doubts about this

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -94,6 +106,22 @@ protected OSInAppMessageController() {
...
         if (tempClickedMessageIdsSet != null)
             clickedClickIds.addAll(tempClickedMessageIdsSet);
+
+        initRedisplayData(dbInstance);
+    }
+
+    /*
+    * For testing purposes
+    * */
+    public DateGenerator getDateGenerator() {
```
jkasten2: We can clean up this `DateGenerator` by using  `System.currentTimeMillis()`  instead.  This way we can use `advanceSystemTimeBy` in our tests for mocking time.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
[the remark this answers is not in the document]
jkasten2: Yes, we can use `MAX_INT` incase the backend passes back a `redisplay` section but without an params. Maybe is just confusing as it is a default on the OSInAppMessage class instead of being defaulted in when parsing.

@Jeasmine @mikechoch What do you think about putting `displaysLimit` as `1` for this class default and using the `Integer.MAX_VALUE` if `redisplay` is present without any `limit` key in the parsing? Thinking this will make this more clear

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
Jeasmine: oh men, why i didn't see this

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -110,7 +110,7 @@ protected OSInAppMessageController(OneSignalDbHelper dbInstance) {
         initRedisplayData(dbInstance);
     }
 
-    void initRedisplayData(OneSignalDbHelper dbInstance) {
+    protected void initRedisplayData(OneSignalDbHelper dbInstance) {
```
mikechoch: Change `dbInstance` to `dbHelper`, most of our codebase is `dbHelper`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -402,7 +402,7 @@ private void setDataForRedisplay(OSInAppMessage message) {
     private void queueMessageForDisplay(@NonNull OSInAppMessage message) {
         synchronized (messageDisplayQueue) {
             // Make sure no message is ever added to the queue more than once
-            if (!isInAppMessageQueued(message)) {
```
mikechoch: Using contains with an oObject is dangerous. This could be true or false and not represent the `String messageId`, this should be changed back

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -467,8 +456,7 @@ void messageWasDismissed(@NonNull OSInAppMessage message) {
     private void dismissCurrentMessage() {
         synchronized (messageDisplayQueue) {
             if (messageDisplayQueue.size() > 0) {
-                String removedMessageId = messageDisplayQueue.get(0).messageId;
-                messageDisplayQueue.remove(0);
+                String removedMessageId = messageDisplayQueue.remove(0).messageId;
```
mikechoch: Nice clean up!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
Jeasmine: Yeah agree, thas why i have the equals and hash overriden

PR #931, opened 2020-01-28
Added null check to fix `NullPointerException` being thrown

mikechoch:
* getRootWindowInsets() from `Activity` `DecorView` is null some cases (most likely when opening the app from cold start)
* null check will now return early with the decorView height instead of the difference of the height, top inset, and bottom inset

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/931)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSViewUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSViewUtils.java
@@ -11,6 +11,7 @@ import android.support.annotation.NonNull;
 import android.util.DisplayMetrics;
 import android.view.View;
 import android.view.Window;
+import android.view.WindowInsets;
 
 import java.lang.ref.WeakReference;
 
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSViewUtils.java
```diff
@@ -91,11 +91,12 @@ else  if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP)
     private static int getWindowHeightAPI23Plus(@NonNull Activity activity) {
         View decorView = activity.getWindow().getDecorView();
         // Use use stable heights as SystemWindowInset subtracts the keyboard
-        int heightWithoutKeyboard =
-           decorView.getHeight() -
-           decorView.getRootWindowInsets().getStableInsetBottom() -
-           decorView.getRootWindowInsets().getStableInsetTop();
-        return heightWithoutKeyboard;
+        if (decorView.getRootWindowInsets() == null)
```
jkasten2: Let's save `decorView.getRootWindowInsets()` into a variable just in case it goes `null` in between calls.

mikechoch: Very true, thanks

PR #942, opened 2020-01-30
IAM still show form the queue when IAM is pasued (Reported from Unity)

mikechoch:
* IAM will show form the queue even when IAM is paused
* Must wrap the check around the display action not the action to add to the queue
* When IAM is paused adding to the queue works now and displaying will be prevented instead

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/942)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -378,11 +379,14 @@ public void displayMessage(@NonNull final OSInAppMessage message) {
...
             @Override
             void onFailure(int statusCode, String response, Throwable throwable) {
+                inAppMessageShowing = false;
+                evaluateInAppMessages();
                 printHttpErrorForInAppMessageRequest("html", statusCode, response);
             }
 
             @Override
             void onSuccess(String response) {
+                inAppMessageShowing = true;
```
jkasten2: This isn't needed here.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -303,63 +308,42 @@ void onFailure(int statusCode, String response, Throwable throwable) {
...
-        }
-
-        queueMessageForDisplay(message);
-    }
-
     // Message has passed triggers and de-duplication logic.
     // Display message now or add it to the queue to be displayed.
     private void queueMessageForDisplay(@NonNull OSInAppMessage message) {
         synchronized (messageDisplayQueue) {
             messageDisplayQueue.add(message);
```
jkasten2: Need to check we don't add the same message more than once

mikechoch: Got it

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
mikechoch: Removing

PR #951, opened 2020-02-10
Remove IAM from db

Jeasmine:
* Remove IAMs with redisplay from DB when lastDisplay is
    older than six month

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/951)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
@@ -30,6 +30,7 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
 
     public static final String IN_APP_MESSAGES_JSON_KEY = "in_app_messages";
     private static final String OS_SAVE_IN_APP_MESSAGE = "OS_SAVE_IN_APP_MESSAGE";
+    private static final String OS_DELETE_IN_APP_MESSAGE = "OS_DELETE_IN_APP_MESSAGE";
 
     OSTriggerController triggerController;
     private OSSystemConditionController systemConditionController;
@@ -49,7 +50,8 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
     // Ordered IAMs queued to display, includes the message currently displaying, if any.
     @NonNull final ArrayList<OSInAppMessage> messageDisplayQueue;
     // IAMs displayed with last displayed time and quantity of displays data
-    @NonNull private List<OSInAppMessage> displayedMessagesData;
+    // This is retrieved from a DB Table that take care of each object to be unique
+    @NonNull private List<OSInAppMessage> redisplayedInAppMessages;
 
     private boolean inAppMessagingEnabled = true;
 
@@ -108,9 +110,9 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
 
     void initRedisplayData(OneSignalDbHelper dbInstance) {
         inAppMessageRepository = new OSInAppMessageRepository(dbInstance);
-        displayedMessagesData = inAppMessageRepository.getAllInAppMessages();
+        redisplayedInAppMessages = inAppMessageRepository.getRedisplayedInAppMessages();
 
-        OneSignal.Log(OneSignal.LOG_LEVEL.DEBUG, "displayedMessagesData: " + displayedMessagesData.toString());
+        OneSignal.Log(OneSignal.LOG_LEVEL.DEBUG, "redisplayedInAppMessages: " + redisplayedInAppMessages.toString());
     }
 
     // Normally we wait until on_session call to download the latest IAMs
@@ -144,6 +146,15 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
         OneSignalPrefs.saveString(OneSignalPrefs.PREFS_ONESIGNAL,
            OneSignalPrefs.PREFS_OS_CACHED_IAMS, json.toString());
         processInAppMessageJson(json);
+
+        // Remove cache redisplayed messages
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Thread.currentThread().setPriority(Process.THREAD_PRIORITY_BACKGROUND);
+                inAppMessageRepository.deleteOldRedisplayedInAppMessages();
+            }
+        }, OS_DELETE_IN_APP_MESSAGE).start();
     }
 
     private void processInAppMessageJson(@NonNull JSONArray json) throws JSONException {
@@ -282,12 +293,14 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
             return;
 
         final String clickId = action.clickId;
+        // If IAM has redisplay the clickId may be available
         boolean clickAvailableByRedisplay = message.getDisplayStats().isRedisplayEnabled() && message.isClickAvailable(clickId);
 
         // Never count multiple clicks for the same click UUID unless that click is from an IAM with redisplay
         if (!clickAvailableByRedisplay && clickedClickIds.contains(clickId))
             return;
         clickedClickIds.add(clickId);
+        // Track clickId per IAM
         message.addClickId(clickId);
 
         try {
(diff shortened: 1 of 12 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -154,16 +155,26 @@ void receivedInAppMessageJson(@NonNull JSONArray json) throws JSONException {
         // Cache copy for quick cold starts
         OneSignalPrefs.saveString(OneSignalPrefs.PREFS_ONESIGNAL,
            OneSignalPrefs.PREFS_OS_CACHED_IAMS, json.toString());
-        processInAppMessageJson(json);
+        ArrayList<OSInAppMessage> newMessages = processInAppMessageJson(json);
+
+        List<OSInAppMessage> updatedRedisplayMessages = inAppMessageRepository.updateInAppMessage(newMessages, redisplayInAppMessages);
```
jkasten2: The source of `newMessages` is driven directly by the results from the  `on_session` network call . However the results from `on_session` is drive by the backend and if the player fits into the Segments defined there. So this means the same IAM could fall in and out of Segment multiple times which would result in the on_session not always have it either.

So we don't reset the display stats noted above we should only delete these local records after a long period of time to be safe or not at all. Can we set this to 6 months to be safe?

Jeasmine: oh ok, i will address this on here and iOS. 6 Months seems like a logic time period

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -154,16 +155,26 @@ void receivedInAppMessageJson(@NonNull JSONArray json) throws JSONException {
...
         ArrayList<OSInAppMessage> newMessages = new ArrayList<>();
         for (int i = 0; i < json.length(); i++) {
             JSONObject messageJson = json.getJSONObject(i);
             OSInAppMessage message = new OSInAppMessage(messageJson);
             newMessages.add(message);
         }
+        return newMessages;
+    }
+
+    private void setMessages( ArrayList<OSInAppMessage> newMessages) {
```
mikechoch: Remove space before `ArrayList`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -472,14 +483,14 @@ public void run() {
 
         //Update the data to enable future re displays
```
mikechoch: Add a space after `//`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -472,14 +483,14 @@ public void run() {
...
         //Avoid calling the repository data again
```
mikechoch: Add a space after `//`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -572,7 +583,7 @@ public void messageTriggerConditionChanged() {
      * */
```
mikechoch: `* */` should just be `*/`

OneSignalSDK/unittest/src/test/java/com/test/onesignal/TestHelpers.java
```diff
@@ -373,6 +398,49 @@ else if (type == Cursor.FIELD_TYPE_FLOAT)
       return iams;
    }
 
+   @WorkerThread
+   static void deleteOldInAppMessages() {
```
jkasten2: This is duplicated, it exists in `com.onesignal.OSInAppMessageRepository` as well. Also I don't see this being called in the tests so I think we can remove this `deleteOldInAppMessages` method here.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/TestHelpers.java
```diff
@@ -373,6 +398,49 @@ else if (type == Cursor.FIELD_TYPE_FLOAT)
...
+            String idToRemove = cursor.getString(cursor.getColumnIndex(OneSignalPackagePrivateHelper.InAppMessageTable.COLUMN_NAME_MESSAGE_ID));
+            deleteInAppMessage(idToRemove);
+         } while (cursor.moveToNext());
+      }
+
+      if (!cursor.isClosed())
+         cursor.close();
+   }
+
+   static void deleteInAppMessage(String messageId) {
```
jkasten2: This is duplicated, it exists in `com.onesignal.OSInAppMessageRepository` as well. Also I don't see this being called in the tests so I think we can remove this `deleteInAppMessage` method here.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageRepository.java
```diff
@@ -15,12 +16,40 @@
...
     OSInAppMessageRepository(OneSignalDbHelper dbHelper) {
         this.dbHelper = dbHelper;
     }
 
+    /**
+     * Remove IAMs that the last display time was six month ago
+     */
+    @WorkerThread
+    void deleteOldInAppMessages(long dateInSeconds) {
+        SQLiteDatabase readableDb = dbHelper.getReadableDbWithRetries();
```
jkasten2: We could use a writable db here and performance a single write. Instead getting a list of ids then doing a delete on each record one at a time.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/TestHelpers.java
Jeasmine: good catch, this was used before but I changed it. TY!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageRepository.java
Jeasmine: Absolutely right

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -144,6 +145,15 @@ void receivedInAppMessageJson(@NonNull JSONArray json) throws JSONException {
...
         processInAppMessageJson(json);
+
+        // Remove cache redisplayed messages
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Thread.currentThread().setPriority(Process.THREAD_PRIORITY_BACKGROUND);
+                inAppMessageRepository.deleteOldRedisplayedInAppMessages();
+            }
+        }, OS_SAVE_IN_APP_MESSAGE).start();
```
mikechoch: Change the tag associated with the thread, this is not a save action is it?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageDisplayStats.java
```diff
@@ -87,13 +87,10 @@ boolean shouldDisplayAgain() {
...
     boolean isDelayTimeSatisfied() {
-        long currentTimeInSeconds = System.currentTimeMillis() / 1000;
-
         if (lastDisplayTime < 0) {
-            lastDisplayTime = currentTimeInSeconds;
             return true;
         }
-
+        long currentTimeInSeconds = System.currentTimeMillis() / 1000;
         //Calculate gap between display times
```
mikechoch: Space after `//`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageRepository.java
```diff
@@ -15,32 +15,24 @@
...
-            if (writableDb != null) {
-                try {
-                    writableDb.endTransaction(); // May throw if transaction was never opened or DB is full.
-                } catch (Throwable t) {
-                    OneSignal.Log(OneSignal.LOG_LEVEL.ERROR, "Error closing transaction! ", t);
-                }
-            }
-        }
+        writableDb.delete(OneSignalDbContract.InAppMessageTable.TABLE_NAME,
+                OneSignalDbContract.InAppMessageTable.COLUMN_NAME_LAST_DISPLAY + "<?",
```
mikechoch: Space between `<?` -> `< ?`

PR #978, opened 2020-03-24
Workaround issue in Lenovo devices

khang-tiki:
Please reference #857 for more details.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/978)


jkasten2:
@khang-tiki Thanks for the PR. Putting the catch as you did in the `isRooted()` method is a good spot, along with a good comment.

It is unfortunate that these Android 6 Lenovo devices have this bug. I wonder if they were trying to prevent root detection or some kind of other issues with the static array and number of elements in it.

PR #986, opened 2020-04-07
Remove dependencies from gradle

Jeasmine:
* Remove appcompat-v4 dependency
  * Change from runtime to compileOnly appcompat-v7 dependency
  * Add log for when no location permissions added on Manifest

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/986)

### DIFF
--- a/OneSignalSDK/onesignal/build.gradle
+++ b/OneSignalSDK/onesignal/build.gradle
@@ -61,9 +61,11 @@ dependencies {
     // Also note, firebase & gms libraries use android.support:26.
     //   - They never refer to 27 or 28
     api 'com.android.support:cardview-v7:[26.0.0, 27.99.99]'
-    api 'com.android.support:support-v4:[26.0.0, 27.99.99]'
+    api 'com.android.support:support-fragment:[26.0.0, 27.99.99]'
     api 'com.android.support:customtabs:[26.0.0, 27.99.99]'
-    api 'com.android.support:appcompat-v7:[26.0.0, 27.99.99]'
+
+    // compileOnly as this is just for fallback code if AppCompatActivity wasn't added to the project.
+    compileOnly 'com.android.support:appcompat-v7:26.1.0'
 }
 
 apply from: 'maven-push.gradle'
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSystemConditionController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSystemConditionController.java
@@ -30,9 +30,13 @@ class OSSystemConditionController {
             return false;
         }
 
-        if (isDialogFragmentShowing(ActivityLifecycleHandler.curActivity)) {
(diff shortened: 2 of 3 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSystemConditionController.java
```diff
@@ -30,9 +30,13 @@ boolean systemConditionsAvailable() {
...
-        if (isDialogFragmentShowing(ActivityLifecycleHandler.curActivity)) {
-            OneSignal.onesignalLog(OneSignal.LOG_LEVEL.WARN, "OSSystemConditionObserver dialog fragment detected");
-            return false;
+        try {
+            if (isDialogFragmentShowing(ActivityLifecycleHandler.curActivity)) {
+                OneSignal.onesignalLog(OneSignal.LOG_LEVEL.WARN, "OSSystemConditionObserver dialog fragment detected");
+                return false;
+            }
+        } catch (NoClassDefFoundError exception) {
+            OneSignal.onesignalLog(OneSignal.LOG_LEVEL.ERROR, "isDialogFragmentShowing AppCompatActivity not imported either used: " + exception);
```
jkasten2, line 39: This can be an info, as it it isn't an error if  they don't have that class in their app.

OneSignalSDK/onesignal/build.gradle
```diff
@@ -61,9 +61,10 @@ dependencies {
     // Also note, firebase & gms libraries use android.support:26.
     //   - They never refer to 27 or 28
     api 'com.android.support:cardview-v7:[26.0.0, 27.99.99]'
-    api 'com.android.support:support-v4:[26.0.0, 27.99.99]'
```
jkasten2, line 64: Even if we remove `com.android.support:support-v4` from here it is still being included through other dependencies.

```
+--- com.google.firebase:firebase-messaging:[10.2.1, 17.3.99] -> 17.3.4
|    +--- com.google.android.gms:play-services-basement:16.0.1
|    |    \--- com.android.support:support-v4:26.1.0
```


By adding the following to your build.gradle you will observe errors in the OneSignal code base due to force removing this dependency.
```
configurations.all {
    exclude group: 'com.android.support', module: 'support-v4'
}
```

jkasten2: It looks like however we might be able to get away with ` api 'com.android.support:support-fragment:` instead if we want to switch to that. It seems to include everything we need.

jkasten2: One last note with dependencies, any OneSignal code that directly uses a library it should be top level in this file as the fundamentally matches up.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSSystemConditionController.java
jkasten2: @Jeasmine This line is still using `OneSignal.LOG_LEVEL.ERROR`. Should be INFO.

PR #993, opened 2020-04-18
WritePrefHandlerThread improvements

jkasten2:
* This add more `Throwables` on top of what PR #975 did as it turns out `InternalError` and `OutOfMemoryError` are not the only things `Thread.start` can throw based o the following stack traces from 3.13.1.
   - https://github.com/OneSignal/OneSignal-Android-SDK/issues/917#issuecomment-615464354
  - https://github.com/OneSignal/OneSignal-Android-SDK/issues/917#issuecomment-615873549
* Moved appContext null check before thread start
  - This is more efficient as we are not creating a thread before it is possible write
  - This eliminates entries points into Thread.start to help debug  issue #917
* Clean up: Removed synchronized on mHandler
  - Cleaned up synchronized on mHandler in both startDelayedWrite and tests
  - This was not needed as startDelayedWrite as a synchronized at the method level
  - Limited access level to mHandler and startDelayedWrite now that tests don't access this anymore directly.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/993)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
@@ -123,7 +123,7 @@ class OneSignalPrefs {
     }
 
     public static class WritePrefHandlerThread extends HandlerThread {
-        public Handler mHandler;
+        private Handler mHandler;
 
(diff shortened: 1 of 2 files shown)


mikechoch:
LGTM! Hopefully no other exceptions show up, but we will have to keep an eye out.

PR #994, opened 2020-04-21
Completely remove global static deviceType int

mikechoch:
* Better to use the OSUtils getDeviceType method since we wont know what part fo the SDK init we are in
* The chance of a 0 deviceType is a race condition we do not want to fight with

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/994)

### DIFF
--- a/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
+++ b/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
@@ -1,15 +1,55 @@
 package com.onesignal;
 
+import android.content.Context;
+
 import org.robolectric.annotation.Implements;
+
+import java.util.HashSet;
 
 @Implements(OSUtils.class)
 public class ShadowOSUtils {
-   static int deviceType = 1;
-   public static String carrierName = "test1";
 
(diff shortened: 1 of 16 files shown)


OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
```diff
@@ -1,23 +1,63 @@
...
+   public static void resetStatics() {
+      deviceType = 0;
+      carrierName = "test1";
+      subscribableStatus = 1;
+
+      availableClasses = new HashSet<>();
+   }
 
    public int getDeviceType() {
+      // Class only available on the FireOS, if this class exists we want to mock Amazon deviceType
```
jkasten2, line 45: This looks good for now, as `getDeviceType` gets more complex with more platforms we will probably want to not mock this method. But that can be done in the task where we add the new platform.

mikechoch: Yeah we would need a better way to mock a class existing within our unit tests and that way when the REAL getDeviceType method is called it just uses that functionality rathe than my `HashSet`

PR #1004, opened 2020-04-29
WIP - Accommodating the foreground handler changes and multiple handlers

mikechoch:
* Removed the AppNotificationExtensionService constructor and now a method auto detects the implemented interface handlers and creates calls the correct setters
  * Instance of the class is used to init the handlers
* Refactored OSInFocusDisplayOption to OSNotificationDisplayOption
* Added bubbling for the two foreground handlers
  * A flag inside of a completion handler allows bubbling from the ext handler to the app handler
  * Added a NotificationWillShowInForegroundCompletionHandler to track when the foreground handler is finished working or not ( controlled by the dev with a X second timeout)
* Removed the OSInFocusDisplay.IN_APP_ALERT enum
* Refactored OSNotificationOpenResult to OSNotificationOpenedResult

NOTE:
* The NotificationProcessingHandler interface was an abstract class but this is now broken and not working at all due to the service being removed form the AndroidManifest and the new goal is to move from a JobIntentService to a WorkManager

Add conflict commit message here.......

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1004)


mikechoch:
Done referencing this one to break it apart into small PRs. This is now completely invalid and being closed.

PR #1006, opened 2020-05-06
Add location permissions not available alert dialog

Jeasmine:
* Inform the user when using preview mode that location permissions are not configured

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1006)

Screenshot_20200506_142646_com onesignal example

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
@@ -27,10 +27,6 @@
 
 package com.onesignal;
 
-import com.google.android.gms.location.LocationListener;
-import com.google.android.gms.location.LocationRequest;
-import com.onesignal.AndroidSupportV4Compat.ContextCompat;
-
 import android.content.Context;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
@@ -43,7 +39,10 @@ import android.support.annotation.NonNull;
 
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.api.GoogleApiClient;
+import com.google.android.gms.location.LocationListener;
+import com.google.android.gms.location.LocationRequest;
 import com.google.android.gms.location.LocationServices;
+import com.onesignal.AndroidSupportV4Compat.ContextCompat;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
@@ -83,11 +82,11 @@ class LocationGMS {
    }
    interface LocationHandler {
       PermissionType getType();
-      void complete(LocationPoint point);
+      void onComplete(LocationPoint point);
    }
 
    abstract static class LocationPromptCompletionHandler implements LocationHandler {
-      void onAnswered(boolean accepted) {}
+      void onAnswered(OneSignal.PromptActionResult result) {}
    }
 
    private static ConcurrentHashMap<PermissionType, LocationHandler> locationHandlers = new ConcurrentHashMap<>();
@@ -135,7 +134,7 @@ class LocationGMS {
       }
    }
 
-   static void sendAndClearPromptHandlers(boolean promptLocation, boolean accepted) {
+   static void sendAndClearPromptHandlers(boolean promptLocation, OneSignal.PromptActionResult result) {
       if (!promptLocation) {
          OneSignal.onesignalLog(OneSignal.LOG_LEVEL.DEBUG, "LocationGMS sendAndClearPromptHandlers from non prompt flow");
          return;
@@ -144,7 +143,7 @@ class LocationGMS {
       synchronized (promptHandlers) {
          OneSignal.onesignalLog(OneSignal.LOG_LEVEL.DEBUG, "LocationGMS calling prompt handlers");
          for (LocationPromptCompletionHandler promptHandler : promptHandlers) {
-            promptHandler.onAnswered(accepted);
+            promptHandler.onAnswered(result);
          }
          // We only call the prompt handlers once
          promptHandlers.clear();
@@ -174,7 +173,7 @@ class LocationGMS {
       locationHandlers.put(handler.getType(), handler);
 
       if (!OneSignal.shareLocation) {
-         sendAndClearPromptHandlers(promptLocation, false);
+         sendAndClearPromptHandlers(promptLocation, OneSignal.PromptActionResult.ERROR);
          fireFailedComplete();
          return;
       }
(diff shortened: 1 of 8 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -189,7 +194,7 @@ static void getLocation(Context context, boolean promptLocation, LocationHandler
 
       if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
          if (locationFinePermission != PackageManager.PERMISSION_GRANTED && locationCoarsePermission != PackageManager.PERMISSION_GRANTED) {
-            sendAndClearPromptHandlers(promptLocation,false);
+            sendAndClearPromptHandlers(promptLocation,false, context.getString(R.string.location_not_available_title), context.getString(R.string.location_not_available_message));
```
mikechoch, line 197: Add a space to this `promptLocation,false,`
`promptLocation, false,`

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -189,7 +194,7 @@ static void getLocation(Context context, boolean promptLocation, LocationHandler
```
mikechoch, line 194:
```
String alertDialogTitle = context.getString(R.string.location_not_available_title);
String alertDialogMessage = context.getString(R.string. location_not_available_message);
```
Pass these into both usages of
`sendAndClearPromptHandlers(promptLocation, false, alertDialogTitle, alertDialogMessage);`

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -223,11 +231,13 @@ else if (permissionList.contains("android.permission.ACCESS_COARSE_LOCATION")) {
                   sendAndClearPromptHandlers(promptLocation, true);
                   startGetLocation();
                } else {
-                  sendAndClearPromptHandlers(promptLocation, false);
+                  String alertDialogTitle = context.getString(R.string.location_not_available_title);
+                  String alertDialogMessage = context.getString(R.string. location_not_available_message);
```
mikechoch, line 235: You use these `Strings` in two places just move them up top outside of the first if statement

mikechoch: Remove space `R.string. location_not_available_message);`
`R.string.location_not_available_message);`

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -187,9 +192,13 @@ static void getLocation(Context context, boolean promptLocation, LocationHandler
          locationCoarse = true;
       }
 
+      String alertDialogTitle = context.getString(R.string.location_not_available_title);
+      String alertDialogMessage = context.getString(R.string.location_not_available_message);
+
       if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
          if (locationFinePermission != PackageManager.PERMISSION_GRANTED && locationCoarsePermission != PackageManager.PERMISSION_GRANTED) {
```
jkasten2, line 199: It isn't clear here by the code but it can only fall under this if block if permission were missing from the `AndroidManifest.xml`

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -223,11 +232,11 @@ else if (permissionList.contains("android.permission.ACCESS_COARSE_LOCATION")) {
                   sendAndClearPromptHandlers(promptLocation, true);
                   startGetLocation();
                } else {
-                  sendAndClearPromptHandlers(promptLocation, false);
+                  sendAndClearPromptHandlers(promptLocation, false, alertDialogTitle, alertDialogMessage);
```
jkasten2, line 235: Careful, we don't want to display a message if the only issue was the user denied location permission.

OneSignalSDK/onesignal/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <string name="location_not_available_title">Location Not Available</string>
+    <string name="location_not_available_message">Looks like this app doesn\'t have location permissions granted</string>
```
jkasten2, line 5: Granted isn't correct. Permission isn't present what we want. We don't want to show any message if the permission was simply declined.

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -187,9 +192,13 @@ static void getLocation(Context context, boolean promptLocation, LocationHandler
          locationCoarse = true;
       }
 
+      String alertDialogTitle = context.getString(R.string.location_not_available_title);
```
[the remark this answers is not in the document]
Jeasmine: ok, agree with this!

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
Jeasmine: This flow is added because the user might forget to add the permissions on Manifest, and location prompting will not work, because of that. So we are only showing this alert when the user is prompting from a preview.

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
Jeasmine: good catch!

OneSignalSDK/onesignal/src/main/res/values/strings.xml
Jeasmine: Good catch, this string should be the one that is on the screenshot, that one is the one that Lee defined

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
```diff
@@ -87,7 +88,7 @@
    }
 
    abstract static class LocationPromptCompletionHandler implements LocationHandler {
-      void onAnswered(boolean accepted) {}
+      void onAnswered(@Nullable String messageTitle, @Nullable String message, boolean accepted) {}
```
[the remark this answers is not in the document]
Jeasmine: onComplete handles to many flows, for prompting we only want to know what the user answered to the prompt, or if the permissions were already granted, we don't care about the location perse.

OneSignalSDK/onesignal/src/main/java/com/onesignal/LocationGMS.java
jkasten2: Sorry didn't finish my thought here. That isn't clear so we should add a comment.

The code originally should have had a comment about that for now it seems it needs it even more.

Jeasmine: Let me know if the changes make sense, in order to make it on iOS

PR #1009, opened 2020-05-07
Add Settings option to denied location prompt

Jeasmine:
* Whenever the user clicks on never ask again option and denies the permission,
     we will show the reason why we need it and the option to open settings and change the location permission

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1009)

Screenshot_20200507_191734_com onesignal example

edited-20200507-193302

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/PermissionsActivity.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/PermissionsActivity.java
@@ -28,8 +28,11 @@
 package com.onesignal;
 
 import android.app.Activity;
+import android.app.AlertDialog;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
@@ -44,10 +47,11 @@ public class PermissionsActivity extends Activity {
    // Default animation duration in milliseconds
    private static final int DELAY_TIME_CALLBACK_CALL = 500;
    private static final int REQUEST_LOCATION = 2;
+   private static final int REQUEST_SETTINGS = 3;
 
-   static boolean waiting, answered;
+   static boolean waiting, answered, fallbackToSettings, neverAskAgainClicked;
    private static ActivityLifecycleHandler.ActivityAvailableListener activityAvailableListener;
-   
+
    @Override
    protected void onCreate(Bundle savedInstanceState) {
       super.onCreate(savedInstanceState);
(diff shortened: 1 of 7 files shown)


Jeasmine:
@leemunroe I will need your help to define the text for this case. This flow is for when the user denies location permission and choose "never ask again". So we want to show the option to go to settings and explain why.

OneSignalSDK/onesignal/src/main/res/values/strings.xml
```diff
@@ -4,4 +4,6 @@
     <string name="location_not_available_title">Location Not Available</string>
     <string name="location_not_available_message">Looks like this app doesn\'t have location permissions granted</string>
 
+    <string name="location_not_available_open_settings">We need location for user .....complete text</string>
```
mikechoch, line 7: We just need the correct text for this and should be good to go!
Maybe rename to `location_not_available_open_settings_message`?

OneSignalSDK/onesignal/src/main/res/values/strings.xml
```diff
@@ -4,4 +4,6 @@
...
+    <string name="settings">Settings</string>
```
mikechoch, line 8: `settings` is a little unclear, maybe lets rename to 
`location_not_available_open_settings_title`?

OneSignalSDK/onesignal/src/main/java/com/onesignal/PermissionsActivity.java
```diff
@@ -104,10 +109,16 @@ public void onRequestPermissionsResult(final int requestCode, @NonNull String pe
...
                boolean granted = grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED;
                LocationGMS.sendAndClearPromptHandlers(true, granted);
-               if (granted)
+               if (granted) {
                   LocationGMS.startGetLocation();
-               else
+               } else {
+                  if (fallbackToSettings &&
+                          neverAskAgainClicked &&
+                          !ActivityCompat.shouldShowRequestPermissionRationale(PermissionsActivity.this, LocationGMS.requestPermission)) {
```
mikechoch, line 117: Maybe we should move this if check into a separate method to clean up this long `if-statement`, something like below?
```
if (shouldShowLocationPermissionSettings)
    showLocationPermissionSettings();
```
**OR**
```
private void attemptToShowLocationPermissionSettings() {
    if (fallbackToSettings 
            && neverAskAgainClicked 
            && !ActivityCompat.shouldShowRequestPermissionRationale(PermissionsActivity.this, LocationGMS.requestPermission))
        showLocationPermissionSettings();
}
```

mikechoch:
Did some physical testing to wrap this one up... One thing I notice is that when you get the Settings prompt to show it shows above the next IAM in the queue. This is not a blocker or issue, but are we okay with this functionality?

PR #1010, opened 2020-05-08
Refactor prep for `InFocusDisplayOption` in `NotifForegroundHandler`

mikechoch:
* `NotificationWillShowInForegroundHandler` will be responsible for setting the notifications display type if a dev decides so
  * Alternative is a global setter that also controls the default for `OSNotificationDisplay` if no `NotificationWillShowInForegroundHandler` is set
  * Renamed `None` to `SILENT` and `Notification` to `NOTIFICATION` to follow a caps snake-case enum convention
* Removed `InAppAlert` from `InFocusDisplayOption`
  * Refactored `InFocusDisplayOption` to `OSNotificationDisplay`
  * Cleaned out `InAppAlert` related `UnitTests`

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -82,19 +82,27 @@ public class OneSignal {
    // If the app is this amount time or longer in the background we will count the session as done
    static final long MIN_ON_SESSION_TIME_MILLIS = 30 * 1_000L;
 
-   public enum LOG_LEVEL {
+    public enum LOG_LEVEL {
       NONE, FATAL, ERROR, WARN, INFO, DEBUG, VERBOSE
    }
 
-   public enum OSInFocusDisplayOption {
-      None, InAppAlert, Notification
+   public enum OSNotificationDisplay {
+      SILENT,
+      NOTIFICATION;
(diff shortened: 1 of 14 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -477,6 +487,18 @@ public static void unsubscribeWhenNotificationsAreDisabled(boolean set) {
       mUnsubscribeWhenNotificationsAreDisabled = set;
    }
 
+   /**
+    * If {@link NotificationWillShowInForegroundHandler} is not set, the ability to control a
+    *    default {@link OSNotificationDisplay} should still exist
+    */
+   public static void setNotificationDisplayOption(OSNotificationDisplay displayOption) {
```
jkasten2, line 494: Redundant APIs create another surface which has a lot of dependencies:
1. More SDK code and code paths
2. Unit Tests
3. Manual testing
4. Documentation & Examples

Not saying we would never want a redundant API, just it should have a good use case. In this case if they wanted to not show a notification when the app is in focus for some reason I can't see way they won't want to setup a notifiationWillShowInForegorund handler.

mikechoch:
Disregard broken UnitTest, fixed in PR off of this PR
#1014

PR #1019, opened 2020-05-15
Fixes rare IllegalThreadStateException from Thread.start()

jkasten2:
* The root issue is that `Thread.start()` could be called a 2nd time if
there was a problem getting a Handler after starting the thread.
Getting a Handler instance was used as the determining factor if the
thread had started or not.
* When calling `Thread.start()` there is no guarantee that the thread
has started after returning or that is may have already completed.
* To overcome the following 2 changes were made
  1. Override `HandlerThread`'s `onLooperPrepared`
     - Moving our `getLooper()` call here guarantees it will never be `null`
     - Then this means we can create a new `Handler` everytime successfully
  2. Added a boolean to make sure we only call `start()` once
     - `threadStartCalled` is set to `true` after `start()` is called.
     - This is needed as `onLooperPrepared` may not fire by the time some
       other part of the OneSignal SDK calls `scheduleFlushToDiskJob`
       again.

Relates to PR https://github.com/OneSignal/OneSignal-Android-SDK/pull/993
Fixes https://github.com/OneSignal/OneSignal-Android-SDK/issues/917

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1019)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
@@ -123,7 +123,7 @@ class OneSignalPrefs {
     }
 
     public static class WritePrefHandlerThread extends HandlerThread {
-        private Handler mHandler;
+        private @Nullable Handler mHandler;
 
         private static final int WRITE_CALL_DELAY_TO_BUFFER_MS = 200;
         private long lastSyncTime = 0L;
@@ -136,9 +136,11 @@ class OneSignalPrefs {
         protected void onLooperPrepared() {
             super.onLooperPrepared();
 
-            // Getting handler here as onLooperPrepared guarantees getLooper() will not return null
+            // Getting handler here as onLooperPrepared guarantees getLooper() will be non-null
             mHandler = new Handler(getLooper());
-            scheduleFlushToDiskJob();
+
+            // Kicks off our first flush, startDelayedWrite will schedule all flushes after that
+            scheduleFlushToDisk();
         }
 
         private synchronized void startDelayedWrite() {
@@ -148,7 +150,7 @@ class OneSignalPrefs {
                 return;
 
             startThread();
-            scheduleFlushToDiskJob();
+            scheduleFlushToDisk();
         }
 
         private boolean threadStartCalled;
@@ -160,7 +162,7 @@ class OneSignalPrefs {
             threadStartCalled = true;
         }
 
-        private synchronized void scheduleFlushToDiskJob() {
+        private synchronized void scheduleFlushToDisk() {
             // Could be null if looper thread just started
             if (mHandler == null)
                 return;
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
```diff
@@ -132,113 +132,52 @@
...
             //   if not available now later OneSignal.setContext will call this again.
             if (OneSignal.appContext == null)
                 return;
 
-            if (mHandler == null) {
-                startThread();
-                mHandler = new Handler(getLooper());
-            }
+            startThread();
+            scheduleFlushToDiskJob();
```
Jeasmine, line 151: do you still need this call, cant you depend only on onLooperPrepared?

jkasten2: `onLooperPrepared` will take care of the first flush. For 2nd, 3rd, etc calls we have to call `scheduleFlushToDiskJob` again to setup a new delayed write.

Since you had this question is there anything I can change to make this less confusing? Maybe just a comment here?

jkasten2: @Jeasmine Fixed!

Pavneet-Sing:
This is a nice approach to fix the issue and possible causes. I hope this will handle all the edge cases. 
I guess [this indicates uncertainty](https://github.com/OneSignal/OneSignal-Android-SDK/blob/0ba8308d796f39cefb923d5d1282a93adb43c076/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java#L164) and can cause unexpected behavior if this ever happens(like user can use different threads, one for initialization and another for saving preference, just wondering 🤔) so how about refactoring this with executor-service without handling all the checks?  [now can use `mHandler = getThreadHandler()l`](https://github.com/OneSignal/OneSignal-Android-SDK/blob/0ba8308d796f39cefb923d5d1282a93adb43c076/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java#L140)

jkasten2:
@Pavneet-Sing Thanks for the review as well! Based on all the catches we added to `Thread.start()` we ruled it out completely. This then pointed me to look into how `getLooper()` works. After digging into the Android ASOP implementation on this I noticed it may return `null` which would case an issue with `new Handler`.  I then found `onLooperPrepared` which I discovered runs later which means `getLooper()` would guarantee to be never `null`.

So `scheduleFlushToDiskJob` will be kicked off the first time from `onLooperPrepared`, and all other times from (possibility any thread) from `startDelayedWrite`. The `mHandler == null` check and the `synchronized` should prevent any possible edge cases.

Good point on the `ExecutorService`, I didn't think of that. If there are no further issue we plan to keep this for now. As we migrate to Kotlin we may move this to a coroutine as this would most likely be even simplier.

Pavneet-Sing:
@jkasten2 Yeah, Nice work there. The implementation should work to prevent the crashes though I was just wondering about edge cases, As you've mentioned, `synchronized` should prevent any possible edge cases.

The probably of `getLooper()` returning `null` is leading to all the `null` checks though I love the idea of migrating to coroutines though didn't suggest it cause it will require changes in the whole file or most likely changes in the whole SDK as well. May the force be with you :) 👍

PR #1022, opened 2020-05-19
Added HMS Notif Open Activity

jkasten2:
* This is the entry point when tapping on a HMS core notification
* This has to be it's own Activity separate from NotificationOpenedActivity
since we do not have full control over the Bundle

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1022)

### DIFF
--- a/OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
+++ b/OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
@@ -1,0 +1,125 @@
+package com.test.onesignal;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.support.annotation.NonNull;
+
+import com.onesignal.NotificationOpenedActivityHMS;
+import com.onesignal.OneSignal;
+import com.onesignal.ShadowCustomTabsClient;
+import com.onesignal.ShadowCustomTabsSession;
+import com.onesignal.ShadowOSUtils;
+import com.onesignal.ShadowOSViewUtils;
+import com.onesignal.ShadowOSWebView;
+import com.onesignal.ShadowOneSignalRestClient;
+import com.onesignal.StaticResetHelper;
+import com.onesignal.example.BlankActivity;
+
+import org.json.JSONObject;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.Robolectric;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.RuntimeEnvironment;
+import org.robolectric.annotation.Config;
+import org.robolectric.shadows.ShadowLog;
+
+import java.util.UUID;
+
+import static com.onesignal.InAppMessagingHelpers.ONESIGNAL_APP_ID;
+import static com.test.onesignal.TestHelpers.fastColdRestartApp;
+import static com.test.onesignal.TestHelpers.threadAndTaskWait;
+import static junit.framework.Assert.assertEquals;
+import static org.robolectric.Shadows.shadowOf;
+
+
+@Config(packageName = "com.onesignal.example",
+    instrumentedPackages = { "com.onesignal" },
+    shadows = {
+        ShadowOSUtils.class,
+        ShadowOneSignalRestClient.class,
+        ShadowCustomTabsClient.class,
+        ShadowOSWebView.class,
+        ShadowOSViewUtils.class,
+        ShadowCustomTabsClient.class,
+        ShadowCustomTabsSession.class
+    },
+    sdk = 26
+)
+
+
+@RunWith(RobolectricTestRunner.class)
+public class NotificationOpenedActivityHMSIntegrationTestsRunner {
+
+    @BeforeClass // Runs only once, before any tests
+    public static void setUpClass() throws Exception {
+        ShadowLog.stream = System.out;
+        TestHelpers.beforeTestSuite();
+        StaticResetHelper.saveStaticValues();
+    }
+
+    @Before
+    public void beforeEachTest() throws Exception {
+        TestHelpers.beforeTestInitAndCleanup();
+    }
+
+    private static Intent helper_baseHMSOpenIntent() {
+        return new Intent()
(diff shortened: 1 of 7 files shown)


OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
```diff
@@ -0,0 +1,125 @@
...
+        ShadowOneSignalRestClient.class,
+        ShadowCustomTabsClient.class,
+        ShadowOSWebView.class,
+        ShadowOSViewUtils.class,
+        ShadowCustomTabsClient.class,
+        ShadowCustomTabsSession.class
+    },
+    sdk = 26
+)
+
```
Jeasmine, line 51: can we remove this enters? or are we changing the convention?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Before
+    public void beforeEachTest() throws Exception {
+        TestHelpers.beforeTestInitAndCleanup();
+    }
+
+    private static Intent helper_baseHMSOpenIntent() {
+        return new Intent()
+                .setFlags(Intent.FLAG_ACTIVITY_NO_HISTORY | Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET | Intent.FLAG_ACTIVITY_MULTIPLE_TASK | Intent.FLAG_ACTIVITY_NEW_TASK)
```
Jeasmine, line 70: FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET is deprecated, and why is needed?

Jeasmine, line 70: this combination Intent.FLAG_ACTIVITY_MULTIPLE_TASK | Intent.FLAG_ACTIVITY_NEW_TASK is forcing to always create a new task, can this lead to having 2 instances at the same time?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
```diff
@@ -0,0 +1,125 @@
...
+
+    // Since the Activity has to be public it could be started outside of a OneSignal flow.
+    // Ensure it doesn't crash the app.
+    @Test
+    public void emptyIntent_doesNotThrow() {
+        helper_startHMSOpenActivity(helper_baseHMSOpenIntent());
+    }
+
+    @Test
+    public void barebonesOSPayload_startsMainActivity() throws Exception {
```
Jeasmine, line 86: this class should test everything around notifications? how will this operate with notifications tracking etc? should we test that?

mikechoch:
LGTM, just needs comments fixed

OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
mikechoch: He explains the reason for these flags in the comment in this PR
```
// HMS Core creates a notification with an Intent when opened to start this Activity.
//   Intent is defined via OneSignal's backend and is sent to HMS.
// This has to be it's own Activity separate from NotificationOpenedActivity since
//   we do not have full control over the Bundle.
// Designed to be started with these flags
// Intent.FLAG_ACTIVITY_NO_HISTORY | Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET | Intent.FLAG_ACTIVITY_MULTIPLE_TASK | Intent.FLAG_ACTIVITY_NEW_TASK
// This way if app developer does not want the app to launch then it won't do so
```

mikechoch: idem

OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
mikechoch: This is only for HMS notifications, but still good question bringing up tracking @Jeasmine

OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
Jeasmine: yeah I saw that but was curious if there were other reason because that flag is deprecated and in the docs said to use FLAG_ACTIVITY_NEW_DOCUMENT but doesn't make sense to me for this scenario 🤔

jkasten2: In testing found out that `FLAG_ACTIVITY_NEW_TASK | FLAG_ACTIVITY_MULTIPLE_TASK` provides the requirement of "Allows the app developer to omit foregrounding the app if they choose".

jkasten2: In testing found out that `FLAG_ACTIVITY_NEW_TASK | FLAG_ACTIVITY_MULTIPLE_TASK` provides the requirement of "Allows the app developer to omit foregrounding the app if they choose".

We always want to create a new task so it doesn't resume some existing task. If it resumes an existing task then it won't be possible to prevent the app from showing. Some developer want to process the click in the background without bring the app to the foreground.

jkasten2: I also added even more detail to the comment at the top of `NotificationOpenedActivityHMS.java` on the requirements and what does not work and why for future reference incase of a refactor of these flags.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/NotificationOpenedActivityHMSIntegrationTestsRunner.java
Jeasmine: Agree with having test for all platforms as a generic way!

PR #1023, opened 2020-05-19
Added HMS Notif Open Activity

jkasten2:
* This is the entry point when tapping on a HMS core notification
* This has to be it's own Activity separate from NotificationOpenedActivity
since we do not have full control over the Bundle

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1023)


jkasten2:
Duplicated of #1022

PR #1026, opened 2020-05-21
HMS device_type detection and HMS token registration 

jkasten2:
* Added Huawei as device_type 13
* Added detecting if HMS exists and only use it if
Google play services is not on the device and is not enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1026)

### DIFF
--- a/OneSignalSDK/onesignal/build.gradle
+++ b/OneSignalSDK/onesignal/build.gradle
@@ -56,6 +56,10 @@ dependencies {
     // firebase-messaging:17.6.0 is the max version since we still have code looking for FirebaseInstanceIdService
     api 'com.google.firebase:firebase-messaging:[10.2.1, 17.3.99]'
 
+    // Huawei PushKit
+    // KEEP as "compileOnly", so OneSignal isn't a direct dependency in the POM file.
+    compileOnly 'com.huawei.hms:push:4.0.2.300'
+
     // Keep under 28 until we switch to AndroidX
     //   otherwise app can get dup classes between 26 & 28 when mixing these versions.
     // Also note, firebase & gms libraries use android.support:26.
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
@@ -14,17 +14,6 @@ import static com.onesignal.OSUtils.getResourceString;
 class GooglePlayServicesUpgradePrompt {
    private static final int PLAY_SERVICES_RESOLUTION_REQUEST = 9_000;
 
-   static boolean isGMSInstalledAndEnabled() {
-      try {
-         PackageManager pm = OneSignal.appContext.getPackageManager();
-         PackageInfo info = pm.getPackageInfo(GoogleApiAvailability.GOOGLE_PLAY_SERVICES_PACKAGE, PackageManager.GET_META_DATA);
-
-         return info.applicationInfo.enabled;
-      } catch (PackageManager.NameNotFoundException e) {}
-
-      return false;
-   }
-
    private static boolean isGooglePlayStoreInstalled() {
       try {
          PackageManager pm = OneSignal.appContext.getPackageManager();
@@ -39,7 +28,7 @@ class GooglePlayServicesUpgradePrompt {
    }
 
    static void ShowUpdateGPSDialog() {
-      if (isGMSInstalledAndEnabled() || !isGooglePlayStoreInstalled())
+      if (new OSUtils().getDeviceType() != UserState.DEVICE_TYPE_ANDROID || !isGooglePlayStoreInstalled())
          return;
 
       boolean userSelectedSkip =
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
@@ -27,8 +27,8 @@ class GooglePlayServicesUpgradePrompt {
       return false;
    }
 
-   static void ShowUpdateGPSDialog() {
-      if (new OSUtils().getDeviceType() != UserState.DEVICE_TYPE_ANDROID || !isGooglePlayStoreInstalled())
+   static void showUpdateGPSDialog() {
+      if (!OSUtils.isAndroidDeviceType() || !isGooglePlayStoreInstalled())
          return;
 
       boolean userSelectedSkip =
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -969,10 +969,9 @@ public class OneSignal {
       if (mPushRegistrator != null)
          return mPushRegistrator;
 
-      int deviceType = osUtils.getDeviceType();
-      if (deviceType == UserState.DEVICE_TYPE_FIREOS)
+      if (OSUtils.isFireOSDeviceType())
          mPushRegistrator = new PushRegistratorADM();
-      else if (new OSUtils().getDeviceType() == UserState.DEVICE_TYPE_ANDROID) {
+      else if (OSUtils.isAndroidDeviceType()) {
          if (OSUtils.hasFCMLibrary())
             mPushRegistrator = new PushRegistratorFCM();
          else
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -972,10 +972,14 @@ public class OneSignal {
       int deviceType = osUtils.getDeviceType();
       if (deviceType == UserState.DEVICE_TYPE_FIREOS)
          mPushRegistrator = new PushRegistratorADM();
-      else if (OSUtils.hasFCMLibrary())
-         mPushRegistrator = new PushRegistratorFCM();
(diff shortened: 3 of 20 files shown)


OneSignalSDK/onesignal/build.gradle
```diff
@@ -56,6 +56,10 @@ dependencies {
     // firebase-messaging:17.6.0 is the max version since we still have code looking for FirebaseInstanceIdService
     api 'com.google.firebase:firebase-messaging:[10.2.1, 17.3.99]'
 
+    // Huawei PushKit
+    // KEEP as "compileOnly", so OneSignal isn't a direct dependency in the POM file.
+    compileOnly 'com.huawei.hms:push:4.0.2.300'
```
mikechoch, line 61: Should we include a range here or keep it specific for now?

OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
```diff
@@ -39,7 +28,7 @@ private static boolean isGooglePlayStoreInstalled() {
    }
 
    static void ShowUpdateGPSDialog() {
```
mikechoch, line 30: Convert `ShowUpdateGPSDialog` to lowercase first letter? 

`showUpdateGPSDialog`

OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
```diff
@@ -39,7 +28,7 @@ private static boolean isGooglePlayStoreInstalled() {
...
-      if (isGMSInstalledAndEnabled() || !isGooglePlayStoreInstalled())
+      if (new OSUtils().getDeviceType() != UserState.DEVICE_TYPE_ANDROID || !isGooglePlayStoreInstalled())
```
mikechoch, line 31: Maybe we can refactor something along these lines across the codebase?
```
UserState deviceType = new OSUtils().getDeviceType();
if (!deviceType.isAndroidDevice() || !isGooglePlayStoreInstalled())
```

This would make the `getDeviceType()` return `UserState` instead of `int` though
Seems a bit out of scope for this but thought I would bring it up

So maybe...?
```
int deviceType = new OSUtils().getDeviceType();
if (deviceType != UserState.DEVICE_TYPE_ANDROID || !isGooglePlayStoreInstalled())
```

OneSignalSDK/unittest/src/test/java/com/test/onesignal/DeviceTypeTestsRunner.java
```diff
@@ -0,0 +1,128 @@
...
+    public void supportsFCMAndADM_PreferADM() {
+        ShadowOSUtils.isGMSInstalledAndEnabled = true;
+        ShadowOSUtils.hasFCMLibrary = true;
+
+        ShadowOSUtils.supportsADM = true;
+
+        // Prefer ADM as if available it will always be native to the device
+        assertEquals(DEVICE_TYPE_FIREOS, getDeviceType());
+    }
+}
```
mikechoch, line 128: Just an overall comment for these `UnitTests`, maybe we should offer some steps explaining what each line truly means or create more specific test names.

This one is a good example and offers clarity
`supportsFCMAndADM_PreferADM`
but even could be `testDevice_withFCMAndADMSupport_prefersADM`

however, this one
`supportedHMS_isHuawei` 
could be
`testDevice_withHMSSupport_prefersHMS`

I'm probably being to critical though since it only took me a few seconds to figure out the pattern here. I guess this could stay the same since I recognized the consistent pattern, but wanted to bring it up to see what you thought.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/PushRegistratorHMSIntegrationTestsRunner.java
```diff
@@ -0,0 +1,128 @@
...
+    public void beforeEachTest() throws Exception {
+        TestHelpers.beforeTestInitAndCleanup();
+
+        ShadowOSUtils.supportsHMS(true);
+
+        blankActivityController = Robolectric.buildActivity(BlankActivity.class).create();
+        blankActivity = blankActivityController.get();
+    }
+
+    private static void assertHuaweiSubscribe() throws JSONException {
```
mikechoch, line 75: Should we throw these `Assert` helpers inside another `class` like `JsonAsserts` and `RestClientAsserts`
maybe `DeviceTypeAsserts`?

OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
```diff
@@ -0,0 +1,88 @@
...
+            OneSignal.Log(LOG_LEVEL.INFO, "Device registered for HMS, push token = " + pushToken);
+            callback.complete(pushToken, UserState.PUSH_STATUS_SUBSCRIBED);
+        }
+        else
+            waitForOnNewPushTokenEvent(callback);
+    }
+
+    private static void doTimeOutWait() {
+        try {
+            Thread.sleep(NEW_TOKEN_TIMEOUT_MS);
```
Jeasmine, line 74: do you need to sleep the thread? cant this be runnable with post delay?

OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
Jeasmine: Uhm maybe have on OSUtils a method

boolean isAndroidDevice(){
return getDeviceType() == UserState.DEVICE_TYPE_ANDROID;
}

if that make sense

OneSignalSDK/onesignal/build.gradle
jkasten2: Since this is `compileOnly` let's not use a version range here so builds are consistent.

OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
jkasten2: @Jeasmine I like your suggestion to clean this up. Using this definition which is `static` and includes `DeviceType` in the name for completeness.

```java
   static boolean isAndroidDeviceType() {
      return new OSUtils().getDeviceType() == UserState.DEVICE_TYPE_ANDROID;
   }
```

OneSignalSDK/onesignal/src/main/java/com/onesignal/GooglePlayServicesUpgradePrompt.java
jkasten2: 👍 Good catch, might as well clean this up while in here

OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
jkasten2: This sleep is running on it's own thread created from the `registerForPush` above. What are the benefits of using `postDelayed`? Since this is it's own thread we would also need to start a looper is why I am asking.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/DeviceTypeTestsRunner.java
jkasten2: Adding `testDevice_` seems reducend if prefixing each test with it. 

The pattern I was going for was `state_expectedResult`.

I agree that we should narrow down a naming pattern, since the scope of this problem is limited to HMS only lets defer

OneSignalSDK/unittest/src/test/java/com/test/onesignal/PushRegistratorHMSIntegrationTestsRunner.java
jkasten2: It could go into `RestClientAsserts`, however it does include `assertRestCalls(2);` which is very specific to the UnitTest in this class.

A method code be added to `RestClientAsserts` that includes these two lines, but since this probably won't be used in any other test classes I think it best to leave as is.
```java
assertHuaweiPlayerCreateAtIndex(1);
assertPlayerCreateSubscribedAtIndex(1);
```

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -972,10 +972,14 @@ private static PushRegistrator getPushRegistrator() {
       int deviceType = osUtils.getDeviceType();
       if (deviceType == UserState.DEVICE_TYPE_FIREOS)
          mPushRegistrator = new PushRegistratorADM();
-      else if (OSUtils.hasFCMLibrary())
-         mPushRegistrator = new PushRegistratorFCM();
+      else if (new OSUtils().getDeviceType() == UserState.DEVICE_TYPE_ANDROID) {
```
Jeasmine, line 975: You can use the new utils function

Jeasmine: oh I think the new util function is in the other PR 😢

OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
Jeasmine: I was worried if the thread was being shared, but is it all runs in the same thread I suppose there is no harm. The postDelay was to not leave a thread without doing nothing.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
jkasten2: Added to this one, since it is a better fit anyway

PR #1034, opened 2020-05-27
Start implementing Worker classes for processing NotificationWillShowInForegroundHandlers

mikechoch:
##
* Add `AndroidX work` dependency with range `2.0.0` to `2.99.99`
* Created abstract `NotificationWillShowInForegroundWorker` class extending `Worker`
* Two classes for processing work (both extending `NotificationWillShowInForegroundWorker`):
  * `ExtNotificationWillShowInForegroundWorker`
  * `AppNotificationWillShowInForegroundWorker`

##
* `OSNotificationWillShowInForegroundManager` class was added
* Now 2 Workers exist extending `Worker` `class`
  * `ExtNotificationWillShowInForegroundWorker`
  * `AppNotificationWillShowInForegroundWorker`
* A Singleton manager class exists as well and keeps track of the jobs that need to be run and exist or dont
  * `NotificationWillShowInForegroundManager`
* Added `work-testing` dependency to manage UnitTests involving the `Worker` and `WorkManager` classes
  * Version `2.3.4` for now

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1034)


OneSignalSDK/onesignal/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     api 'androidx.legacy:legacy-support-v4:[1.0.0, 1.99.99]'
     api 'androidx.browser:browser:[1.0.0, 1.99.99]'
     api 'androidx.appcompat:appcompat:[1.0.0, 1.99.99]'
+    api 'androidx.work:work-runtime:[2.0.0, 2.99.99]'
```
jkasten2, line 52: Let's see if we can use 1.0.0. Ok if not, just better compatibility.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWillShowInForegroundManager.java
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            // Make sure the key we need to to obtain our notifJob exists in the data input of the Worker
+            if (!getInputData().hasKeyWithValueOfType(APP_NOTIF_JOB_WORKER_KEY, String.class)) {
+                OneSignal.onesignalLog(OneSignal.LOG_LEVEL.ERROR, "getInputData() does not have the key: " + EXT_NOTIF_JOB_WORKER_KEY + " with class type: String, returning Result.failure for Worker");
+                return Result.failure();
+            }
+
+            // The job key must not be null and a job should exist for the key
+            String appNotifJobKey = getInputData().getString(APP_NOTIF_JOB_WORKER_KEY);
```
jkasten2, line 181: We can do this in a follow up PR but we should just put a whole payload here instead of a key look. This way if the app process restarts we still have all the data to do the job.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWillShowInForegroundManager.java
```diff
@@ -0,0 +1,208 @@
...
+            super(context, workerParams);
+        }
+
+        @NonNull
+        @Override
+        public Result doWork() {
+            OneSignal.onesignalLog(OneSignal.LOG_LEVEL.INFO, "doWork() called for AppNotificationWillShowInForegroundWorker");
+
+            // appContext my have been used to begin "doWork"
+            if (OneSignal.appContext == null) {
```
jkasten2, line 169: We can remove null check, we will always have a context here.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1985,18 +2074,51 @@ public void run() {
...
+    * @see OneSignal.ExtNotificationWillShowInForegroundHandler
+    * @see OSNotificationWillShowInForegroundManager.ExtNotificationWillShowInForegroundWorker
+    * @see OneSignal.AppNotificationWillShowInForegroundHandler
+    * @see OSNotificationWillShowInForegroundManager.AppNotificationWillShowInForegroundWorker
+    */
+   static void fireNotificationWillShowInForegroundHandlers(OSNotificationGenerationJob notifJob) {
+      // No ExtNotificationWillShowInForegroundHandler or AppNotificationWillShowInForegroundHandler was setup, show the notification
+      if (extNotificationWillShowInForegroundHandler == null && appNotificationWillShowInForegroundHandler == null) {
+         OneSignal.onesignalLog(OneSignal.LOG_LEVEL.INFO, "extNotificationWillShowInForegroundHandler and appNotificationWillShowInForegroundHandler are both null, moving on to showing OS notification id: " + notifJob.getApiNotificationId());
+         GenerateNotification.fromJsonPayload(notifJob);
```
jkasten2, line 2109: `GenerateNotification.fromJsonPayload(notifJob);` is called from within this `fireNotificationWillShowInForegroundHandlers`. Would be more clear if `fireNotificationWillShowInForegroundHandlers` would return a `boolean` that it did create an event or not then call `GenerateNotification.fromJsonPayload(notifJob);` outside of this method.

OneSignalSDK/onesignal/build.gradle
mikechoch: `work-runtime` started at min `2.0.0`, we discussed we will keep range `[2.0.0, 2.99.99]`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWillShowInForegroundManager.java
mikechoch: Changing to a worker that spans the processing work all the way to the notification showing instead

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWillShowInForegroundManager.java
mikechoch: Removing!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
mikechoch: Got it, sounds good!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    int getAndroidIdWithoutCreate() {
+        if (overrideSettings == null || overrideSettings.androidNotificationId == null)
+            return -1;
+
+        return overrideSettings.androidNotificationId;
+    }
+
+    Integer getAndroidId() {
```
Jeasmine, line 82: can this be a primitive?

PR #1038, opened 2020-06-01
Refactor Influence module to Kotlin

Jeasmine:
* Change influence module classes from java to kotlin
  * Remove instrumentedPackage from roboelectric tests, this will make time shadow to not work
  * Add OSTime for time method access
  * Mock OSTime on Tests

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1038)


OneSignalSDK/unittest/build.gradle
```diff
@@ -52,4 +54,10 @@ dependencies {
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.robolectric:robolectric:4.3.1'
     testImplementation 'org.awaitility:awaitility:3.1.5'
+    testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+
+    testImplementation 'org.reflections:reflections:0.9.12'
+}
+repositories {
+    mavenCentral()
```
jkasten2, line 62: This should be in the root `build.gradle`.

OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSChannelTracker.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Get the current session based on state + if outcomes features are enabled.
+     */
+    val currentSessionInfluence: OSInfluence
+        get() {
+            val sessionInfluence = OSInfluence(channelType, OSInfluenceType.DISABLED, null)
+            // Channel weren't init yet because application is starting
+            if (influenceType == null) initInfluencedTypeFromCache()
+            influenceType?.let {
```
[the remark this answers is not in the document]
Jeasmine: made an auxiliary currentInfluenceType to avoid null check

OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSChannelTracker.kt
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    override fun hashCode(): Int {
+        var result = influenceType.hashCode()
+        result = 31 * result + idTag.hashCode()
+        return result
+    }
+
+    companion object {
+        private const val TIME = "time"
```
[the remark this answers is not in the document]
Jeasmine: Thanks for this comment! I was looking to de decompiled java code for companion object and object inside a class, for both cases constants object are inlined where used, but a static class is defined and not used, so I checked that the best way is to have a single object file, in that way we don't have companion/object in the classes. Also, companion is a special case of an object. We should only use a companion for static initialization or method access. For reference https://blog.egorand.me/where-do-i-put-my-constants-in-kotlin/

OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSChannelTracker.kt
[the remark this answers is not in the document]
jkasten2: 👍 `val currentInfluenceType = influenceType ?: OSInfluenceType.DISABLED`

OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSChannelTracker.kt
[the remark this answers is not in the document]
jkasten2: 👍

OneSignalSDK/build.gradle
```diff
@@ -18,6 +19,8 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:3.6.2'
 
+        classpath 'com.google.gms:google-services:4.3.2'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
mikechoch, line 23: Super nit picky here but lets use `'` instead of `"` so its consistent with others

mikechoch: Also any reason we don't just add '1.3.72' to the end of the `classpath` instead of `ext.kotlin_version = '1.3.72'`

OneSignalSDK/onesignal/build.gradle
```diff
@@ -55,6 +57,8 @@ dependencies {
     api 'androidx.browser:browser:[1.0.0, 1.99.99]'
     api 'androidx.appcompat:appcompat:[1.0.0, 1.99.99]'
     api 'androidx.work:work-runtime:[2.0.0, 2.99.99]'
+
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
mikechoch, line 61: Super nit picky here but lets use ' instead of " so its consistent with others

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -3072,9 +3073,17 @@ static void fireEmailUpdateFailure() {
       }
    }
 
+   @NonNull
+   public static OSTime getTime() {
```
mikechoch, line 3077: Do we want this to be `public`?

OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/data/OSChannelTracker.kt
```diff
@@ -0,0 +1,175 @@
...
+                    influenceType = OSInfluenceType.UNATTRIBUTED
+                }
+            }
+
+            return sessionInfluence
+        }
+
+    /**
+     * Get all received ids that may influence actions
+     *
```
mikechoch, line 76: Remove new line here

OneSignalSDK/unittest/src/test/java/com/onesignal/MockOSTime.java
```diff
@@ -0,0 +1,43 @@
+package com.onesignal;
+
+import android.os.SystemClock;
+
+public class MockOSTime implements OSTime {
```
mikechoch, line 5: Rename to `MockOSTimeImpl` to match our real `OSTimeImpl`?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
```diff
@@ -836,7 +842,7 @@ public void shouldRestoreNotifications() {
 
       // Go forward 1 week
       // Note: Does not effect the SQL function strftime
-      advanceSystemTimeBy(604_801);
+      advanceSystemTimeBy(time, 604_801);
```
mikechoch, line 845: Instead of passing everything into the `advanceSystemTimeBy` maybe we should just call a method inside of `time`?

**Example:**
`time.advanceSystemTimeBy(604_801)`?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -2516,12 +2531,13 @@ public void ensureSyncJobServiceRescheduleOnApiTimeout() throws Exception {
    }
 
    private void useAppFor2minThenBackground() throws Exception {
+      advanceSystemAndElapsedTimeBy(time,0);
```
mikechoch, line 2534: Add a space `advanceSystemAndElapsedTimeBy(time, 0);`

OneSignalSDK/unittest/src/test/java/com/test/onesignal/TestHelpers.java
```diff
@@ -241,11 +241,11 @@ static void fastColdRestartApp() throws Exception {
       StaticResetHelper.restSetStaticFields();
    }
 
-   static void restartAppAndElapseTimeToNextSession() throws Exception {
+   static void restartAppAndElapseTimeToNextSession(MockOSTime time) throws Exception {
       stopAllOSThreads();
       flushBufferedSharedPrefs();
       StaticResetHelper.restSetStaticFields();
-      advanceSystemTimeBy(31);
+      advanceSystemAndElapsedTimeBy(time,31);
```
mikechoch, line 248: Add a space `advanceSystemAndElapsedTimeBy(time, 31);`

OneSignalSDK/build.gradle
Jeasmine: I wanted to start cleaning this, I usually use ext to define versions, but is not a must, I can change to make all use this new format or remove it, let me know what you think

OneSignalSDK/onesignal/build.gradle
Jeasmine: oh good eye!

Jeasmine: oooh actually if we want to start using $kotlin_version variables on gradle, it only use it with " let mek nos if you prefer to continue with the old format

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: good catch!

OneSignalSDK/unittest/src/test/java/com/onesignal/MockOSTime.java
Jeasmine: love this one!

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
Jeasmine: :not_bad: 👍

OneSignalSDK/onesignal/build.gradle
mikechoch: Then maybe as we start replacing with `ext.` we will add `"` to all other dependencies

PR #1042, opened 2020-06-02
Refactor Outcome module to Kotlin

Jeasmine:
* Change outcome module classes from java to kotlin

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1042)


OneSignalSDK/onesignal/src/main/java/com/onesignal/outcomes/OSOutcomeEventsV1Repository.kt
```diff
@@ -0,0 +1,69 @@
...
+                    .also { jsonObject ->
+                        outcomeEventsService.sendOutcomeEvent(jsonObject, responseHandler)
+                    }
+        } catch (e: JSONException) {
+            logger.error("Generating unattributed outcome:JSON Failed.", e)
+        }
+    }
+
+    companion object {
+        private const val DIRECT = "direct"
```
jkasten2, line 67: Might be better to use just object so there isn't a companion object under the hood.
https://stackoverflow.com/a/50520935
@Jeasmine Thoughts on this?

Jeasmine: Agree, still need to fix the companion for DB, but I need to rebase first due to some changes on that classes, I need them in order to don't do too much re-work

jkasten2: 👍 I see this is now cleaned up
https://github.com/OneSignal/OneSignal-Android-SDK/pull/1042/files#diff-ebb1a5f7f3ff02ae231e90aa99ed7818R3

PR #1043, opened 2020-06-03
HMS location fix ups

jkasten2:
* Changed gradle to compileOnly so it isn't a direct dependecy
* Renamed ShadowFusedLocationProviderClient -> ShadowHMSFusedLocationProviderClient
   - Naming more clear this is only for HMS
* Removed hms gradle plugin from onesignal SDK target
   - It only applies to app targets

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1043)

### DIFF
--- a/OneSignalSDK/onesignal/build.gradle
+++ b/OneSignalSDK/onesignal/build.gradle
@@ -1,5 +1,4 @@
 apply plugin: 'com.android.library'
-apply plugin: 'com.huawei.agconnect'
 
 android {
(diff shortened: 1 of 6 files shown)


jkasten2:
After making the SDK `compileOnly` forgot to add the them to unittests target as `implements`. ALso upgraded to version 4 as it was on the old version 3 which needed a small rename for their new `HWLocation` class.

PR #1049, opened 2020-06-09
Fix DIRECT IAM influence being cache

Jeasmine:
* IAM Direct influence shoudn't be cache because DIRECT Influence will be only while the IAM is being display
   * Instead of caching DIRECT cache INDIRECT

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSInAppMessageTracker.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSInAppMessageTracker.java
@@ -95,6 +95,7 @@ class OSInAppMessageTracker extends OSChannelTracker {
 
     @Override
     public void cacheState() {
-        dataRepository.cacheIAMInfluenceType(influenceType == null ? OSInfluenceType.UNATTRIBUTED : influenceType);
+        OSInfluenceType influenceTypeToCache = influenceType == null ? OSInfluenceType.UNATTRIBUTED : influenceType;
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/influence/OSInAppMessageTracker.java
```diff
@@ -95,6 +95,7 @@ void addSessionData(@NonNull JSONObject jsonObject, OSInfluence influence) {
 
     @Override
     public void cacheState() {
-        dataRepository.cacheIAMInfluenceType(influenceType == null ? OSInfluenceType.UNATTRIBUTED : influenceType);
+        OSInfluenceType influenceTypeToCache = influenceType == null ? OSInfluenceType.UNATTRIBUTED : influenceType;
+        dataRepository.cacheIAMInfluenceType(influenceTypeToCache == OSInfluenceType.DIRECT ? OSInfluenceType.INDIRECT : influenceTypeToCache);
     }
```
mikechoch, line 100: A comment should be added to explain that within `cacheState` for IAM tracker we always downgrade `DIRECT` since leaving the app while showing an IAM would be a `DIRECT` state

mikechoch, line 100: A comment should be added to explain that within `cacheState` for IAM tracker we always downgrade `DIRECT` since leaving the app while showing an IAM would be a `DIRECT` state

PR #1050, opened 2020-06-09
Adding 2 layers of control to prevent locks in SQL DB and minor clean up

mikechoch:
* Removed the commented `enableWriteAheadLogging`
  * No need for this as this allows concurrent read and writes to occur and we would rather pause a read while a write occurs
  * Pausing will ensure that we are querying the most updated data
* 2 layers of security for SQL DB regarding `OSInAppMessageRepository.java`
  * First one will be transactions using `beginTransaction`, `setTransactionSuccessful`, and `endTransaction` methods wherever we `insert`, `delete`, or `update`
  * Second one is keeping TABLE manipulation or reading within the same file and setting these methods as synchronized so that any other TABLE touching that occurs won't happen till any current methods are complete
  * The combination of the two steps should help prevent any conflicts between SQL DB and solve the locking issues

* WIP - Making sure any DB reading and writing is from the same file so `synchronized` along with transactions provide this 2 layer security around our single SQL DB instance

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1050)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
@@ -32,7 +32,6 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
 
     public static final String IN_APP_MESSAGES_JSON_KEY = "in_app_messages";
     private static final String OS_SAVE_IN_APP_MESSAGE = "OS_SAVE_IN_APP_MESSAGE";
-    private static final String OS_DELETE_IN_APP_MESSAGE = "OS_DELETE_IN_APP_MESSAGE";
 
     OSTriggerController triggerController;
     private OSSystemConditionController systemConditionController;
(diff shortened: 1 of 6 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalCacheCleaner.java
```diff
@@ -1,24 +1,19 @@
...
-import java.util.Set;
+import com.onesignal.OneSignalDbContract.NotificationTable;
+import com.onesignal.influence.model.OSInfluenceChannel;
+import com.onesignal.outcomes.OSOutcomeTableProvider;
 
 class OneSignalCacheCleaner {
 
     private final static long NOTIFICATION_CACHE_DATA_LIFETIME = 604_800L; // 7 days in seconds
-    private final static long IAM_CACHE_DATA_LIFETIME = 15_552_000L; // 6 months in seconds
+    final static long IAM_CACHE_DATA_LIFETIME = 15_552_000L; // 6 months in seconds
```
Jeasmine, line 16: maybe move it to the IAM Repository file?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalCacheCleaner.java
```diff
@@ -61,73 +66,13 @@ public void run() {
...
-                // 2. Delete old IAMs from SQL
-                writableDb.delete(
-                        OneSignalDbContract.InAppMessageTable.TABLE_NAME,
-                        whereStr,
-                        whereArgs);
-
-                // 3. Use queried data to clean SharedPreferences
-                cleanInAppMessageIds(oldMessageIds);
-                cleanInAppMessageClickedClickIds(oldClickedClickIds);
+                new OSInAppMessageRepository(dbHelper).cleanCachedInAppMessages();
```
Jeasmine, line 74: maybe access it throw OneSignal.java? or add it to the method params? because there should be already a repository instance alive if we have more that one instance then te synchronized won't work

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -122,8 +121,15 @@ protected OSInAppMessageController(OneSignalDbHelper dbHelper) {
         initRedisplayData(dbHelper);
     }
 
+    OSInAppMessageRepository getInAppMessageRepository(OneSignalDbHelper dbHelper) {
```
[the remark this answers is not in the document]
jkasten2: Wasn't the issue, test is flaky on it's own. Nothing in this PR would have affected the failing tests so closing.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalCacheCleaner.java
mikechoch: Moved to IAM Repository file

PR #1061, opened 2020-06-11
Only attempt to get Google ad_id if using FCM

jkasten2:
* This cleans up the "Error getting Google Ad id"  on HMS only devices

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1061)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -452,7 +452,15 @@ public class OneSignal {
    @Nullable private static OSOutcomeEventsController outcomeEventsController;
    @Nullable private static OSOutcomeEventsFactory outcomeEventsFactory;
 
-   private static AdvertisingIdentifierProvider mainAdIdProvider = new AdvertisingIdProviderGPS();
+   @Nullable private static AdvertisingIdentifierProvider adIdProvider;
+   private static synchronized @Nullable AdvertisingIdentifierProvider getAdIdProvider() {
+      if (adIdProvider == null) {
+         if (OSUtils.isAndroidDeviceType())
+            adIdProvider = new AdvertisingIdProviderGPS();
+      }
+
+      return adIdProvider;
+   }
 
    @SuppressWarnings("WeakerAccess")
    public static String sdkType = "native";
@@ -1342,9 +1350,11 @@ public class OneSignal {
 
       deviceInfo.put("app_id", getSavedAppId());
 
-      String adId = mainAdIdProvider.getIdentifier(appContext);
-      if (adId != null)
-         deviceInfo.put("ad_id", adId);
+      if (getAdIdProvider() != null) {
+         String adId = getAdIdProvider().getIdentifier(appContext);
+         if (adId != null)
+            deviceInfo.put("ad_id", adId);
+      }
       deviceInfo.put("device_os", Build.VERSION.RELEASE);
       deviceInfo.put("timezone", getTimeZoneOffset());
       deviceInfo.put("language", OSUtils.getCorrectedLanguage());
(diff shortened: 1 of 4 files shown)


mikechoch:
I see we check the `ad_id` is `null` for the Huawei tests, should we add a UnitTest for Amazon devices checking against the `ad_id` as well? 

Maybe a separate PR, but otherwise LGTM!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -452,7 +452,15 @@ public void onSessionEnding(@NonNull List<OSInfluence> lastInfluences) {
    @Nullable private static OSOutcomeEventsController outcomeEventsController;
    @Nullable private static OSOutcomeEventsFactory outcomeEventsFactory;
 
-   private static AdvertisingIdentifierProvider mainAdIdProvider = new AdvertisingIdProviderGPS();
+   @Nullable private static AdvertisingIdentifierProvider adIdProvider;
+   private static synchronized @Nullable AdvertisingIdentifierProvider getAdIdProvider() {
+      if (adIdProvider == null) {
+         if (OSUtils.isAndroidDeviceType())
```
Jeasmine, line 458: maybe put on the same if?

jkasten2: I set it up nested like this as we may want to add an `else if (OSUtils.isHuaweiDeviceType()` ect if we can support other ad providers in the future.

Jeasmine: sounds good!

jkasten2:
> I see we check the `ad_id` is `null` for the Huawei tests, should we add a UnitTest for Amazon devices checking against the `ad_id` as well?
> 
> Maybe a separate PR, but otherwise LGTM!

I think it would be good to have these tesst in this PR with the code changes. I'll work on that

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -322,6 +323,59 @@ public void testDeviceTypeIsAmazon_forPlayerCreate() throws Exception {
...
+      threadAndTaskWait();
+
+      // 3. Make sure device_type is Huawei (13) in player create
+      assertHuaweiPlayerCreateAtIndex(1);
+
+      // 4. Assert Player Create does NOT have an ad_id
+      ShadowOneSignalRestClient.Request request = ShadowOneSignalRestClient.requests.get(1);
+      JsonAsserts.doesNotContainKeys(request.payload, new ArrayList<String>() {{ add("ad_id"); }});
+
+      // 4. Assert we did NOT try to get a Google Ad id
```
mikechoch, line 375: `4.` should be `5.`

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -322,6 +323,59 @@ public void testDeviceTypeIsAmazon_forPlayerCreate() throws Exception {
...
+      threadAndTaskWait();
+
+      // 3. Make sure device_type is Amazon (2) in player create
+      assertAmazonPlayerCreateAtIndex(1);
+
+      // 4. Assert Player Create does NOT have an ad_id
+      ShadowOneSignalRestClient.Request request = ShadowOneSignalRestClient.requests.get(1);
+      JsonAsserts.doesNotContainKeys(request.payload, new ArrayList<String>() {{ add("ad_id"); }});
+
+      // 4. Assert we did NOT try to get a Google Ad id
```
mikechoch, line 342: `4.` should be `5.`

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
jkasten2: Done

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
jkasten2: Done

PR #1064, opened 2020-06-11
Hms consumer proguard rules

jkasten2:
Added `dontwarn` for `com.huawei.**`

* This is to avoid any progaurd build errors / warns not finding these
classes when HMS libraries are not used in an app with proguard / R8.

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1064)


jkasten2:
> LGTM, but what warnings does this suppress?
> Maybe we should just take note for future reference?

Basically something like can't find the class referenced. However this PR should prevent it in all cases as it part of the .aar and is automatically applied by Gradle.

PR #1066, opened 2020-06-12
Add OSDevice

Jeasmine:
* Unify all user device data on a same class
   * Getters for notification permission, user subscription, is user susbscribed,
     user id, push token, email id, email addess

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1066)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
@@ -1,0 +1,87 @@
+/**
+ * Modified MIT License
+ *
+ * Copyright 2020 OneSignal
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * 1. The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * 2. All copies of substantial portions of the Software may only be used in connection
+ * with services provided by OneSignal.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class OSDevice {
+
+    /**
+     * Get the app's notification permission
+     *
+     * @return false if the user disabled notifications for the app, otherwise true
+     * */
+    public boolean isNotificationEnabled() {
+        return OneSignal.getPermissionSubscriptionState().permissionStatus.getEnabled();
```
mikechoch, line 38: Lets rename `getPermissionSubscriptionState` so its a little more clear that
```
subscriptionStatus
permissionStatus
emailSubscriptionStatus
```
all live within this getter.

Maybe 
`getDeviceObservers`?
`getSubscriptionObservers`?
`getObservers`?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -581,6 +581,14 @@ private static OSEmailSubscriptionState getLastEmailSubscriptionState(Context co
    }
    // End EmailSubscriptionState
 
+   private static OSDevice currentDevice;
```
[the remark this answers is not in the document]
Jeasmine: userDevice works? because m prefix is for instance variables, this one is static, so in that case should have S prefix, which I don't like 😂

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
Jeasmine: you mean renaming permissionStatus getters right? this wont have a problem with the wrappers?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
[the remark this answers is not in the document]
mikechoch: `userDevice` sounds fine

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
mikechoch: We should still have a major release ticket for refactoring this naming `getPermissionSubscriptionState`

PR #1075, opened 2020-06-23
Fix Init context

Jeasmine:
* Since foreground is being calculate from an activity context, init method should reference to initial context
  * Fix osIAMPreview_showsPreview and shouldCallOnSessionEvenIfSyncJobStarted_Huawei test

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1075)


Jeasmine:
> LGTM, but not sure how this differs exactly from before. Seems like we are just adding a parameter to `init` method now instead of access `static` `appContext`

The main difference is in handleActivityLifecycleHandler(context);
Because on that method we check if the context is an Activity, If that is true then inForeground is set to true, and with that some logic flow are being able to continue

PR #1077, opened 2020-06-25
Moved event logic into OneSignalHmsEventBridge

jkasten2:
* Moved logic from HmsMessageServiceOneSignal into to a new
OneSignalHmsEventBridge class.
* OneSignalHmsEventBridge need to be called if by the app if:
   1. Another push provider library / SDK is in the app.
   2. The app has it's own HmsMessageService class.
* This is required as HMS only creates one HmsMessageService instance.
* See comemnts in this commit for even more details.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1077)


jkasten2:
One test run passed, the other one failed on an unrelated `testMixedTriggersScheduleTimer` which indicates it is flaky.

jkasten2:
Merging this PR into master. Making another PR for dev app integration. Bridge is working in my local testing right now.

PR #1083, opened 2020-07-01
Fix/only call foreground handler while app in foreground

mikechoch:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1083)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -2131,8 +2131,30 @@ public class OneSignal {
    }
 
    /**
-    * A few checks before before firing the normal expected scenario where a ExtNotificationWillShowInForegroundHandler
-    *  and AppNotificationWillShowInForegroundHandler are set
+    * Checks if the app is in the background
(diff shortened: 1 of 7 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2142,7 +2142,7 @@ static void handleNotificationReceived(OSNotificationGenerationJob notifJob, boo
     */
    static void fireNotificationWillShowInForegroundHandlers(OSNotificationGenerationJob notifJob) {
       // No ExtNotificationWillShowInForegroundHandler or AppNotificationWillShowInForegroundHandler was setup, show the notification
-      if (extNotificationWillShowInForegroundHandler == null && appNotificationWillShowInForegroundHandler == null) {
+      if (extNotificationWillShowInForegroundHandler == null && appNotificationWillShowInForegroundHandler == null || !isInForeground()) {
```
jkasten2, line 2145: Would like to see the `isInForeground()` check as its own if statement so this is easier to read.

mikechoch: Yeah I like that, agreed

PR #1090, opened 2020-07-13
Fix background location update

Jeasmine:
* Avoid reschedule an already scheduled or running service
   * If need to reschedule, reschedule under jobFinished

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1090)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalSyncServiceUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalSyncServiceUtils.java
@@ -63,6 +63,7 @@ class OneSignalSyncServiceUtils {
    private static final long SYNC_AFTER_BG_DELAY_MS = OneSignal.MIN_ON_SESSION_TIME_MILLIS;
 
    private static Long nextScheduledSyncTimeMs = 0L;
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalSyncServiceUtils.java
```diff
@@ -142,20 +141,21 @@ private static boolean hasBootPermission(Context context) {
...
 
    @RequiresApi(21)
-   private static boolean scheduleSyncServiceAsJob(Context context, long delayMs) {
+   private static void scheduleSyncServiceAsJob(Context context, long delayMs) {
       OneSignal.Log(OneSignal.LOG_LEVEL.VERBOSE, "scheduleSyncServiceAsJob:atTime: " + delayMs);
 
       if (isJobIdRunning(context)) {
          OneSignal.Log(OneSignal.LOG_LEVEL.VERBOSE, "scheduleSyncServiceAsJob Scheduler already running!");
-         return false;
+         needsJobReschedule = true;
```
jkasten2, line 157: Could use a comment for this if block of why we can't schedule another job now, and that we will once the current job is finished.

PR #1091, opened 2020-07-13
Single Worker for notification processing and foreground handlers

mikechoch:
`Worker` class resides within the `OSNotificationWorkManager` class
  * `NotificationWorker` extends `Worker` class
  * Constructs a `OSNotificationReceived` instance inside `doWork` override and fires the `OneSignal.NotificationProcessingHandler`
___
Converted `NotificationExtenderService` to `OSNotificationExtender`
* Removed extended `JobIntentService`
* Removed `onNotificationProcessing` and created a public interface `OneSignal.NotificationProcessingHandler`
* Belongs to the `OSNotificationReceived` class now
___
`OSNotificationReceived` now has 3 methods for handling the `OneSignal.NotificationProcessingHandler`
* Extending `OSTimeoutHandler` now and this calls complete method if the `postDelayed` `runnable` is ran
* `setModifiedContent`, `display`, and `complete`
* Responsible for the `OSNotificationExtender` class so that `setModifiedContent`, `display`, and `complete` may be called
___
Replaced by `NotificationProcessingHandler`, which will be an implemented `interface` in a developers custom app `NotificationExtensionService`
* Follows same similar implementation to the `ExtNotificationWillShowInForegroundHandler` as well as the `NotificationOpenedHandler`
___
Added a `try-catch` around the `NotificationProcessingHandler.notificationProcessing` method call
* If something fails we can bail and still show the notification the correct way
* Might need to do this in the `NotificationWillShowInForegroundHandler` as well
___
Added `OSNotificationExtender.setupNotificationExtensionServiceClass()` to `OneSignalPackagePrivateHelper`
* For UnitTests so we don't need OneSignal init (`setAppId` and `setAppContext`) to setup a dev's extension service
  * When calling `showNotification` we will want to call `join` after so that work on thread is completely finished
___
**Fixed UnitTest(s):**
* `shouldNotFailedNotificationExtenderServiceWhenAlertIsNull` is now `testNotificationProcessing_whenAlertIsNull`
* `notificationExtenderServiceOverridePropertiesWithSummaryApi17` is now `testNotificationExtensionServiceOverridePropertiesWithSummaryApi17`
* `notificationExtenderServiceOverridePropertiesWithSummary` is now `testNotificationExtensionServiceOverridePropertiesWithSummary`
* `notificationExtenderServiceOverrideShouldOverrideAndroidNotificationId` is now `testNotificationProcessing_twoNotificationsWithSameOverrideAndroidNotificationId`
* `shouldFireNotificationExtenderService` is now `testNotificationExtensionService_notificationProcessingProperties`
___
WIP - Need to find a solution for the high priority notifications
* Previously processed through a `JobIntentService` relying on `completeWakefulIntent`
* Commented out `shouldStartFCMServiceOnAndroidOWhenPriorityIsHighAndContainsRemoteResource` temporarily
  * No solution for high priority (> 9) notifications as of now with the `WorkManager`

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1091)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
```diff
@@ -0,0 +1,91 @@
...
+        // TODO: Do we need isHighPriority only more?
+
+        Data inputData = new Data.Builder()
+                .putInt(ANDROID_NOTIF_ID_WORKER_DATA_PARAM, androidNotificationId)
+                .putString(JSON_PAYLOAD_WORKER_DATA_PARAM, jsonPayload)
+                .putLong(TIMESTAMP_WORKER_DATA_PARAM, timestamp)
+                .putBoolean(IS_RESTORING_WORKER_DATA_PARAM, isRestoring)
+                .build();
+
+        OneTimeWorkRequest workRequest = new OneTimeWorkRequest.Builder(NotificationWorker.class)
```
Jeasmine, line 32: Can it happend that for the same notification we call more than once this method?

mikechoch: Not sure but its a good point so a good idea would be to use the OS id as a tag for the job and make sure a job does not exist with that id name already. Otherwise im pretty sure we check for notifications being duplicate before this line of code

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
```diff
@@ -114,12 +112,20 @@ static void fromJsonPayload(final OSNotificationGenerationJob notifJob) {
...
          public void run() {
             showNotification(notifJob);
          }
-      }, OS_SHOW_NOTIFICATION_THREAD).start();
+      }, OS_SHOW_NOTIFICATION_THREAD);
+      thread.start();
+
+      // Stop all code until thread is done doing work to show notification
+      try {
+         thread.join();
```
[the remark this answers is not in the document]
mikechoch: Would easier to explain over a call if you have time.

Basically the join method will await the current thread until the new thread is done with its work. It does not immediately join actually.

These [docs](https://docs.oracle.com/javase/tutorial/essential/concurrency/join.html) explain it well as well as the diagram below
Thread Join in Java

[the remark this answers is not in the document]
jkasten2: But where is the 2nd thread we are kicking off and where are we joining it?
Still looking through the PR so that may uncover this question.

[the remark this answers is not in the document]
mikechoch: Ahh I understand, so there are two places we call `GenerateNotification.fromJsonPayload` and both routes come from this method `NotificationBundleProcessor.processJobForDisplay`

One is at this line
https://github.com/OneSignal/OneSignal-Android-SDK/blob/76b94f52ea4aabba6b580710dc35a692042780ea/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java#L127

Other is at this line
https://github.com/OneSignal/OneSignal-Android-SDK/blob/76b94f52ea4aabba6b580710dc35a692042780ea/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java#L301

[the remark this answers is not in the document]
mikechoch: Also that diagram is only showing you how the `Thread.join` method works, we only kick off Thread 1 technically and its only for the `showNotification` method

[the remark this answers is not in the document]
jkasten2: Update based on our discussion.
Adding the thread here prevents throwing an exception of downloading on the main thread, however this can create an ANR if downloading takes 5+ sec

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
```diff
@@ -40,258 +40,266 @@
...
+         */
+        public synchronized void complete(boolean bubble) {
+            destroyTimeout();
+
+            if (isComplete)
+                return;
+
+            isComplete = true;
+
+            // Move on to showing notification if no AppNotificationWillShowInForegroundHandler exi
```
jkasten2, line 255: `exi` should be `exists`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
```diff
@@ -40,258 +40,266 @@
...
-         isComplete = true;
+            if (isComplete)
+                return;
 
-         GenerateNotification.fromJsonPayload(getNotifJob());
-      }
-   }
+            isComplete = true;
+
+            //
```
jkasten2, line 300: Remove blank comment

mikechoch: I actually should add a comment for this haha

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
mikechoch: Good catch!

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/notification/AppNotificationExtensionService.java
```diff
@@ -1,43 +1,50 @@
...
-            return builder.setColor(getResources().getColor(R.color.colorPrimary));
+            return builder.setColor(context.getResources().getColor(R.color.colorPrimary));
          }
       };
 
-      displayNotification(overrideSettings);
+      notification.setModifiedContent(overrideSettings);
+
+      OSNotificationDisplayedResult notificationDisplayedResult = notification.display();
+      System.out.println("Android notification id: " + notificationDisplayedResult.androidNotificationId);
```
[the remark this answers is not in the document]
mikechoch: This is just demo app but yes we can replace `sysout` with `Log.i`

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
mikechoch: Fixed!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
mikechoch: Fixed!

OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
```diff
@@ -59,534 +58,530 @@
...
-            if (button.has(PUSH_MINIFIED_BUTTON_ICON)) {
-               button.put("icon", button.getString(PUSH_MINIFIED_BUTTON_ICON));
-               button.remove(PUSH_MINIFIED_BUTTON_ICON);
+        } catch (Throwable t) {
+            OneSignal.Log(OneSignal.LOG_LEVEL.ERROR, "Error assigning OSNotificationPayload values!", t);
+        }
+
+        return notification;
+    }
+
```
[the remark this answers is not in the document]
mikechoch: Fixed!

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/notification/AppNotificationExtensionService.java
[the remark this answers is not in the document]
mikechoch: Fixed!

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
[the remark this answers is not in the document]
mikechoch: Another update is we are not calling `showNotification` from Main Thread and therefore the `new Thread` is not necessary. Good talk though and after meeting with Josh to discuss this further we decided we should add a `RuntimeException` check right before the `showNotification` method [PR for Main Thread RuntimeException check](https://github.com/OneSignal/OneSignal-Android-SDK/pull/1110)

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
mikechoch: Here is my solution to what you are brining up, but maybe we want to change this or modify... Just an idea for now though...
```Java
WorkManager.getInstance(context)
    .enqueueUniqueWork(osNotificationId, ExistingWorkPolicy.KEEP, workRequest);
```

What does `ExistingWorkPolicy.KEEP` do? 

Well... If there is existing pending (uncompleted) work with the same unique name (in this case, OneSignal notification id), do nothing. Otherwise, insert the newly-specified work.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2117,6 +2129,10 @@ private static OSNotificationOpenResult generateOsNotificationOpenResult(JSONArr
    }
 
    private static void fireNotificationOpenedHandler(final OSNotificationOpenResult openedResult) {
+      // TODO: Is there a reason we need the opened handler to be fired from main thread?
```
jkasten2, line 2132: We do this so app devs can run their UI update code without having to worry about getting themselves on the main thread.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2117,6 +2129,10 @@ private static OSNotificationOpenResult generateOsNotificationOpenResult(JSONArr
...
+
+      // TODO: Once the NotificationOpenedHandler gets a Worker, we should make sure we add a catch
```
jkasten2, line 2134: All the default behavior is done on a different thread before this call.

The reason we did a catch was so a notification would still be displayed, even though the app is going to crash from an uncaught exception.

PR #1105, opened 2020-07-30
Fix IAM showing bug with Android 6 API 23

mikechoch:
* `CardView` class can not be animated with a elevation (drop shadow) set beyond 0
  * Solution is setting the elevation of the `CardView` to 0 if detected device using API 23
  * At end of the animation we then add back the elevation necessary to show the drop shadow
  * For all other devices, elevation is set from start

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1105)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java
@@ -7,6 +7,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.graphics.Color;
 import android.graphics.drawable.ColorDrawable;
+import android.os.Build;
 import android.os.Handler;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java
```diff
@@ -475,62 +485,83 @@ private void dereferenceViews() {
     }
 
     private void animateInAppMessage(WebViewManager.Position displayLocation, View messageView, View backgroundView) {
+        final CardView messageViewCardView = messageView.findViewWithTag(IN_APP_MESSAGE_CARD_VIEW_TAG);
+        Animation.AnimationListener cardViewAnimCallback = createAnimationListenerForAndroidApi23Elevation(messageViewCardView);
```
Jeasmine, line 489: shoudn't we set this only for Android 6? Since its the only case that we will listen for changes

mikechoch: Ahh true good point, ill add a check for this

mikechoch: Fixed!

PR #1099, opened 2020-07-24
Fix ANR due to on_focus call being call from MainThread

Jeasmine:
* When attempSessionUpgrade is call, onFocus call can end being call from mainThread
* Add MainThread check on makeRequest
* Throw RuntimeException when running network calls from MainThread

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1099)

PR #1110, opened 2020-08-10
Added runtime check before showing a notification

mikechoch:
* Notification should never be processed for display from the Main Thread
  * This can cause an `ANR` if work is taking too long before showing the notification
* Created a `OSThrowable` class with a `OSMainThreadException`
* Fixed `UnitTests` thinking they were on the Main Thread
  * Added a shadow in `ShadowOSUtils` for `isRunningOnMainThread` method

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1110)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSThrowable.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSThrowable.java
@@ -27,42 +27,11 @@
 
 package com.onesignal;
 
-import android.os.Handler;
-import android.os.Looper;
+class OSThrowable {
 
-import androidx.annotation.NonNull;
-
-class OSTimeoutHandler {
-
-    private long timeout;
-
-    private Handler timeoutHandler;
-    private Runnable timeoutRunnable;
-
-    void setTimeout(long timeout) {
-        this.timeout = timeout;
+    static class OSMainThreadException extends RuntimeException {
+        public OSMainThreadException(String message) {
+            super(message);
+        }
     }
-
-    synchronized void startTimeout(@NonNull final Runnable runnable) {
-        // If the handler or runnable isn't null we do not want to start another
-        if (this.timeoutHandler != null && this.timeoutRunnable != null)
-            return;
-
-        this.timeoutRunnable = runnable;
-
-        if (Looper.myLooper() == null)
-            Looper.prepare();
-
-        timeoutHandler = new Handler();
-        timeoutHandler.postDelayed(timeoutRunnable, timeout);
-    }
-
-    synchronized void destroyTimeout() {
-        if (timeoutHandler != null)
-            timeoutHandler.removeCallbacks(timeoutRunnable);
-
-        timeoutHandler = null;
-        timeoutRunnable = null;
-    }
-
 }
--- a/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
+++ b/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
@@ -7,14 +7,8 @@ import org.robolectric.annotation.Implements;
 @Implements(OSUtils.class)
 public class ShadowOSUtils {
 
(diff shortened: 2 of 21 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSThrowable.java
```diff
@@ -0,0 +1,37 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.onesignal;
+
+class OSThrowable {
```
Jeasmine, line 30: any reason for having this "container" class?

OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
```diff
@@ -7,9 +7,15 @@
 @Implements(OSUtils.class)
 public class ShadowOSUtils {
 
+   public static boolean isRunningOnMainThread;
+
    public static String carrierName;
    public static int subscribableStatus;
 
+   public static boolean isRunningOnMainThread() {
```
Jeasmine, line 15: do we need this to be shadowed? shoudn't the test fail if it is running on the main thread? in that way we can fix it

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSThrowable.java
mikechoch: I assume we may want custom OS `Exceptions` in the future, so now we have a class to keep things separate an organized and easy to access. Do you have a better idea though? Def willing to change this

OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
mikechoch: UnitTests always think we are on Main Thread and will fail due to the RuntimeException, so now we have a shadow so we can easily test when we want to assume Main Thread or not

Jeasmine: yeah, but this shadow is being used in many places, so we are hiding tests that we actually want to check if it is not running on the main thread, josh pointed this on the ANR fix https://github.com/OneSignal/OneSignal-Android-SDK/pull/1099#discussion_r460240990

so if there are tests that are failing due to this runtime exception then we should make those tests don't run on the main thread

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSThrowable.java
Jeasmine: Got it, yeah seems right, we can continue with this structure

OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOSUtils.java
```diff
@@ -72,14 +78,14 @@ public static void resetStatics() {
       carrierName = "test1";
       subscribableStatus = 1;
 
+      isRunningOnMainThread = false;
```
jkasten2, line 81: We would like to have new `isRunningOnMainThread` check fail in tests when it should, so we know if something broke when travisCI runs.

Also this line is dangerous as it will hide other future checks we add to other things.

jkasten2: Which tests fail if we don't have this line?

Issue #1113, opened 2020-08-13
Crash - After IAM click button with "Send to test Device"


restart916:
**Description:**

Crash app after IAM click button
It happens after add setInAppMessageClickHandler and "Send to test Device" from admin page

**Environment**

compileSdkVersion 28
targetSdkVersion 28
implementation 'com.onesignal:OneSignal:3.15.2'

**Steps to Reproduce Issue:**

1. [ADMIN] Create two inapp on admin page. 
  * first 
    * has button with click action -> add some action name
  * second
    * set `3. Triggers` -> `When certain conditions are satisfied during the session`
      * `IN-APP TRIGGER` -> any key, any value
1. [Android] add `setInAppMessageClickHandler` with onesignal.init
1. [Android] add any `OneSignal.addTriggers(addTriggers);` in `inAppMessageClicked`
1. [ADMIN] Go first inapp edit page. Click "Send to test Device" on onesignal admin page
1. [Android] Click button on android virtual device
1. [Android] Crash app 

**Anything else:**

```
2020-08-13 15:49:12.603 5962-5962/com.mindclass.android E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.mindclass.android, PID: 5962
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
        at com.onesignal.OSInAppMessage.equals(OSInAppMessage.java:238)
        at java.util.ArrayList.indexOf(ArrayList.java:331)
        at java.util.ArrayList.contains(ArrayList.java:314)
        at com.onesignal.OSInAppMessageController.dismissCurrentMessage(OSInAppMessageController.java:582)
        at com.onesignal.OSInAppMessageController.messageWasDismissed(OSInAppMessageController.java:565)
        at com.onesignal.OSInAppMessageController.messageWasDismissed(OSInAppMessageController.java:541)
        at com.onesignal.WebViewManager$6.onMessageWasDismissed(WebViewManager.java:352)
        at com.onesignal.InAppMessageView.cleanupViewsAfterDismiss(InAppMessageView.java:457)
        at com.onesignal.InAppMessageView.access$1900(InAppMessageView.java:45)
        at com.onesignal.InAppMessageView$8.onAnimationEnd(InAppMessageView.java:577)
        at android.animation.Animator$AnimatorListener.onAnimationEnd(Animator.java:554)
[...]

PR #1115, opened 2020-08-14
Fix/validating removed message

restart916:
Fix Issue #1113 

After clicking the test in-app button, there is a message but there is no messageId.

This is a temporary measure
I don't know the details of the inside so you can make a better solution.

Thanks.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1115)


nan-li:
Thank you @restart916 for this PR. 
Sorry it's been a while and this has since been fixed and released.

PR #1126, opened 2020-08-26
Moved myLooper().quit() to destroyTimeout

jkasten2:
* We had `myLooper().quit()` looper at the bottom of  `doWork` however
`Looper.loop()` was blocking the thread so it would never get there.
* This fixes an issue where only the first few notifications would display
since we were hitting the max number of running workers on the WorkManager
    - This was 3 if you are testing on an emu.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1126)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
@@ -66,11 +66,6 @@ class OSNotificationWorkManager {
                 e.printStackTrace();
                 return Result.failure();
             }
-
-            // TODO: This line stops the Looper.loop call once the notification processing is all done
-            //  The reason for all of this is because we are firing handlers while running a background thread using the Worker
-            if (!OSUtils.isRunningOnMainThread())
-                Looper.myLooper().quit();
             return Result.success();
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
```diff
@@ -84,23 +79,23 @@ private void processNotificationData(Context context, int androidNotificationId,
                     timestamp
             );
 
-            if (OneSignal.notificationProcessingHandler != null)
+            if (OneSignal.notificationProcessingHandler != null) {
                 try {
                     OneSignal.notificationProcessingHandler.notificationProcessing(context, notificationReceived);
                 } catch (Throwable t) {
                     if (!notificationReceived.displayed()) {
```
Jeasmine, line 86: I think this check was ok, this display check inside the if (OneSignal.notificationProcessingHandler != null) and catch, because if the user was able to call display, then the flow continues, we should call quit on the else case

jkasten2:
Discarding this PR for now, `Looper.loop()` isn't the correct implementation here. Keep branch for a bit incase anything is useful here.

PR #1127, opened 2020-08-28
Make OSTimeoutHandler be a HandlerThread

Jeasmine:
* Workers dependens on the device cores to have a pool of threads available.
  * Loopers handling are hard to debug, and maintain
  * Make a Handler Thread available for post delays runnables
  * Make OSTimeoutHandler available betweeen jobs

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1127)

### DIFF
--- a/OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
+++ b/OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
@@ -39,7 +39,6 @@ import android.database.sqlite.SQLiteDatabase;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
-import android.os.Handler;
 import android.os.SystemClock;
 import android.util.Log;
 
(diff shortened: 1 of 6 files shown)


OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
```diff
@@ -2121,6 +2125,7 @@ public void notificationWillShowInForeground(OSNotificationGenerationJob.ExtNoti
    }
 
    @Test
+   @Config(shadows = { ShadowTimeoutHandler.class })
    public void testNotificationWillShowInForegroundHandler_workTimeLongerThanTimeout() throws Exception {
```
jkasten2, line 2129: This test is now taking ~35 seconds to complete when it wasn't before. 

I investigated this a bit and it is due to the `Thread.sleep` in `notificationWillShowInForeground` in the 
`com.test.onesignal.GenerateNotificationRunner$NotificationExtensionService_workTimeLongerThanTimeout` class never being interrupted.

The other app specific one `notificationWillShowInForeground` is being interrupted correctly however.

Jeasmine: there is no need for the 35 seconds, due that we are reducing the timeout time to 1, to avoid waiting that long. And made the thread sleep of 3 seconds

PR #1130, opened 2020-09-01
Clean build.gradle file

Jeasmine:
* Have centralized version definition

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1130)

### DIFF
--- a/OneSignalSDK/build.gradle
+++ b/OneSignalSDK/build.gradle
@@ -1,36 +1,45 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.3.72'
+
+    ext {
+        buildVersions = [
+                compileSdkVersion: 28,
+                targetSdkVersion: 28
+        ]
+        androidGradlePluginVersion = '3.6.2'
+        googleServicesGradlePluginVersion = '4.3.2'
+        huaweiAgconnectVersion = '1.2.1.301'
+        huaweiHMSPushVersion = '4.0.2.300'
+        huaweiHMSLocationVersion = '4.0.0.300'
+        kotlinVersion = '1.3.72'
+        onesignalGradlePluginVersion = '[0.12.6, 0.99.99]'
+    }
+
     repositories {
         google()
         jcenter()
         maven { url 'https://maven.google.com' }
 
         // OneSignal-Gradle-Plugin - Local testing
-//        maven { url uri('../../repo') }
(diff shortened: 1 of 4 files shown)


OneSignalSDK/build.gradle
```diff
@@ -1,36 +1,59 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.kotlin_version = '1.3.72'
+
+    ext {
```
jkasten2, line 5: In this root `build.gradle` I believe we should only define `ext` versions we want to share between more than 1 `build.gradle` file. Example `jUnitVersion` doesn't need to be her since it will only ever be used in the `unittest` project.

Jeasmine: Ok I divided on different ext, but kept the definition of the variables just for coding convention

Jeasmine: Let me know what you think

OneSignalSDK/build.gradle
```diff
@@ -1,36 +1,45 @@
...
+        buildVersions = [
+                compileSdkVersion: 28,
+                targetSdkVersion: 28
+        ]
+        googleServicesVersion = '4.3.2'
```
jkasten2, line 10: I think we should be more specific on this one.
Can we call it `googleServiceGradlePluginVersion`?

OneSignalSDK/build.gradle
```diff
@@ -1,36 +1,45 @@
...
+        gradleVersion = '3.6.2'
```
jkasten2, line 11: This is being used for `com.android.tools.build:gradle` so we should call it `androidGradlePluginVersion`.

OneSignalSDK/build.gradle
```diff
@@ -1,36 +1,45 @@
...
+        huaweiHMSPushVersion = '4.0.2.300'
+        huaweiHMSLocationVersion = '4.0.0.300'
+        huaweiVersion = '1.2.1.301'
```
jkasten2, line 14: I think we should be more specific on this one. 
Can we call it `huaweiAgconnectVersion`?

jkasten2 (changes requested):

OneSignalSDK/build.gradle
jkasten2: I like it, looks good to me!

jkasten2 (approved):

PR #1131, opened 2020-09-02
Removed no longer needed looper quit

jkasten2:
* This cleans up logcat error entry about this as well.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1131)


Jeasmine (approved):
good catch!

PR #1134, opened 2020-09-03
Make OSDevice a snapshot for state getters

Jeasmine:
* Create a new OSDevice on every getter
  * Make OSDevice depend on current OSPermissionSubscriptionState
 
Rename OSDevice to OSDeviceState

  * Rename isNotificationEnabled to areNotificationsEnabled
  * Rename isUserSubscribed to isPushDisabled and change logic

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1134)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDevice.java
@@ -1,21 +1,21 @@
 /**
  * Modified MIT License
- *
+ * <p>
  * Copyright 2020 OneSignal
- *
+ * <p>
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


jkasten2 (changes requested):
I don't believe this is a snapshot as `subscriptionStatus`, `permissionStatus`, and `emailSubscriptionStatus` are references that will be updated internally by the SDK.
Can we right a failing test to confirm this?

Jeasmine:
> I don't believe this is a snapshot as `subscriptionStatus`, `permissionStatus`, and `emailSubscriptionStatus` are references that will be updated internally by the SDK.
> Can we write a failing test to confirm this?

Good catch will add a test to check this

jkasten2 (approved):

PR #1136, opened 2020-09-07
Remove ExtNotificationWillShowInForegroundHandler

Jeasmine:
* There is no difference between ExtNotificationWillShowInForegroundHandler and AppNotificationWillShowInForegroundHandler
    responsibility, to avoid confusion ExtNotificationWillShowInForegroundHandler was removed

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1136)


jkasten2 (changes requested):
Since we only have one handler we should drop the "App" prefix.
Rename `AppNotificationWillShowInForegroundHandler` to `NotificationWillShowInForegroundHandler`

jkasten2 (approved):

PR #1137, opened 2020-09-07
Avoid displaying notification if user not calls display() 

Jeasmine:
* If display() is not called on notificationProcessing under NotificationProcessingHandler service, notification should not display and the foreground handler should not be called.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1137)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationExtender.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationExtender.java
@@ -133,27 +133,14 @@ public class OSNotificationExtender {
     * <br/><br/>
     * @see OSNotificationReceived#complete()
     */
-   void processNotification() {
-      // If the developer did not call display from notificationProcessing handler
-      if (!developerProcessed) {
-
+   void processNotification(boolean internalComplete) {
+      // If processNotification comes from an OneSignal flow, no callbacks/extenders/services called then continue with common flow.
+      // Check for display not called by developer to avoid duplicated handles under developer extenders service implementation failure
+      if (internalComplete && !developerProcessed) {
          // Save as processed to prevent possible duplicate calls from canonical ids
          boolean display = NotificationBundleProcessor.shouldDisplay(jsonPayload.optString("alert"));
-         if (!display) {
-            if (!isRestoring) {
-               OSNotificationGenerationJob notificationJob = new OSNotificationGenerationJob(context);
-               notificationJob.jsonPayload = jsonPayload;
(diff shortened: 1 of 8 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationExtender.java
```diff
@@ -133,34 +133,42 @@ OSNotificationDisplayedResult displayNotification() {
     * <br/><br/>
     * @see OSNotificationReceived#complete()
     */
-   void processNotification() {
-      // If the developer did not call display from notificationProcessing handler
-      if (!developerProcessed) {
-
+   void processNotification(boolean internalFlow) {
```
jkasten2, line 136: The `boolean` `internalComplete` was added to `OSNotificationReceived` in this PR. I think we should use this name here to keep it consistent.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
```diff
@@ -1553,6 +1554,167 @@ public void notificationProcessing(Context context, OSNotificationReceived notif
...
+   public static class NotificationExtensionService_notificationProcessingCompleteNotCalled implements OneSignal.NotificationProcessingHandler {
+
+      @Override
+      public void notificationProcessing(Context context, OSNotificationReceived notification) {
+         lastNotificationReceived = notification;
+
+         // Display called to show notification
+         OSNotificationDisplayedResult notificationDisplayedResult = notification.display();
+
+         // Complete not called to end NotificationProcessingHandler, depend on timeout to finish
```
jkasten2, line 1714: Should have `threadAndTaskWait();` or something here to wait for the timeout to fire.

jkasten2 (changes requested):

OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java
Jeasmine: We already have the 

```
FCMBroadcastReceiver_processBundle(blankActivity, getBaseNotifBundle());
threadAndTaskWait();
```

that threadAndTaskWait should cover the timeout fire 🤔

jkasten2: ah sorry my bad, this is a class to be used by the two tests above. Ignore my comment here.

jkasten2 (approved):

PR #1140, opened 2020-09-08
Encapsulate OSInAppMessageAction

Jeasmine:
* User can access OSInAppMessageAction via InAppMessageClickHandler
* Provide user only read access

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1140)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
@@ -35,50 +35,52 @@ public class OSInAppMessageAction {
      * Package-private to track which element was tapped to report to the OneSignal dashboard.
      */
     @NonNull
-    String clickId;
+    private String clickId;
 
     /**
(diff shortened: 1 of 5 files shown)


andrespelaezp (approved):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
```diff
@@ -116,6 +118,52 @@ private void parsePrompts(JSONObject json) throws JSONException {
...
+
+    public boolean isFirstClick() {
+        return firstClick;
+    }
+
+    void setFirstClick(boolean firstClick) {
+        this.firstClick = firstClick;
+    }
+
+    public boolean isClosesMessage() {
```
jkasten2, line 163: `isClosesMessage` doesn't make sense. Think this should be `doesCloseMessage`

jkasten2 (changes requested):

jkasten2 (approved):

PR #1147, opened 2020-09-14
Add toJSONObject to missing public classes

Jeasmine:
* Add toJSONObject for better wrapper interoperabiliyy

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1147)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDeviceState.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDeviceState.java
@@ -27,9 +27,11 @@
 
 package com.onesignal;
 
+import org.json.JSONObject;
+
 public class OSDeviceState {
 
-    private final boolean notificationEnabled;
+    private final boolean areNotificationsEnabled;
     private final boolean pushDisabled;
     private final boolean subscribed;
     private final String userId;
@@ -42,7 +44,7 @@ public class OSDeviceState {
         OSPermissionState permissionStatus = permissionSubscriptionState.permissionStatus;
         OSEmailSubscriptionState emailSubscriptionStatus = permissionSubscriptionState.emailSubscriptionStatus;
 
-        notificationEnabled = permissionStatus.getEnabled();
+        areNotificationsEnabled = permissionStatus.getEnabled();
         pushDisabled = !subscriptionStatus.getUserSubscriptionSetting();
         subscribed = subscriptionStatus.getSubscribed();
         userId = subscriptionStatus.getUserId();
@@ -57,7 +59,7 @@ public class OSDeviceState {
      * @return false if the user disabled notifications for the app, otherwise true
      */
     public boolean areNotificationsEnabled() {
-        return notificationEnabled;
+        return areNotificationsEnabled;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDeviceState.java
```diff
@@ -113,4 +115,22 @@ public String getEmailUserId() {
     public String getEmailAddress() {
         return emailAddress;
     }
+
+    public JSONObject toJSONObject() {
+        JSONObject mainObj = new JSONObject();
+
+        try {
+            mainObj.put("notification_enabled", notificationEnabled);
```
jkasten2, line 123: Should be `are_notifications_enabled`

jkasten2: Looks like we forgot to update the field name for `notificationEnabled` to be `areNotificationsEnabled` in a previous PR but can we do this here?

jkasten2 (changes requested):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDeviceState.java
Jeasmine: since boolean getters add "is" or "are" prefix to the method as style convention, I didn't add the "are" to the variable. Do we want it to be called areNotificationsEnabled?

Jeasmine: also if we are changing to are_notifications_enabled should we change subscribed to is_subscribed? push_disabled to is_push_disabled?

jkasten2: yes, let's change the boolean field to `areNotificationsEnabled` so it matches the getter method name exactly.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationAction.java
```diff
@@ -57,4 +59,18 @@ public ActionType getType() {
    public String getActionID() {
       return actionID;
    }
+
+   public JSONObject toJSONObject() {
+      JSONObject mainObj = new JSONObject();
+
+      try {
+         mainObj.put("type", type.ordinal());
+         mainObj.put("actionID", actionID);
```
jkasten2, line 68: `actionID` should be `actionId`

jkasten2 (changes requested):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationAction.java
Jeasmine: should I change de variables too? and getter to getActionId?

rgomezp: I think we should for consistency

jkasten2: I agree for consistently as well

jkasten2 (changes requested):
@Jeasmine Let's swith to camel case to match iOS.
https://github.com/OneSignal/OneSignal-iOS-SDK/pull/747#pullrequestreview-493045777

Jeasmine:
> @Jeasmine Let's switch to camel case to match iOS.
> [OneSignal/OneSignal-iOS-SDK#747 (review)](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/747#pullrequestreview-493045777)

iOS is adding "is" prefix to the names, is that ok?

jkasten2 (approved):

jkasten2:
> > @Jeasmine Let's switch to camel case to match iOS.
> > [OneSignal/OneSignal-iOS-SDK#747 (review)](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/747#pullrequestreview-493045777)
> 
> iOS is adding "is" prefix to the names, is that ok?

yes, good call. Approved!

PR #1149, opened 2020-09-15
Fix IAM not being redisplay by session duration

Jeasmine:
* Dynamic triggers weren't being scheduled on the second pass due to !dismissedMessages.contains(message.messageId) check being call before evaluateMessageTriggers, and this is not reset until setDataForRedisplay evaluates to true. If evaluateMessageTriggers is not called then dynamic triggers aren't scheduled
* Dynamic triggers weren't calling makeRedisplayMessagesAvailableWithTriggers, so no redisplay logic was being handled
* Add better debug logging

Improvements:

populateRedisplayMessagesTriggers, to avoid iterating over messages, iterate only over redisplay messages. Redisplay messages on the worst case will be O(n)

Use Cases:

Display IAM after the user is 30 seconds on the app every day.
Display IAM 5 seconds after last IAM, every day.

Warnings:

If the user has an IAM with session duration or last IAM with a short-timer period and short redisplay delay, this might end on IAM displaying multiple times in the same session.

Reset dynamic session launch time

* If the user puts the app on the background and opens the app after 30 seconds, on a new session, the session launch time was not being reset and due to that IAM with redisplay was being shown immediately.
* When a new session is started, reset session launch time. This will make IAM with session time trigger and redisplay work properly
* Refactor IAM Controller getController to avoid static reference

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1149)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
@@ -69,23 +69,6 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
     Date lastTimeInAppDismissed = null;
     private int htmlNetworkRequestAttemptCount = 0;
 
-    @Nullable
-    private static OSInAppMessageController sharedInstance;
-
-    public static synchronized OSInAppMessageController getController() {
-        OneSignalDbHelper dbHelper = OneSignal.getDBHelperInstance();
-
-        // Make sure only Android 4.4 devices and higher can use IAMs
-        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.JELLY_BEAN_MR2) {
-            sharedInstance = new OSInAppMessageDummyController(null);
-        }
-
-        if (sharedInstance == null)
-            sharedInstance = new OSInAppMessageController(dbHelper);
-
-        return sharedInstance;
-    }
-
     protected OSInAppMessageController(OneSignalDbHelper dbHelper) {
         messages = new ArrayList<>();
         dismissedMessages = OSUtils.newConcurrentSet();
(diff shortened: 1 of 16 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -189,19 +190,36 @@ private void processInAppMessageJson(@NonNull JSONArray json) throws JSONExcepti
         }
         messages = newMessages;
 
+        populateRedisplayMessagesTriggers();
         evaluateInAppMessages();
     }
 
+    private void populateRedisplayMessagesTriggers() {
+        for (OSInAppMessage redisplayMessage : redisplayedInAppMessages) {
```
jkasten2, line 198: This logic seems odd to me by using a foreach style logic but then looking up an idex. Couldn't this work with a `contains` check instead?

jkasten2 (approved):
Just a nit on making the for loop easier to read.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
Jeasmine: we need the index in order to update, nevertheless, this logic can be improved by using the processInAppMessageJson loop

jkasten2: 👍 This make it easier to read

jkasten2 (approved):

jkasten2:
I approved already but it looks like we want to add the session time reset to match iOS and possibly some other logic to sync with iOS.
https://github.com/OneSignal/OneSignal-iOS-SDK/pull/745

Jeasmine:
> I approved already but it looks like we want to add the session time reset to match iOS and possibly some other logic to sync with iOS.
> [OneSignal/OneSignal-iOS-SDK#745](https://github.com/OneSignal/OneSignal-iOS-SDK/pull/745)

Please review the second commit @jkasten2

jkasten2 (approved):

PR #1150, opened 2020-09-16
Fix equal trigger when using number

Jeasmine:
* Trim ceros for string equal matching

Use case:

Is user value is coming from an edit text that always returns a double. If the user puts 1, edit text if configure with decimals the result will be 1.0 , then equals between "1.0" and "1" won't match

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1150)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSTriggerController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSTriggerController.java
@@ -5,6 +5,7 @@ import androidx.annotation.Nullable;
 
 import com.onesignal.OSDynamicTriggerController.OSDynamicTriggerControllerObserver;
 
+import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSTriggerController.java
```diff
@@ -116,8 +117,16 @@ private boolean triggerMatchesFlex(@Nullable Object triggerValue, @NonNull Objec
...
-            return triggerMatchesStringValue(triggerValue.toString(), deviceValue.toString(), operator);
+        if (operator.checksEquality()) {
+            String triggerValueString = triggerValue.toString();
+            String deviceValueString = deviceValue.toString();
+            if (deviceValue instanceof Number) {
+                // User may have an input text that converts 5 to 5.0, we only care about the raw value on equals
+                DecimalFormat format = new DecimalFormat("0.#");
+                deviceValueString = format.format(deviceValue);
+            }
+            return triggerMatchesStringValue(triggerValueString, deviceValueString, operator);
```
jkasten2, line 128: It's not clear to me how this change makes the new `testMessageDisplayedAfterAddTriggerEqualWithStringVsNumber` test you added pass. Since `triggerValue` is `"5"` in your test and never get converted, as we are only formatting the device value.

Jeasmine: This problem started happening because the demo app input box converts integer to doubles, so whener we put 5 it will put 5.0, 1 -> 1.0 and so on.

Server numbers come always trimed or they came as strings. So we need to trim user input to match server value, that is why device value is the only one formatted.

The tests cover when the user adds 5.0 or 5.50 inputs and the server trigger value is a string "5" and "5.5". In order to match EQUALS, we need to trim ceros from the user input for the string equals to match.

Make sense?

jkasten2: got it, I think I flipped it around when reading this code

PR #1151, opened 2020-09-16
Fix foreground check when calling initWithContext with Activity

Jeasmine:
* Since appContext is always init with Application Context, handleActivityLifecycleHandler is neveer initializing inForeground
* Check for Activity lifecycle activity for foreground status

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1151)


jkasten2 (approved):
Looks good to me, however tests seem to be failing on travisCI

Jeasmine:
> Looks good to me, however tests seem to be failing on travisCI

Needed to make ActivityLifecycleHandler as instance static variable, for static reset. This also could be injected on the future too

jkasten2 (approved):

andrespelaezp (approved):

Issue #855, opened 2019-10-06
Egypt blocking OneSignal - OneSignal Android Studio failed to get Android Parameters issue


mtm19:
**Description:**
I am trying to integrate OneSignal to my android app after successfully integrating it to iOS with no problems. I have added android platform to my app in my OneSignal account and gone through all the steps. However, when I run the app, it just fails to get the parameters and returns userid = null, pushtoken is not null, permissionStatus.enabled = true, subscriptionStatus.subscribed = false. I have literally tried everything since 3 days but with no luck. It doesn't event recognize in the website that any device has requested but got an error. I have tried using simulators and real devices but it never worked. I have also tried to go to the link its trying to access through the computers browser and it gets the json data.

**Environment**
Android Studio 3.4.2 (Windows 10)
Android Studio 3.5.1 (Mac OS Mojave 10.14.6)
Pixel 3 XL Simulator running (Android 9, API 28) 
Pixel 2 XL Simulator running (Android 8.1, API 27)
Huawei Honor real device running (Android 7.0, API 24)
OneSignal.sdktype = native
OneSignal.VERSION = 031104

**Steps to Reproduce Issue:**
1. Integrate OneSignal SDK by going through all the steps
2. Runing the app and try to get information such as user id

**Code**

public class ApplicationClass extends Application {

    private static Context context;

    public static Context getContext() {
        return context;
    }

    @Override
    public void onCreate() {
        super.onCreate();
        context = getApplicationContext();
        
        // OneSignal Initialization
        OneSignal.setLogLevel(OneSignal.LOG_LEVEL.DEBUG, OneSignal.LOG_LEVEL.NONE);
        OneSignal.startInit(this).inFocusDisplaying(OneSignal.OSInFocusDisplayOption.Notification).unsubscribeWhenNotificationsAreDisabled(true)
                .init();

    }
}

**OneSignal Logs**
```
2019-10-06 12:15:11.498 30972-31049/com.mtm.edrequests W/OneSignal: OneSignalRestClient: null Error thrown from network stack. 
    java.net.SocketTimeoutException: failed to connect to onesignal.com/2606:4700::6810:d720 (port 443) from /fec0::1cb6:8803:6c1:556 (port 42418) after 60000ms
        at libcore.io.IoBridge.connectErrno(IoBridge.java:185)
        at libcore.io.IoBridge.connect(IoBridge.java:129)
[...]

PR #1152, opened 2020-09-16
Custom base url support added

luis901101:
This is a feature request to allow developers to opt in to use a proxy server as OneSignal api base url. It happen that OneSignal domains are blocked in some countries and by some ISPs… so I think this could be a good way to allow devs to use OneSignal.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1152)

### DIFF
--- a/OneSignalSDK/app/src/main/AndroidManifest.xml
+++ b/OneSignalSDK/app/src/main/AndroidManifest.xml
@@ -67,6 +67,7 @@
         <activity
             android:name=".BlankActivity"
             android:label="OneSignal Example"
+            android:exported="true"
             >
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
--- a/OneSignalSDK/onesignal/src/debug/AndroidManifest.xml
+++ b/OneSignalSDK/onesignal/src/debug/AndroidManifest.xml
@@ -13,7 +13,8 @@
 
         <receiver
             android:name="com.onesignal.GcmBroadcastReceiver"
-            android:permission="com.google.android.c2dm.permission.SEND" >
+            android:permission="com.google.android.c2dm.permission.SEND"
+            android:exported="true" >
             <!-- High priority so OneSignal payloads can be filtered from other GCM receivers if filterOtherGCMReceivers is enabled. -->
             <intent-filter android:priority="999" >
                 <action android:name="com.google.android.c2dm.intent.RECEIVE" />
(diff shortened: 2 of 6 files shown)


luis901101 (commented):
There are 2 unit tests failing **testNotificationReceivedWhenAppInFocus** and **shouldStillFireReceivedHandlerWhenNotificationExtenderServiceIsUsed** I just reverted my changes locally to check if this unittests passes on clear master and it failed as well...
So what can I do next?

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1152#-:-MHmQUQK5KdVGMH_84PV:b-w70v8d)* status: 0 of 2 files reviewed, all discussions resolved

jkasten2:
@luis901101 Thanks for the PR, you noted "OneSignal domains are blocked in some countries and by some ISPs ...". Could you let us know which ones? We would like to solve the issue by trying to reach out to them to unblock onesignal.com. As requiring work arounds such as setup up a proxy isn't ideal.

luis901101:
@jkasten2 thanks for your answer.
Currently I just can experienced this blocking issue from Cuba where I live, but I have found similar issues int he past like these ones: 
https://github.com/OneSignal/OneSignal-iOS-SDK/issues/554
https://github.com/OneSignal/OneSignal-Android-SDK/issues/855

I understand there are legal implications in this matter as commented here: https://wordpress.org/support/topic/country-or-region-ip-banned/

But there is something I can’t understand, for instance in my particular case “Cuba”, the only thing is blocked is the subscription to OneSignal I mean any api request will get a 403 error, nevertheless if I use a VPN only for the subscription request then I turn off the VPN and after that I receive push notifications normally for Android as well as iOS. This means the push notification service from Google Firebase and Apple are not blocked for Cuba, **so why not to allow devs to use a custom base url as reverse proxy?**

lfdominguez:
> @luis901101 Thanks for the PR, you noted "OneSignal domains are blocked in some countries and by some ISPs ...". Could you let us know which ones? We would like to solve the issue by trying to reach out to them to unblock onesignal.com. As requiring work arounds such as setup up a proxy isn't ideal.

The request is not unblock your domain, this is a problem of your ISP, the request is only add a customURL support that the SDK will contact vs the direct domain to allow to use proxies.

jkasten2:
There currently isn't a plan to support this feature in the SDK itself. Please contact OneSignal support if this still a feature you require.

PR #1167, opened 2020-09-28
Fix HMS register crash on old Huawei Devices

alirezaafkar:
After updating to the recent version of OneSignalSDK, we have some issues on old Huawei phones. I found the problem and reason for the error is that you have ApiException before checking HMS libraries on device. Thus I moved this part of code, and the error has resolved.

For reproducing the crash call this method directly on old Huawei devices(Without HMS).

```
new PushRegistratorHMS().registerForPush(this, "123", new PushRegistrator.RegisteredHandler() {
        @Override
    public void complete(String id, int status) {

    }
});
```

> com.onesignal.PushRegistratorHMS.registerForPush (PushRegistratorHMS.java:35)
> com.onesignal.OneSignal.registerForPushToken (OneSignal.java:1002)
> com.onesignal.OneSignal.access$1800 (OneSignal.java:86)
> com.onesignal.OneSignal$5.complete (OneSignal.java:1075)
> com.onesignal.OneSignalRemoteParams.processJson (OneSignalRemoteParams.java:188)
> com.onesignal.OneSignalRemoteParams.access$100 (OneSignalRemoteParams.java:12)
> com.onesignal.OneSignalRemoteParams$1.onSuccess (OneSignalRemoteParams.java:140)
> com.onesignal.OneSignalRestClient$5.run (OneSignalRestClient.java:270)
> java.lang.Thread.run (Thread.java:841)

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1167)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
@@ -7,7 +7,9 @@ import android.text.TextUtils;
 
 import com.huawei.agconnect.config.AGConnectServicesConfig;
 import com.huawei.hms.aaid.HmsInstanceId;
+import com.huawei.hms.common.ApiException;
 import com.huawei.hms.push.HmsMessaging;
+import com.huawei.hms.support.api.entity.core.CommonCode;
 
 import static com.onesignal.OneSignal.LOG_LEVEL;
 
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java
@@ -7,9 +7,7 @@ import android.text.TextUtils;
 
 import com.huawei.agconnect.config.AGConnectServicesConfig;
 import com.huawei.hms.aaid.HmsInstanceId;
-import com.huawei.hms.common.ApiException;
 import com.huawei.hms.push.HmsMessaging;
-import com.huawei.hms.support.api.entity.core.CommonCode;
 
 import static com.onesignal.OneSignal.LOG_LEVEL;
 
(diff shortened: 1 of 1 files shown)


jkasten2 (changes requested):
@alirezaafkar Thanks for digging in and making a PR. However there are few things to address before we can merge in.

**Can you confirm that there was a crash before this change?**
`OneSignal.Log(LOG_LEVEL.ERROR, "HMS ApiException getting Huawei push token!", e);` will print the full exception with a stacktrace however it from this catch and moves on without an issue.
https://github.com/OneSignal/OneSignal-Android-SDK/blob/master/OneSignalSDK/onesignal/src/main/java/com/onesignal/PushRegistratorHMS.java#L41

**What logic was change in this PR?**
It seems the `ApiException` catch was moved inside of `getHMSTokenTask` however it doesn't result in any different behavior. `getHMSTokenTask` is a private method too. Lastly this `PushRegistratorHMS` is package-private so it can't be called outside of the OneSignal SDK so this shouldn't be a factor either.

alirezaafkar:
@jkasten2 
I've attached an image from Crashlytics that shows crashes on our users' device.
About logic, I think the crash cause is ApiException; You use it before checking HMS libraries on the user's device.
Yes, getHMSTokenTask is private, but I changed it to the public for testing in the example project.

photo_2020-10-01_11-20-50

jkasten2:
@alirezaafkar All HMS API being called in `com.onesignal.PushRegistratorHMS.registerForPush` is wrappered in a try-catch with `ApiException` exception so it isn't possible for this specific exception to create a crash.

Is the stacktrace you reported here from your device? Can you extract the full stacktrace with the exact exception from the Google Play console to ensure it is the same one? Or from any other crash reporting tool you are using the track production crash.

alirezaafkar:
why don't you just test it? I've fixed it in my fork and we've no more crashes.

moallemi:
@jkasten2 any update on this?

jkasten2:
@alirezaafkar The devices in your screenshot are devices from around 2014 it seems. We can get one of these devices to test with but don't have one on hand so this will delay this PR. However if you can provide the exception name and message it should provide the details we need to catch it and take the correct action. Your original stacktrace is missing the first few lines where this would normally be included.

jkasten2:
Closing due to no other reports of this issue, as well as the age of the devices. Closing this PR as stale.

moallemi:
@jkasten2 Good job! 👏👏👏

PR #1168, opened 2020-09-28
Null exception setting email

andrespelaezp:
RemoteParams null validation was added once again, also a state for the dialog to be closed when the Set Email task is scheduled.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1168)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -262,6 +262,7 @@ public class OneSignal {
    public interface EmailUpdateHandler {
       void onSuccess();
       void onFailure(EmailUpdateError error);
+      void onTaskQueued();
    }
 
    private static EmailUpdateHandler emailUpdateHandler;
@@ -1299,6 +1300,8 @@ public class OneSignal {
       if (taskController.shouldQueueTaskForInit(OSTaskController.SET_EMAIL)) {
          logger.error("Waiting for remote params. " +
                  "Moving " + OSTaskController.SET_EMAIL + " operation to a pending task queue.");
+         if (callback != null)
+            callback.onTaskQueued();
          taskController.addTaskToQueue(new Runnable() {
             @Override
             public void run() {
@@ -1319,7 +1322,7 @@ public class OneSignal {
          return;
       }
 
-      if (getRemoteParams().useEmailAuth && emailAuthHash == null) {
+      if (getRemoteParams() != null && getRemoteParams().useEmailAuth && emailAuthHash == null) {
          String errorMessage = "Email authentication (auth token) is set to REQUIRED for this application. Please provide an auth token from your backend server or change the setting in the OneSignal dashboard.";
          if (callback != null)
             callback.onFailure(new EmailUpdateError(EmailErrorType.REQUIRES_EMAIL_AUTH, errorMessage));
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -262,7 +262,6 @@ public class OneSignal {
    public interface EmailUpdateHandler {
       void onSuccess();
       void onFailure(EmailUpdateError error);
-      void onTaskQueued();
    }
 
    private static EmailUpdateHandler emailUpdateHandler;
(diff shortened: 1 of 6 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1319,7 +1322,7 @@ public void run() {
          return;
       }
 
-      if (getRemoteParams().useEmailAuth && emailAuthHash == null) {
+      if (getRemoteParams() != null && getRemoteParams().useEmailAuth && emailAuthHash == null) {
```
Jeasmine, line 1325: remote params shouldn't be null at this point, because of the shouldQueueTaskForInit check. Why is it needed?

andrespelaezp: When there is a connection issue and the remote configuration is not previously setup, then `getRemoteParams()` is Null and this is the breaking point for the issue found.

Jeasmine: Sry I might be missing some part, but shouldQueueTaskForInit check for remote == null, if remote is still null because the network failed, then the task should be queue again, or why is it not being queue again?

andrespelaezp: It does get queued, but it also continues evaluating the statements below. Another approach could be a return statement within the task schedule block, what do you think?

Jeasmine: oooh I see, that method is missing the return! good catch, so that is why is failing

Jeasmine: No other changes should be needed but the return stmt

Jeasmine: compare with the other methods using shouldQueueTaskForInit

Jeasmine (commented):
The last commit is LGTM, we should agree if we want the user to know if the task got queue.
`void onTaskQueued();`
From my point of view, that is something the user doesn't have to know, the only information he wants to know if success and failure. Let me know what you think, also @jkasten2 thoughts?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1299,13 +1300,16 @@ public static void setEmail(@NonNull final String email, @Nullable final String
       if (taskController.shouldQueueTaskForInit(OSTaskController.SET_EMAIL)) {
          logger.error("Waiting for remote params. " +
                  "Moving " + OSTaskController.SET_EMAIL + " operation to a pending task queue.");
+         if (callback != null)
+            callback.onTaskQueued();
```
jkasten2, line 1304: I don't think we should change / add new APIs to address this bug fix PR. However I am open to a proposal on a new API to best handle this scenario for all our methods for consistency.

Just some feedback on the API itself though, I would expect a return value from `setEmail` instead of a callback that it has been queued.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
jkasten2: This is redundant code, the `shouldQueueTaskForInit` check and the `return` added above would cover this so I think we should remove the `null` check that was added in this PR.

jkasten2 (changes requested):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
andrespelaezp: Noted, I will keep this in mind for a future integration

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -1637,6 +1637,11 @@ public void onSuccess() {
       public void onFailure(OneSignal.EmailUpdateError error) {
          emailFiredFailure = error;
       }
+
+      @Override
```
jkasten2, line 1641: This also needs to be removed

jkasten2, line 1641: TravisCI is failing due to a Java compile error due to this interface method no longer existing.

jkasten2 (changes requested):

jkasten2 (approved):
Looks good to me. I recommend we squash all commit down to one before merging however.

PR #1171, opened 2020-10-01
Fix/iam issues

Jeasmine:
Iterate over message instead of redisplay

* For better readability
* Keep message list as single source of true

Init IAM with cached IAMs on new session

* When app is closed and opened if it is a new session, the IAMs will be delayed until on_session call, this end with the idea that IAMs stop working
* Always init app with cache IAMs

Make IAM not display under splas screenh when back pressing …

* If user implements splash with a handler timer activity and back press while displaying an IAM, IAM will be displayed under splash
* Add callback for when the app lost focus, dismiss IAM but not save it as dismissed this will allow the IAM to be displayed again until the user dismiss it
* If the user back press and come too quickly to the app, focus detection won't happen, IAM will still be shown under the splash screen

Review commit by commit

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1171)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java
@@ -50,7 +50,10 @@ class ActivityLifecycleHandler {
         void available(@NonNull Activity activity) {
         }
 
-        void stopped(WeakReference<Activity> reference) {
+        void stopped() {
+        }
+
+        void lostFocus() {
         }
     }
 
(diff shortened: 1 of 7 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java
```diff
@@ -50,7 +50,10 @@
         void available(@NonNull Activity activity) {
         }
 
-        void stopped(WeakReference<Activity> reference) {
+        void stopped() {
+        }
+
+        void lostFocus() {
         }
```
andrespelaezp, line 57: Do all of these functions have no body? Is this okay?

Jeasmine: yeah it's for the inheritance classes to implemente

andrespelaezp (approved):

jkasten2 (approved):

Issue #1117, opened 2020-08-18
NullPointerException at com.onesignal.BundleCompatPersistableBundle.getString


nurlansu:
**Description:**

I have encountered a RuntimeException caused by NullPointerException on the library.

**Environment**

1. OneSignal SDK: 3.15.1
1. Android SDK: 29
1. Gradle: 5.6.4
1. Gradle Android plugin: 3.6.1

**Steps to Reproduce Issue:**

Unfortunately it was not possible to reproduce the error. The error was reported by reports in Analytics.

**Anything else:**

It occurs mostly on Samsung devices.

**_Android 10_**

| Model | Count |
|----|----|
| Galaxy S20 Ultra 5G (z3s) | 39 |
| Galaxy S20 (x1s)          | 11 |
| Galaxy Note10 (d1)        | 11 |
| Galaxy S10+ (beyond2)     | 10 |
| Galaxy A70 (a70q)         | 9  |
| Galaxy S10 (beyond1)      | 7  |
| Galaxy Fold (winner)      | 7  |
| Galaxy A80 (r1q)          | 5  |
| Galaxy A6+ (a6plte)       | 3  |

**_Android 9_**

| Model | Count |
|----|----|
| Galaxy Note8 (greatlte) | 22 |

**_Stacktrace_**:

<details>
  <summary>Samsung Galaxy S20 Ultra 5G (z3s), 11008MB RAM, Android 10</summary>

```java
java.lang.RuntimeException:

1. at android.os.AsyncTask$4.done (AsyncTask.java:399)
1. at java.util.concurrent.FutureTask.finishCompletion (FutureTask.java:383)
1. at java.util.concurrent.FutureTask.setException (FutureTask.java:252)
1. at java.util.concurrent.FutureTask.run (FutureTask.java:271)
1. at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
1. at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
1. at java.lang.Thread.run (Thread.java:919)

Caused by: java.lang.NullPointerException:
1. at com.onesignal.BundleCompatPersistableBundle.getString (BundleCompat.java:91)
1. at com.onesignal.NotificationBundleProcessor.ProcessFromGCMIntentService (NotificationBundleProcessor.java:75)
[...]

PR #1173, opened 2020-10-02
NPE on notification received from BundleCompat

andrespelaezp:
Added exception handler to further investigate mBundle null state

* Created an exception called `OSBundleCompatException` 
* Throw it when accessing mBundle when it's null

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1173)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/BundleCompat.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/BundleCompat.java
@@ -41,7 +41,7 @@ public interface BundleCompat<T> {
    void putLong(String key, Long value);
    void putBoolean(String key, Boolean value);
    
-   String getString(String key);
+   String getString(String key) throws OSBundleCompatException;
    Integer getInt(String key);
    Long getLong(String key);
    boolean getBoolean(String key);
@@ -87,8 +87,11 @@ class BundleCompatPersistableBundle implements BundleCompat<PersistableBundle> {
    }
    
    @Override
-   public String getString(String key) {
-      return mBundle.getString(key);
+   public String getString(String key) throws OSBundleCompatException {
+      if (mBundle != null)
+         return mBundle.getString(key);
+      else
+         throw new OSBundleCompatException("unable to get mBundle " + this.toString());
    }
    
    @Override
@@ -200,6 +203,12 @@ class BundleCompatBundle implements BundleCompat<Bundle> {
    @Override
    public boolean getBoolean(String key, boolean value) {
      return mBundle.getBoolean(key, value);
+   }
+}
+
+class OSBundleCompatException extends Exception {
+   public OSBundleCompatException(String message) {
+      super(message);
    }
 }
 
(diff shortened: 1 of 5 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/BundleCompat.java
```diff
@@ -87,8 +87,11 @@ public void putBoolean(String key, Boolean value) {
    }
    
    @Override
-   public String getString(String key) {
-      return mBundle.getString(key);
+   public String getString(String key) throws OSBundleCompatException {
+      if (mBundle != null)
+         return mBundle.getString(key);
+      else
+         throw new OSBundleCompatException("unable to get mBundle " + this.toString());
```
Jeasmine, line 94: no toString method is implemented, is this printing all the data properly?

andrespelaezp: You make a good point, since mBundle is the only piece of information valuable but is null at the moment, should I just remove it then?

Jeasmine: We already know there are some cases where the bundle is null. I think will give more information if places were processFromFCMIntentService is called catch the exception, in order to give more information about the cases where the bundle is null

jkasten2 (changes requested):
I took a look at the stack traces from #1117 and it points to the following lines which are relevant to the issue.
* https://github.com/OneSignal/OneSignal-Android-SDK/blob/3.15.1/OneSignalSDK/onesignal/src/main/java/com/onesignal/GcmIntentJobService.java#L22
* https://github.com/OneSignal/OneSignal-Android-SDK/blob/3.15.1/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java#L75

The issue seems to be that `GcmIntentJobService` was started with an `Intent` and `Bundle` however these is no `BUNDLE_EXTRA` when we try to get it from `intent.getExtras().getParcelable(BUNDLE_EXTRA)`, see [GcmIntentJobService.java:20](https://github.com/OneSignal/OneSignal-Android-SDK/blob/3.15.1/OneSignalSDK/onesignal/src/main/java/com/onesignal/GcmIntentJobService.java#L20)
Since this the unexpected `null` entry we could throw here with a custom exception.

The call to `BundleCompat.getString` would be not what we want as we already know the exact key based on the stacktrace and the root of the issue is actually the bundle itself being `null`.

Fast forwarding to the 4.0.0 release it seems this has changed sufficiently.
As this new entry point stack.
* https://github.com/OneSignal/OneSignal-Android-SDK/blob/major_release_4.0.0/OneSignalSDK/onesignal/src/main/java/com/onesignal/FCMIntentJobService.java#L28
* https://github.com/OneSignal/OneSignal-Android-SDK/blob/major_release_4.0.0/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java#L360

There is a `null` check here now for he bundle as a whole
https://github.com/OneSignal/OneSignal-Android-SDK/blob/major_release_4.0.0/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java#L364

@andrespelaezp So it seems in 4.0.0 this has changed so this won't be an issue any more. It seems in 4.0.0 we don't use `BundleCompat` at all and have some dead code using it `startFCMService`. I can make a follow up PR to clean this PR up in 4.0.0.
@andrespelaezp Please review my notes here and I think we can close without merging. Then update #1117 letting them know to give the 4.0.0 beta a try which should address the issue.

PR #1174, opened 2020-10-02
Fix foreground check for wrappers

Jeasmine:
* If user init with activity context foreground check should work even if on resume callback is not fire

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1174)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java
@@ -61,7 +61,8 @@ class OSNotificationRestoreWorkManager {
         public Result doWork() {
             Context context = getApplicationContext();
 
-            OneSignal.initWithContext(context);
+            if (OneSignal.appContext == null)
+                OneSignal.initWithContext(context);
 
             if (!OSUtils.areNotificationsEnabled(context))
                 return Result.failure();
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -574,7 +574,7 @@ public class OneSignal {
       }
 
       if (appActivity != null && appActivity.get() != null)
-         initWithContext(appActivity.get());
+         init(appActivity.get());
       else
          init(appContext);
    }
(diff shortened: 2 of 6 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java
```diff
@@ -61,7 +61,8 @@ public NotificationRestoreWorker(@NonNull Context context, @NonNull WorkerParame
         public Result doWork() {
             Context context = getApplicationContext();
 
-            OneSignal.initWithContext(context);
+            if (OneSignal.appContext == null)
```
jkasten2, line 64: It seems that `Worker` would always have a `Context` so I won't expect a `null` check is need here.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -594,6 +599,8 @@ public static void initWithContext(@NonNull Context context) {
 
       boolean wasAppContextNull = (appContext == null);
       appContext = context.getApplicationContext();
+      if (context instanceof  Activity)
```
jkasten2, line 602: nit, extra space after `instanceof `

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -792,10 +792,23 @@ public void testInitWithContext_ActivityResumedBeforeInit() throws Exception {
...
       blankActivityController.resume();
+      OneSignal.initWithContext(blankActivity);
+      OneSignal.setAppId(ONESIGNAL_APP_ID);
+      threadAndTaskWait();
 
+      assertTrue(OneSignal_isInForeground());
+   }
+
+   @Test
+   public void testInitWithContextAppIdSet_ActivityResumedBeforeInit() throws Exception {
```
jkasten2, line 804: 👍 This covers the scenario we are looking to fix!

jkasten2 (changes requested):
See comments above

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java
Jeasmine: This is to avoid a new OneSignal.initWithContext(context); while the first initWithContext was called but not finished yet

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: good catch!

jkasten2 (approved):

PR #1177, opened 2020-10-05
Match OSSubscriptionState and OSPermissionState with OSDeviceState

Jeasmine:
* Rename properties, getters and setters to match OSDeviceState

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1177)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -419,7 +419,7 @@ public class OneSignal {
          return null;
 
       if (currentSubscriptionState == null) {
-         currentSubscriptionState = new OSSubscriptionState(false, getCurrentPermissionState(context).getEnabled());
+         currentSubscriptionState = new OSSubscriptionState(false, getCurrentPermissionState(context).areNotificationsEnabled());
          getCurrentPermissionState(context).observable.addObserver(currentSubscriptionState);
          currentSubscriptionState.observable.addObserverStrong(new OSSubscriptionChangedInternalObserver());
       }
@@ -2283,8 +2283,8 @@ public class OneSignal {
    }
 
    /**
-    * You can call this method with {@code false} to opt users out of receiving all notifications through
-    * OneSignal. You can pass {@code true} later to opt users back into notifications.
+    * You can call this method with {@code true} to opt users out of receiving all notifications through
+    * OneSignal. You can pass {@code false} later to opt users back into notifications.
     * @param disable whether to subscribe the user to notifications or not
     */
    public static void disablePush(final boolean disable) {
@@ -2305,8 +2305,8 @@ public class OneSignal {
       if (shouldLogUserPrivacyConsentErrorMessageForMethodName(OSTaskController.SET_SUBSCRIPTION))
          return;
 
-      getCurrentSubscriptionState(appContext).setUserSubscriptionSetting(disable);
-      OneSignalStateSynchronizer.setSubscription(disable);
+      getCurrentSubscriptionState(appContext).setPushDisabled(disable);
+      OneSignalStateSynchronizer.setSubscription(!disable);
    }
 
    /**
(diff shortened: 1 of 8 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2305,7 +2305,7 @@ public void run() {
       if (shouldLogUserPrivacyConsentErrorMessageForMethodName(OSTaskController.SET_SUBSCRIPTION))
          return;
 
-      getCurrentSubscriptionState(appContext).setUserSubscriptionSetting(disable);
+      getCurrentSubscriptionState(appContext).setPushDisabled(!disable);
```
Jeasmine, line 2308: with the properties names changes, is the 
`public static void disablePush(final boolean disable)` method readable? 

The comments say

   * You can call this method with {@code false} to opt users out of receiving all notifications through
    * OneSignal. You can pass {@code true} later to opt users back into notifications.
    * @param disable whether to subscribe the user to notifications or not

shouldn't true be for opt users out of receiving all notifications through and false otherwise?

jkasten2: ah yes, the comment is backwards and needs to be updated.

jkasten2, line 2308: Remove `!` we shouldn't be flipping the logic here. Your changes in `OSSubscriptionState.java` already account for this.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
```diff
@@ -74,7 +74,7 @@
     public static final String PREFS_ONESIGNAL_USERSTATE_DEPENDVALYES_ = "ONESIGNAL_USERSTATE_DEPENDVALYES_";
     public static final String PREFS_ONESIGNAL_USERSTATE_SYNCVALYES_ = "ONESIGNAL_USERSTATE_SYNCVALYES_";
     public static final String PREFS_ONESIGNAL_ACCEPTED_NOTIFICATION_LAST = "ONESIGNAL_ACCEPTED_NOTIFICATION_LAST";
-    public static final String PREFS_ONESIGNAL_SUBSCRIPTION_LAST = "ONESIGNAL_SUBSCRIPTION_LAST";
+    public static final String PREFS_ONESIGNAL_PUSH_DISABLED = "PREFS_ONESIGNAL_PUSH_DISABLED";
```
jkasten2, line 77: Must keep the string `"PREFS_ONESIGNAL_PUSH_DISABLED"` the same otherwise the value won't carry over when the app update from 3.x to 4.x.

I recommend we keep this the same for now since this is only internally used

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2305,7 +2305,7 @@ public void run() {
...
       OneSignalStateSynchronizer.setSubscription(disable);
```
jkasten2, line 2309: This line need to be updated to `OneSignalStateSynchronizer.setSubscription(!disable);` or we change the name `disablePush` as well an update the internals to match.

jkasten2 (changes requested):
See my comments above

Also #1145 changed the name from ` OneSignal.setSubscription` to ` OneSignal.disablePush` but it did not flip the boolean logic.
@Jeasmine Can you fix this in this PR or a follow up one?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: Given that subscription can also change email Synchronizer, seems more readable to keep it as setSubscription

jkasten2 (approved):

PR #1192, opened 2020-10-20
Notification Open should use Activity instead of Broadcast

Jeasmine:
* Switch to using NotificationOpenedActivity.class in GenerateNotification.java
* Remove Intent flags from startOrResumeApp and use Activity Context instead of the normal context
* Update tests to account for the new behavior

Handle delete intent with broadcast receiver …

* Using activities for handling delete can end on some weird behavior on the UI

Tested on API 21, 25, 29

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1192)

demo


OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
```diff
@@ -133,28 +122,20 @@ private static CharSequence getTitle(JSONObject fcmJson) {
...
-          .putExtra(BUNDLE_KEY_ANDROID_NOTIFICATION_ID, notificationId)
-          .putExtra("dismissed", true);
-
-      if (openerIsBroadcast)
-         return intent;
-      return intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_MULTIPLE_TASK | Intent.FLAG_ACTIVITY_NO_ANIMATION);
+      return new Intent(currentContext, notificationOpenedClass)
+              .putExtra(BUNDLE_KEY_ANDROID_NOTIFICATION_ID, notificationId)
+              .putExtra("dismissed", true)
+              .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_MULTIPLE_TASK | Intent.FLAG_ACTIVITY_NO_ANIMATION);
```
jkasten2, line 138: Mike tried to switch this notification dismiss action to an `Activity` before in the past and it resulted in a visual flicker. Do these flags completely prevent that?

OneSignalSDK/onesignal/src/main/AndroidManifest.xml
```diff
@@ -104,7 +104,9 @@
             </intent-filter>
         </receiver>
 
-        <activity android:name="com.onesignal.NotificationOpenedActivity"/>
+        <activity
+          android:name="com.onesignal.NotificationOpenedActivity"
+          android:noHistory="true" />
```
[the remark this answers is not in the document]
Jeasmine: Couldn't found any difference between having and not having this attribute

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
Jeasmine: good catch! there is no flicker but the process is being closed

jkasten2: We might have to use a `BroadcastReceiver` for the dismisses. If we use an `Activity` for dismisses it could result in odd focus behavior in the app with extra onPause and onResume calls each time a notification is swiped away we would want to avoid.

Jeasmine: Yeah, seems like the most reasonable solution, using intent for dismissing tracking, ends on weirds behaviors

OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotification.java
```diff
@@ -122,6 +130,16 @@ private static CharSequence getTitle(JSONObject fcmJson) {
...
    private static PendingIntent getNewActionPendingIntent(int requestCode, Intent intent) {
+     return  getNewActionPendingIntent(false, requestCode, intent);
+   }
+
+   /**
+    * Notification delete is processed by Broadcast Receiver to avoid creation of activities that can end
+    * on weird UI interaction
+    */
+   private static PendingIntent getNewActionPendingIntent(boolean openerIsBroadcast, int requestCode, Intent intent) {
+      if (openerIsBroadcast)
```
jkasten2, line 141: We can safely remove the `openerIsBroadcast` check as the open action will always be an `Activity`

jkasten2 (changes requested):

jkasten2 (approved):

PR #1193, opened 2020-10-20
Add notification duplicate check under Workers

Jeasmine:
* If the same notification arrives twice, if the first notification did not end processing, the notification will be processed twice
* Keep notificationIds in memory for short time duplicates, then we rely on databse duplicate check

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1193)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
@@ -10,12 +10,12 @@ import androidx.work.WorkManager;
 import androidx.work.Worker;
 import androidx.work.WorkerParameters;
 
-import com.google.android.gms.common.util.Strings;
-
 import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.util.Set;
+
+import static com.onesignal.OSUtils.isStringNotEmpty;
 
 class OSNotificationWorkManager {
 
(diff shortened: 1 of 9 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
```diff
@@ -10,16 +10,28 @@
 import androidx.work.Worker;
 import androidx.work.WorkerParameters;
 
+import com.google.android.gms.common.util.Strings;
```
jkasten2, line 13: This won't be present if someone decides to omit Google Play Services from their app for FireOS or Huawei build of their app. I recommend using `import android.text.TextUtils` instead as this is in AndroidX. Then we can use `TextUtils.isEmpty` instead of `Strings.isEmptyOrWhitespace`.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
```diff
@@ -33,6 +45,19 @@ static void beginEnqueueingWork(Context context, String osNotificationId, int an
                 .setInputData(inputData)
                 .build();
 
+        // Duplicate control
+        // Keep in memory in going processed notifications, to avoid fast duplicates that already finished work process but are not completed yet
+        // enqueueUniqueWork might not be enough, if the work already finished then the duplicate notification work might be queued again
+        if (!Strings.isEmptyOrWhitespace(osNotificationId)) {
+            if (notificationIds.contains(osNotificationId)) {
```
jkasten2, line 52: Can we combine this logic with `OneSignal.isDuplicateNotification` somehow? It isn't obvious that we have two ways of checking for duplicates.

jkasten2 (changes requested):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java
Jeasmine: Good catch!

jkasten2 (changes requested):
@Jeasmine Looks like we forgot to update this 

```
> Task :onesignal:compileReleaseKotlin FAILED
e: /home/travis/build/OneSignal/OneSignal-Android-SDK/OneSignalSDK/onesignal/src/main/java/com/onesignal/ADMMessageHandlerJob.kt: (12, 85): Type mismatch: inferred type is Context? but Context was expected
```

https://travis-ci.org/github/OneSignal/OneSignal-Android-SDK/builds/740942674#L1608

Jeasmine:
> @Jeasmine Looks like we forgot to update this
> 
> ```
> > Task :onesignal:compileReleaseKotlin FAILED
> e: /home/travis/build/OneSignal/OneSignal-Android-SDK/OneSignalSDK/onesignal/src/main/java/com/onesignal/ADMMessageHandlerJob.kt: (12, 85): Type mismatch: inferred type is Context? but Context was expected
> ```
> 
> https://travis-ci.org/github/OneSignal/OneSignal-Android-SDK/builds/740942674#L1608

oh my bad! updated

jkasten2 (approved):

PR #1196, opened 2020-10-23
Enable background player updates

Jeasmine:
* If user send tags under ServiceExtension implementation, player update should happen
* Add sendTags on background test

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1196)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -371,7 +371,7 @@ public class OneSignal {
 
    @NonNull private static OSUtils osUtils = new OSUtils();
 
-   private static boolean registerForPushFired, locationFired, promptedLocation, getTagsCall, waitingToPostStateSync;
+   private static boolean registerForPushFired, locationFired, promptedLocation, getTagsCall, waitingToPostStateSync, androidParamsRequestStarted;
 
    private static LocationController.LocationPoint lastLocationPoint;
 
@@ -669,7 +669,7 @@ public class OneSignal {
          // Wrapper SDK's call init twice and pass null as the appId on the first call
          //  the app ID is required to download parameters, so do not download params until the appID is provided
          if (lastAppId != null && context != null)
-            makeAndroidParamsRequest(lastAppId, getUserId());
+            makeAndroidParamsRequest(lastAppId, getUserId(), false);
          return;
       }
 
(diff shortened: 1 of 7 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -815,25 +816,27 @@ private static void handleAmazonPurchase() {
...
+
          sessionManager.attemptSessionUpgrade(getAppEntryState());
       }
       getInAppMessageController().initWithCachedInAppMessages();
 
       // We still want register the user to OneSignal if the SDK was initialized
       //   in the background for the first time.
       if (!inForeground && hasUserId())
-         return;
+         logger.debug("doSessionInit on background with already registered user");
```
emawby, line 839: The return here was removed was that intentional?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1224,16 +1234,18 @@ private static int getTimeZoneOffset() {
...
          "registerUser:" +
          "registerForPushFired:" + registerForPushFired +
          ", locationFired: " + locationFired +
          ", remoteParams: " + getRemoteParams() +
          ", appId: " + appId
       );
 
-      if (!registerForPushFired || !locationFired || getRemoteParams() == null || appId == null)
+      if (!registerForPushFired || !locationFired || getRemoteParams() == null || appId == null) {
+         logger.debug("registerUser not possible");
```
emawby, line 1246: I think it would be useful to print the values here so we know why registerUser isn't possible since there are 4 possibilities

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: yeah, this was preventing the background update

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: we already had the 
```

logger.debug(
         "registerUser:" +
         "registerForPushFired:" + registerForPushFired +
         ", locationFired: " + locationFired +
         ", remoteParams: " + getRemoteParams() +
         ", appId: " + appId
      );
```

didn't want to repeat the same data

emawby (approved):

jkasten2 (approved):

PR #1198, opened 2020-10-27
Fix focus update not schedule when app backgrounded

Jeasmine:
Make an abstract class for SyncService common methods
    
    * We want to re-use SyncServiceUtils background sync methods

Schedule focus delay with service
    
    * Focus delay was a postDelay that if the application ended before postDelay ran then on_focus schedule won't happen
    * Create OSFocusDelaySync and reuse OSBackgroundSync
    * Save unsent time and influences as fast as the user left the application
    * Create OSFocusTimeProcessorFactory to get proper processor from influences

New flaky test testAppOnFocus_wontContainOutcomeData_withOutcomeEventFlagsDisabled
Review commit by commit

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1198)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -316,6 +316,10 @@ static boolean isInitDone() {
    static boolean isInForeground() {
       return inForeground;
    }
+   public static void setInForeground(boolean inForeground) {
```
jkasten2, line 319: We don't want this to be a public method, it should only be called internally by the SDK.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/TestHelpers.java
```diff
@@ -499,26 +503,97 @@ public static void assertMainThread() {
...
+
+   public static void assertNextJob(Class jobClass) {
+      assertNextJob(jobClass, 0);
+   }
+
+   public static void assertNextJob(Class jobClass, int index) {
+      assertEquals(jobClass.getName(), getJob(index).getService().getClassName());
+   }
+
+   public static void pauseActivity(ActivityController activityController) throws Exception {
```
jkasten2, line 581: 👍 Nice helper!

It was a lot of line changes, would be nice in the future if this it was its own commit. That way it is easier to review the other changes in smaller chunks.

jkasten2 (changes requested):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: good catch! this was actually only for debugging

jkasten2 (approved):

jkasten2 (commented):
The test `testAfterLastInAppTimeIsDisplayed` failed on both TravisCI runs.
@Jeasmine Does it fail locally for you on this branch?

Jeasmine:
> The test `testAfterLastInAppTimeIsDisplayed` failed on both TravisCI runs.
> @Jeasmine Does it fail locally for you on this branch?

If I ran all together sometimes it fails but if I run it individually it doesn't fail

jkasten2 (approved):

PR #1206, opened 2020-11-10
Add Main thread check under OSNotificationReceivedEvent

Jeasmine:
* Flutter invoked methods are runned under MainThread, OSNotificationReceivedEvent complete method ends calling a MainThread check that crashes.
* Add MainThread check under OSNotificationReceivedEvent complete method
* If running on MainThread create a new thread to complete the notification

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1206)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationReceivedEvent.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationReceivedEvent.java
@@ -67,7 +67,7 @@ public class OSNotificationReceivedEvent {
     * @param notification can be null to omit displaying the notification,
     *                     or OSMutableNotification to modify the notification to display
     */
-   public synchronized void complete(@Nullable OSNotification notification) {
+   public synchronized void complete(@Nullable final OSNotification notification) {
       timeoutHandler.destroyTimeout(timeoutRunnable);
 
       if (isComplete) {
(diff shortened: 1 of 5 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationReceivedEvent.java
```diff
@@ -77,6 +77,21 @@ public synchronized void complete(@Nullable OSNotification notification) {
 
       isComplete = true;
 
+      // Logic for Wrappers
```
jkasten2, line 80: Logic isn't specifically for wrappers. A native developer could call the complete from the main thread as well.

jkasten2 (changes requested):
Seeing additional tests failing around notifications that seem they are related to this change. Possibly due to the introduction of this new thread and not waiting on it.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationReceivedEvent.java
Jeasmine: yeah, but will need to specially make it from the main thread, on native the service extender is being called from a worker

Jeasmine:
> Seeing additional tests failing around notifications that seem they are related to this change. Possibly due to the introduction of this new thread and not waiting on it.

Tests fixed, thread wait is needed due to the new thread creation and roboelectric running on main thread

jkasten2 (approved):

PR #1210, opened 2020-11-12
Add auth hash to external Id

Jeasmine:
* Add auth to setExternalId
* Add tests for new functionality

Needs name params validation

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1210)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -200,8 +200,31 @@ public class OneSignal {
       public String getMessage() { return message; }
    }
 
+   public enum ExternalIdErrorType {
+      REQUIRES_EXTERNAL_ID_AUTH, INVALID_OPERATION, NETWORK
+   }
+
+   public static class ExternalIdError {
+      private ExternalIdErrorType type;
+      private String message;
+
+      ExternalIdError(ExternalIdErrorType type, String message) {
+         this.type = type;
+         this.message = message;
+      }
+
+      public ExternalIdErrorType getType() {
+         return type;
+      }
+
+      public String getMessage() {
+         return message;
+      }
+   }
+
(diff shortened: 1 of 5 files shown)


jkasten2 (approved):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1565,8 +1595,20 @@ public void run() {
                return;
             }
 
+            if (remoteParams != null && remoteParams.useEmailAuth && externalIdAuthHash == null) {
```
jkasten2, line 1598: `remoteParams.useEmailAuth` should be ` remoteParams.useUserIdAuth`

jkasten2 (changes requested):
Need to add `require_user_id_auth` to remote params

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -1565,8 +1595,20 @@ public void run() {
...
+               String errorMessage = "External Id authentication (auth token) is set to REQUIRED for this application. Please provide an auth token from your backend server or change the setting in the OneSignal dashboard.";
+               if (completionCallback != null)
+                  completionCallback.onFailure(new ExternalIdError(ExternalIdErrorType.REQUIRES_EXTERNAL_ID_AUTH, errorMessage));
+               Log(LOG_LEVEL.ERROR, errorMessage);
+               return;
+            }
+
+            String internalExternalIdAuthHash = externalIdAuthHash;
```
emawby, line 1606: Can we change the name of internalExternalIdAuthHash? Is the only change we are making is making it all lower case? If that is the case maybe include the lowercase aspect in the name?

emawby (changes requested):

jkasten2 (approved):
`external_user_id_auth_hash` is correctly added here so I think it looks good to me.

PR #1225, opened 2020-12-01
Fix failing test

Jeasmine:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1225)


jkasten2 (changes requested):
Thanks for fixing, looks like this PR fixes `shouldHandleBasicNotifications`.

Another that always fails on both travisCI (2 runs total) and locally (~10 runs). 
`testNotificationWillShowInForegroundHandler_notCallCompleteShowsNotificationAfterTimeout`

Looks like we need to add `ShadowNotificationReceivedEvent.class` to this one too.

jkasten2 (approved):
Looks good to me!

These tests look to be still flaky but you fixed the ones always failing so fixes for these can be in a follow up PR later.
* `testAppOnFocus_wontContainOutcomeData_withOutcomeEventFlagsDisabled`
* `sendsOnFocusToEmail`

PR #1233, opened 2020-12-06
proguard rule OSRemoteNotificationReceivedHandler

jkasten2:
* Added OSRemoteNotificationReceivedHandler to consumer-proguard-rules.pro with a comment on why.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1233)


Jeasmine (approved):
Good catch!

PR #1239, opened 2020-12-10
Iam carousel page impressions

emawby:
This PR adds SDK support for individual IAM Carousel page impressions.

It adds a new model `OSInAppMessagePage.kt` and event type pageChange that reads the pageId from the new page_change event that is received from the carousel's javascript.

To send the page impression we use the new pageImpression endpoint using the new request `fireRESTCallForPageChange`
Page impressions are only sent once per Impression of a Carousel. The ids are tracked in a new set `viewedPageIds` on the `OSInAppMessage` object

Additionally we receive the pageId for click events, but do not currently use them.

Question: Do we need to save impressioned pages in OneSignalPrefs like we do for normal impression? (commented out code right now)

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1239)

### DIFF
--- a/Examples/OneSignalDemo/local.properties
+++ b/Examples/OneSignalDemo/local.properties
@@ -4,6 +4,6 @@
 # Location of the SDK. This is only used by Gradle.
 # For customization when using a Version Control System, please read the
 # header note.
-#Tue Jul 23 19:22:28 PDT 2019
+#Wed Sep 09 11:15:20 PDT 2020
 ndk.dir=/Users/mikechoch/Library/Android/sdk/ndk-bundle
-sdk.dir=/Users/mikechoch/Library/Android/sdk
+sdk.dir=/Users/elliotmawby/Library/Android/sdk
--- a/Examples/OneSignalDemo/local.properties
+++ b/Examples/OneSignalDemo/local.properties
@@ -4,6 +4,6 @@
 # Location of the SDK. This is only used by Gradle.
 # For customization when using a Version Control System, please read the
 # header note.
-#Wed Sep 09 11:15:20 PDT 2020
+#Tue Jul 23 19:22:28 PDT 2019
 ndk.dir=/Users/mikechoch/Library/Android/sdk/ndk-bundle
-sdk.dir=/Users/elliotmawby/Library/Android/sdk
+sdk.dir=/Users/mikechoch/Library/Android/sdk
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
@@ -17,6 +17,7 @@ public class OSInAppMessageAction {
     private static final String ID = "id";
     private static final String NAME = "name";
     private static final String URL = "url";
+    private static final String PAGE_ID = "page_id";
     private static final String URL_TARGET = "url_target";
     private static final String CLOSE = "close";
     private static final String CLICK_NAME = "click_name";
@@ -53,6 +54,12 @@ public class OSInAppMessageAction {
     private String clickUrl;
 
     /**
+     * UUID for the page in an IAM Carousel
+     */
+    @Nullable
+    private String pageId;
+
+    /**
      * Outcome for action
      */
     @NonNull
@@ -83,6 +90,7 @@ public class OSInAppMessageAction {
         clickId = json.optString(ID, null);
         clickName = json.optString(NAME, null);
         clickUrl = json.optString(URL, null);
+        pageId = json.optString(PAGE_ID, null);
         urlTarget = OSInAppMessageActionUrlType.fromString(json.optString(URL_TARGET, null));
         if (urlTarget == null)
             urlTarget = OSInAppMessageActionUrlType.IN_APP_WEBVIEW;
@@ -123,6 +131,11 @@ public class OSInAppMessageAction {
     @Nullable
     public String getClickName() {
         return clickName;
+    }
+
+    @NonNull
+    String getPageId() {
+        return pageId;
     }
 
     @Nullable
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
@@ -17,7 +17,7 @@ public class OSInAppMessageAction {
     private static final String ID = "id";
     private static final String NAME = "name";
     private static final String URL = "url";
-    private static final String PAGE_ID = "page_id";
+    private static final String PAGE_ID = "pageId";
     private static final String URL_TARGET = "url_target";
     private static final String CLOSE = "close";
     private static final String CLICK_NAME = "click_name";
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
@@ -133,7 +133,6 @@ public class OSInAppMessageAction {
         return clickName;
     }
 
-    @NonNull
     String getPageId() {
         return pageId;
     }
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
@@ -286,9 +286,6 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
         if (message.isPreview) {
             return;
         }
-        if (message.getViewedPageIds().contains(newPage.getPageId())) {
-            return;
-        }
         fireRESTCallForPageChange(message, newPage);
     }
 
@@ -426,7 +423,6 @@ class OSInAppMessageController implements OSDynamicTriggerControllerObserver, OS
             return;
 
         final String pageId = page.getPageId();
-
 
         // Never send multiple page impressions for the same message UUID unless that page change is from an IAM with redisplay
         if (message.getViewedPageIds().contains(pageId))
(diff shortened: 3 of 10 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -409,6 +420,53 @@ else if (action.getUrlTarget() == OSInAppMessageAction.OSInAppMessageActionUrlTy
...
+                put("variant_id", variantId);
+                put("device_type", new OSUtils().getDeviceType());
+                put("page_id", pageId);
+            }};
+
+            OneSignalRestClient.post("in_app_messages/" + message.messageId + "/pageImpression", json, new ResponseHandler() {
+                @Override
+                void onSuccess(String response) {
+                    printHttpSuccessForInAppMessageRequest("page impression", response);
+                    /*OneSignalPrefs.saveStringSet(
```
jkasten2, line 450: Looks like this is a TODO yet?

jkasten2 (changes requested):

Examples/OneSignalDemo/local.properties
```diff
@@ -4,6 +4,6 @@
 # Location of the SDK. This is only used by Gradle.
 # For customization when using a Version Control System, please read the
 # header note.
-#Tue Jul 23 19:22:28 PDT 2019
```
Jeasmine, line 7: this should be a ignored file

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
```diff
@@ -125,6 +133,11 @@ public String getClickName() {
         return clickName;
     }
 
+    @NonNull
```
Jeasmine, line 136: this is marked as notNull but 

`pageId = json.optString(PAGE_ID, null);` it could end being null, maybe set a default value? tr change to Nullable

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -409,6 +420,53 @@ else if (action.getUrlTarget() == OSInAppMessageAction.OSInAppMessageActionUrlTy
         }
     }
 
+    private void fireRESTCallForPageChange(@NonNull final OSInAppMessage message, @NonNull final OSInAppMessagePage page) {
+        final String variantId = variantIdForMessage(message);
+        if (variantId == null)
+            return;
+
+        final String pageId = page.getPageId();
+
```
Jeasmine, line 429: remove extra enter

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
emawby: Yes this was my question at the end of the PR do we need to save the ids to the prefs?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -409,6 +420,53 @@ else if (action.getUrlTarget() == OSInAppMessageAction.OSInAppMessageActionUrlTy
...
+
+        // Never send multiple page impressions for the same message UUID unless that page change is from an IAM with redisplay
+        if (message.getViewedPageIds().contains(pageId))
```
Jeasmine, line 432: this is checked under onPageChanged too, is that ok?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessagePage.kt
```diff
@@ -0,0 +1,29 @@
...
+class OSInAppMessagePage constructor(jsonObject: JSONObject) {
+    var pageId: String? = null
+    var pageIndex: String? = null
+
+    init {
+        pageId = jsonObject.optString(PAGE_ID, null)
+        pageIndex = jsonObject.optString(PAGE_INDEX, null)
+    }
+
+    fun toJSONObject(): JSONObject? {
```
Jeasmine, line 19: is the optional needed?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
```diff
@@ -616,4 +616,19 @@ public void testOnMessageWasShown() throws Exception {
         assertEquals(1, iamImpressionRequest.payload.get("device_type"));
         assertEquals(true, iamImpressionRequest.payload.get("first_impression"));
     }
+
+    @Test
+    public void testOnPageChanged() throws Exception {
+        threadAndTaskWait();
```
Jeasmine, line 622: is this needed?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
```diff
@@ -616,4 +616,19 @@ public void testOnMessageWasShown() throws Exception {
...
+
+        OneSignalPackagePrivateHelper.onPageChanged(message, InAppMessagingHelpers.buildTestPageJson());
```
Jeasmine, line 624: is ok to test without mocking the IAM display?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
Jeasmine: The only scenario that I think needs this.
1 - Opening app 
2 - Display IAM
3 - Impression sent
4 - Close application (swipe away) without dismiss IAM
5 - Open app
6 - IAM should display again
7 - Impression shouldn't be sent again

emawby: Cool confirmed that this is an issue so I will store them in the prefs

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
emawby: Now only checking in the rest method call for consistency with iOS and better separation of concerns

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessagePage.kt
emawby: nope fixed

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
emawby: Yes we test without mocking the iam display for clicks as well and they should be entirely independent

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
emawby: yes it crashes without it

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
emawby: fixed!

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageAction.java
emawby: fixed!

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
Jeasmine: oh, weird 🤔  test setup should not break tests, can you point out which error it is?

Examples/OneSignalDemo/local.properties
Jeasmine: can we remove this file from the PR?

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -409,6 +429,62 @@ else if (action.getUrlTarget() == OSInAppMessageAction.OSInAppMessageActionUrlTy
...
+        final String variantId = variantIdForMessage(message);
+        if (variantId == null)
+            return;
+
+        final String pageId = page.getPageId();
+
+        final String messagePrefixedPageId = message.messageId + pageId;
+
+        // Never send multiple page impressions for the same message UUID unless that page change is from an IAM with redisplay
+        if (message.getViewedPageIds().contains(pageId) || viewedPageIds.contains(messagePrefixedPageId)) {
```
Jeasmine, line 450: same question as on iOS, see https://github.com/OneSignal/OneSignal-iOS-SDK/pull/735#discussion_r542662673

emawby: same fix as ios. only tracking globally now, but it is still per inapp since we don't want them to affect each other (in local testing I was getting duplicate page ids for different inapps which is why I prefixed the page id with the message id)

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessagingUnitTests.java
emawby: Ah sorry the crash can be fixed, but you need to initialize to have the player id etc for the request

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java
```diff
@@ -494,6 +568,8 @@ private void setDataForRedisplay(OSInAppMessage message) {
 
                 dismissedMessages.remove(message.messageId);
                 impressionedMessages.remove(message.messageId);
+                viewedPageIds.clear();
```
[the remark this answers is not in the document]
emawby: done!

Jeasmine (approved):

PR #1246, opened 2020-12-22
Fix Displaying IAMs past end time

emawby:
This change involves reading in end_time from JSON when creating an IAM
so that we can decide to not show the IAM (Cached case) if the IAM
should be finished. The date comes in iso8601 format.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1246)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
@@ -96,7 +96,7 @@ class OSInAppMessage {
             return null;
         }
 
-        SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US);
+        SimpleDateFormat format = OneSignalSimpleDateFormat.iso8601Format();
         try {
             Date date = format.parse(endTimeString);
             return date;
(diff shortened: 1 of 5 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -76,11 +82,31 @@
...
 
+    private Date parseEndTimeJson(JSONObject json) {
+        String endTimeString;
+        try {
+            endTimeString = json.getString(END_TIME);
+        } catch (JSONException e) {
+            return null;
+        }
+
+        SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US);
```
Jeasmine, line 99: can we put this on a constant?

Jeasmine (commented):
LGTM, nit comment

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
Jeasmine: Can we have the same iOS object https://github.com/OneSignal/OneSignal-iOS-SDK/pull/841#discussion_r547501668 in here?

emawby: created a OneSignalSimpleDateFormatter class with a static method to get the iso8601 format

Jeasmine (approved):

PR #1247, opened 2020-12-24
Fix IAM redisplay when no dismiss

Jeasmine:
* If user navigates with backpress while an IAM is being display, IAM should continue displaying until user dismiss it.
* Create new scenario for displaying IAM again when changing activity

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1247)


jkasten2 (approved):

jkasten2:
Checked failing tests, they are unrelated flaky tests.

PR #1264, opened 2021-01-26
Fix crash when receiving IAM on android

gonzalonarbaiz:
The range that we were using was causing that gradle build to use version 20.2.2 of the firebase messaging library, which was making the app crash. 
That error is fixed on later versions of the library.
This error is reported in https://firebase.google.com/support/release-notes/android#messaging_v20-2-2
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1264)

### DIFF
--- a/OneSignalSDK/onesignal/build.gradle
+++ b/OneSignalSDK/onesignal/build.gradle
@@ -5,7 +5,7 @@ ext {
 
     androidXVersion = '[1.0.0, 1.99.99]'
     androidWorkVersion = '[2.0.0, 2.99.99]'
-    firebaseMessagingVersion = '[19.0.0, 20.99.99]'
+    firebaseMessagingVersion = '[19.0.0, 21.99.99]'
     googleGMSVersion = '[17.0.0, 17.99.99]'
 }
 


OneSignalSDK/onesignal/build.gradle
```diff
@@ -5,7 +5,7 @@ ext {
 
     androidXVersion = '[1.0.0, 1.99.99]'
     androidWorkVersion = '[2.0.0, 2.99.99]'
-    firebaseMessagingVersion = '[19.0.0, 20.99.99]'
+    firebaseMessagingVersion = '[19.0.0, 21.99.99]'
```
Jeasmine, line 8: Can you put more information about the error and what specific under this version solve the problem?

gonzalonarbaiz: Yes, there is a known issue https://firebase.google.com/support/release-notes/android#messaging_v20-2-2 and we are asked not to use this version. For some reason I couldn't find when the version range is set to [19.0.0, 20.99.99] on my builds didn't go for 20.3.0 which already has a fix for that.

Jeasmine: @jkasten2 FYI

jkasten2 (approved):

Jeasmine (approved):

Jeasmine:
Seems like users are facing this issue check https://github.com/OneSignal/OneSignal-Android-SDK/issues/1267 this should be merge and ready for next release

PR #1268, opened 2021-01-29
Feature/add onesignal suppress launch urls

gonzalonarbaiz:
* Added the suppression to keep 1:1 behavior with iOS launch URL suppression. 
* added test for new metadata.

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1268)

### DIFF
--- a/Examples/OneSignalDemo/local.properties
+++ b/Examples/OneSignalDemo/local.properties
@@ -4,6 +4,6 @@
 # Location of the SDK. This is only used by Gradle.
 # For customization when using a Version Control System, please read the
 # header note.
-#Tue Jul 23 19:22:28 PDT 2019
+#Wed Jan 27 19:24:35 ART 2021
 ndk.dir=/Users/mikechoch/Library/Android/sdk/ndk-bundle
-sdk.dir=/Users/mikechoch/Library/Android/sdk
+sdk.dir=/Users/gonzalonarbaiz/Library/Android/sdk
--- a/Examples/OneSignalDemo/local.properties
+++ b/Examples/OneSignalDemo/local.properties
@@ -4,6 +4,6 @@
 # Location of the SDK. This is only used by Gradle.
 # For customization when using a Version Control System, please read the
 # header note.
-#Wed Jan 27 19:24:35 ART 2021
+#Tue Jul 23 19:22:28 PDT 2019
 ndk.dir=/Users/mikechoch/Library/Android/sdk/ndk-bundle
-sdk.dir=/Users/gonzalonarbaiz/Library/Android/sdk
+sdk.dir=/Users/mikechoch/Library/Android/sdk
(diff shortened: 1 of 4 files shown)


Examples/OneSignalDemo/local.properties
```diff
@@ -4,6 +4,6 @@
 # Location of the SDK. This is only used by Gradle.
 # For customization when using a Version Control System, please read the
 # header note.
-#Tue Jul 23 19:22:28 PDT 2019
+#Wed Jan 27 19:24:35 ART 2021
 ndk.dir=/Users/mikechoch/Library/Android/sdk/ndk-bundle
-sdk.dir=/Users/mikechoch/Library/Android/sdk
+sdk.dir=/Users/gonzalonarbaiz/Library/Android/sdk
```
Jeasmine: this should be ignored

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2149,13 +2149,15 @@ public void run() {
 
       boolean urlOpened = false;
       boolean defaultOpenActionDisabled = "DISABLE".equals(OSUtils.getManifestMeta(context, "com.onesignal.NotificationOpened.DEFAULT"));
+      boolean launchUrlSuppress = "true".equals(OSUtils.getManifestMeta(context, "com.onesignal.suppressLaunchURLs"));
 
-      if (!defaultOpenActionDisabled)
+      if (!defaultOpenActionDisabled && !launchUrlSuppress)
```
Jeasmine: can we add testing for this?

Jeasmine (commented):
Commit message should explain the reason for the change and what is changed.

OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java
```diff
@@ -1168,6 +1167,19 @@ public void testDisableOpeningLauncherActivityOnNotificationOpen() throws Except
       assertEquals("Test Msg", lastNotificationOpenedBody);
    }
 
+   @Test
+   public void testLaunchUrlSuppressTrue() throws Exception {
+      // Add the 'com.onesignal.suppressLaunchURLs' as 'true' meta-data tag
+      OneSignalShadowPackageManager.addManifestMetaData("com.onesignal.suppressLaunchURLs", "true");
```
Jeasmine, line 1173: can we have the case for false?

gonzalonarbaiz: Yes, I can add one, but the default value is false, wouldn't this value be inferred in the test that opens the URL (testOpeningLaunchUrl())?

Jeasmine: Yeah but still we should check all scenarios. It will be 
`OneSignalShadowPackageManager.addManifestMetaData("com.onesignal.suppressLaunchURLs", "false");`

jkasten2 (changes requested):
I also recommend adding a false test case as noted in:
https://github.com/OneSignal/OneSignal-Android-SDK/pull/1268/files#r567015270

Otherwise PR looks good after this.

Jeasmine (approved):

jkasten2 (approved):

PR #1276, opened 2021-02-03
Fix notification open not launching main activity by default on app cold start

jkasten2:
### Fix
This is fixes a bug where the app would not be open when the app was completely closed when tapping on a notification.

* The app process would still be started and the NotificationOpenHandler would still fire, just the default auto launch was not happening in this case.
### Clean up
Clean up to internals of the private `startOrResumeApp` method.

### Automated 
Normally a failing test should be added before a fix is done but since the bugged code path was completely removed it isn't needed. There is an existing test already covering the launching the main activity so all lines of code are covered in this modification.

#### App w/ splash and single Activity (the example app in this repo)
1. In focus notifi open, no flicker, back button works
1. Notif open from cold start, back button works.
1. Notif open after backgrounding app, app resumed correctly, back button works

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1276)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -2174,12 +2174,7 @@ public class OneSignal {
       logger.debug("startOrResumeApp from context: " + inContext + " isRoot: " + inContext.isTaskRoot() + " with launchIntent: " + launchIntent);
       // Make sure we have a launcher intent.
       if (launchIntent != null) {
-         if (inContext.isTaskRoot()) {
-            inContext.startActivity(launchIntent);
-         } else {
-            launchIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-            inContext.startActivity(launchIntent);
-         }
+         inContext.startActivity(launchIntent);
          return true;
       }
 
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -2178,10 +2178,11 @@ public class OneSignal {
             inContext.startActivity(launchIntent);
          } else {
             launchIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -2169,20 +2169,16 @@ static void applicationOpenedByNotification(@Nullable final String notificationI
...
-         }
-         return true;
-      }
-      return false;
+   static boolean startOrResumeApp(Activity activity) {
+      Intent launchIntent = activity.getPackageManager().getLaunchIntentForPackage(activity.getPackageName());
+      logger.debug("startOrResumeApp from context: " + activity + " isRoot: " + activity.isTaskRoot() + " with launchIntent: " + launchIntent);
+
+      // Not all apps have a launcher intent, such one that only provides a homescreen widget
+      if (launchIntent == null)
```
Jeasmine, line 2177: How does this prevent the application from being launched from the launching activity when the application is in the background, instead of resuming the activity on top of the stack?
Doesn't this remove the behavior of emulating app icon click when the app is in the background when clicking a notification?

jkasten2: This bug was we were missing a `startActivity` in the else of the `isTaskRoot` check. This was fixed in commit 0326d93abb354aa3569e945ba6365aeab9dcd9d0 of this PR.

I noticed `getLaunchIntentForPackage` defaults to `FLAG_ACTIVITY_NEW_TASK` so the whole `isTaskRoot` check and setting of it isn't needed. Notes to this and ASOP source linked of this in commit b27ac140eb5e74bfe55468ea4de9426b7bff2a3a

I didn't notice any other behavior differences before and after this PR. See the manual test scenarios above for what I tested.

This PR is scoped to fixing this one specific cold start bug. However I did notice a resume issue which I will be making another PR based on this one to fix that as well. We can follow up there on that.

jkasten2:
Test failing due to unrelated flaky tests.

Jeasmine (approved):

jkasten2:
Ran the following to clean up fixup commits
```
git rebase -i --autosquash master
git push -f
```

PR #1277, opened 2021-02-04
Fix notification open not always resuming app

jkasten2:
### Fix
This fix prevents notification opens affecting the Activity backstack.
Fixes issue where if the app's main Activity was not the last Activity opened tapping on a notification would create another instance of it. Basically this behavior fix aligns with everything an Android Launcher would do.

### Solutions
Solution D below was used but all options that were attempted / considered.

#### A. Android API
Tired searching for a built in Android API that directly resumes the app. This doesn't exist and wasn't expecting it to but looked for completeness.

#### B. startActivity Intent.FLAG_ACTIVITY_* Flags
I read though the description and tried a few different [`Intent.FLAG_ACTIVITY_*`](https://developer.android.com/reference/android/content/Intent) flags to our `startActivity` in our `startOrResumeApp`  to get the app to resume without affecting the backstack. However `FLAG_ACTIVITY_SINGLE_TOP`, `FLAG_ACTIVITY_BROUGHT_TO_FRONT`, and some others I tried did not have the resume affected in all cases or not at all.

#### C. Show non-visible Activity
As a "workaround" it is possible to show an non-visible Activity and call `finish()` right away to bring the app forward to then to put the app back to the last Activity. This would require adding another `Activity` and it would create extra `onResume` and `onPause` events would be mostly acceptable, but could results in unintended side effects.
This something I confirmed as a valid option, but was also [discovered by others](https://stackoverflow.com/questions/11470520/android-application-current-activity-bring-to-front-issue/11470867#11470867).

#### D. Omit package from Intent passed to startActivity
Using `Intent.setPackage(null)` seems to have a side effect of always resuming the app instead of effecting the backstace. I wasn't able to find any official Google documentation on this effect but it does seems to be what Android Launchers depend on. I confirmed this by checking the source of  an open source launcher, `OpenLauncherTeam/openlauncher` and confirmed package would be `null` on their `Intent`.
This solution was found here:
https://stackoverflow.com/a/42002705/1244574

I also confirmed that [` getLaunchIntentForPackage` in ASOP generates an `Intent` that uses `Intent.setClassName`](https://cs.android.com/android/platform/superproject/+/android-11.0.0_r3:frameworks/base/core/java/android/app/ApplicationPackageManager.java;l=252-253) that includes both the package name & `Activity` name so there is no risk in a different app being started if it happened to have the same fully namespaced `Activity`.

Based on testing on Android 11 (API 30) (both an emu and Samsung device) using `setPackage(null)` isn't required, the resume behavior works with or without it. I wasn't able to find any documentation on this behavior change so it is unknown if it was intentional.

### Automated 
Robolectric does not have an API to get the backstack write an integration test for this. The closest thing I could find is `getNextStartedActivity` which would only tell me the call order but none of the behavior changes.

#### App w/ splash and single Activity (the example app in this repo)
1. In focus notifi open, no flicker, back button works
1. Notif open from cold start, back button works.
1. Notif open after backgrounding app, app resumed correctly, back button works

#### App w/ two Activities, Main Activity A starts B (the example app in this repo but `makeRestartActivityTask` in example removed)
1. In focus notifi open, no flicker, back button works
1. Notif open after backgrounding app, app resumed Activity B correctly, back button works
1. Notif open from cold start, kicked out of ram, app resumes to Activity B, back button works
1. Notif open from cold start, app swiped away, app opens Activity A, back button works

Tested on Emulators - Android 11 & 10.
Tested on OnePlus 6T - Android 10
Tested on LG G7 - Android 9

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1276)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -2169,7 +2169,14 @@ public class OneSignal {
       sessionManager.onDirectInfluenceFromNotificationOpen(appEntryState, notificationId);
    }
 
-   static boolean startOrResumeApp(Activity activity) {
+   // This opens the app in the same way an Android homescreen launcher does.
+   // This means we expect the following behavior:
+   //    1. Starts the Activity defined in the app's AndroidManifest.xml as "android.intent.action.MAIN"
+   //    2. If the app is already running, instead the last activity will be resumed
+   //    3. If the app is not running (due to being push out of memory), the last activity will be resumed
+   //    4. If the app is no longer in the recent apps list, it is not resumed, same as #1 above.
+   //        - App is removed from the recent app's list if it is swiped away or "clear all" is pressed.
+   static boolean startOrResumeApp(@NonNull Activity activity) {
       Intent launchIntent = activity.getPackageManager().getLaunchIntentForPackage(activity.getPackageName());
       logger.debug("startOrResumeApp from context: " + activity + " isRoot: " + activity.isTaskRoot() + " with launchIntent: " + launchIntent);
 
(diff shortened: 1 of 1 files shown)


Jeasmine (approved):

jkasten2:
Ran the following after https://github.com/OneSignal/OneSignal-Android-SDK/pull/1276#issuecomment-773855149 was done to clean up fixups
```
git rebase --onto fix/notif_open_cold_start_launch_activity fix/notif_open_resume_app~1 fix/notif_open_resume_app
git push -f
```

Issue #1284, opened 2021-03-03
ANR  at com.onesignal.OneSignalDbHelper.query (OneSignalDbHelper.java:187)


kovrizhin:
Hello Team,

Thanks for providing a good solution and supporting it.

**Description:**
I had a look into ANRs in my application report and found some ANR from OneSignal; this ANR happens quite frequently.

The root cause is:
You are trying to take the lock when you query the data to prevent duplicate of notification or something like this, and you do it in the main thread:

```
   @Override
   public Cursor query(@NonNull String table, @Nullable String[] columns, @Nullable String selection,
                       String[] selectionArgs, @Nullable String groupBy, @Nullable String having,
                       @Nullable String orderBy) {
      synchronized (LOCK) {
         return getSQLiteDatabaseWithRetries().query(table, columns, selection, selectionArgs, groupBy, having, orderBy);
      }
   }
```

But at the same time you are clean cache in a new thread for the same lock, and this thread starts before previous query:

```
   @Override
   public void delete(@NonNull String table, @Nullable String whereClause, @Nullable String[] whereArgs) {
      synchronized (LOCK) {
         SQLiteDatabase writableDb = getSQLiteDatabaseWithRetries();
         try {
            writableDb.beginTransaction();
            writableDb.delete(table, whereClause, whereArgs);
            writableDb.setTransactionSuccessful();
         } catch (SQLiteException e) {
            logger.error("Error deleting on table: " + table + " with whereClause: " + whereClause + " and whereArgs: " + whereArgs, e);
         } catch (IllegalStateException e) {
            logger.error("Error under delete transaction under table: " + table + " with whereClause: " + whereClause + " and whereArgs: " + whereArgs, e);
         } finally {
            if (writableDb != null) {
               try {
                  writableDb.endTransaction(); // May throw if transaction was never opened or DB is full.
               } catch (IllegalStateException e) {
                  logger.error("Error closing transaction! ", e);
               } catch (SQLiteException e) {
                  logger.error("Error closing transaction! ", e);
               }
            }
         }
      }
   }
```

In this case main thread are locked by this new thread which cleaning up the cache and android throw an ANR.

**Environment**
gradle android sdk
implementation 'com.onesignal:OneSignal:3.15.4'

**Steps to Reproduce Issue:**
I don't know how to reproduce it, but it fails and I see it in the console

**Anything else:**
"main" prio=5 tid=1 Blocked
[...]

PR #1290, opened 2021-03-12
Improve thread DB access handling

Jeasmine:
### The Problem
isDuplicateNotification methods sometimes ended called on the main thread, this can en on an ANR is Notification table is being blocked by other thread.

### Solution
Move all database access from OneSignal.java to the new class OSNotificationDataController. In the future, we will want all notification table access to be on the same class to have a single responsible class, also to have few threads handling notification table access.

Move outcome table access to OSOutcomeEventsCache
* OneSignalCacheCleaner shouldn't have access to the outcome table when we have a class responsible for managing the outcome database table

Move IAM cache cleaning to OSInAppMessageController
* IAM database access and all regarding IAM should live in OSInAppMessageController

Rename OneSignalCacheCleaner to OSNotificationDataController
* OSNotificationDataController will be the new class responsible for handling notification table access

Move OneSignal db access to OSNotificationDataController 
* Move clearOneSignalNotifications from OneSignal.java to OSNotificationDataController
* Move removeGroupedNotifications from OneSignal.java to OSNotificationDataController
* Move removeNotification from OneSignal.java to OSNotificationDataController
* Move isDuplicateNotification from OneSignal.java to OSNotificationDataController
* isDuplicateNotification no longer returns a boolean, all methods using isDuplicateNotification wait for callback result

Fix https://github.com/OneSignal/OneSignal-Android-SDK/issues/1284

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1290)


jkasten2 (approved):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1290)* status: all files reviewed, 2 unresolved discussions (waiting on @Jeasmine)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/ADMMessageHandlerJob.kt, line 17 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1290#-MVyqnUd0EK3J4zLe8yQ:-MW0xzEA9RsjYtGOt1CS:bh6zhck) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/b5b1792605f0cae01abc1f9d4878c45c82a480a5/OneSignalSDK/onesignal/src/main/java/com/onesignal/ADMMessageHandlerJob.kt#L17)):*

Not sure what the best practices in general are for Kotlin but I think it would be cleaner like this.
```Kotlin
if (processedResult?.processed() == true) return
```
</blockquote></details>

This is really more of a nit, not a blocker.

PR #1342, opened 2021-05-23
Fix Tests Stalling

jkasten2:
## One Line Summary
Fixes tests stalling and never finishing. Tests still fail due to some flaky tests but it always completes now.

### Shutdown ExecutorService instances after each test
* Stalling would happen on stopAllOSThreads() where it would
continue to try to call interrupt() on a OS_PENDING_EXECUTOR_ thread
over and over. The thread can't be interrupted as an Android
ExecutorService will catch it.
* Added a shutdownNow() call to OSTaskController that tests can run to
correctly shutdown the pending tasks so it's thread can be stopped.

### Log of rerunning tests
* Run 1 - no stall
* Run 2 - stalled https://travis-ci.com/github/OneSignal/OneSignal-Android-SDK/builds/226617186
#### Added after test cleanup to HMSDataMessageReceivedIntegrationTestsRunner
* Run 1 - no stall https://travis-ci.com/github/OneSignal/OneSignal-Android-SDK/builds/226620653
* Run 2 - no stall
* Run 3 - no stall
* Run 4 - no stall
* Run 5 - no stall
* Run 6 - no stall
* Run 7 - no stall
* Run 8 - no stall

## Note

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1342)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSTaskController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSTaskController.java
@@ -105,6 +105,12 @@ class OSTaskController {
         return taskQueueWaitingForInit;
     }
 
+    void shutdownNow() {
+        if (pendingTaskExecutor != null) {
+            pendingTaskExecutor.shutdownNow();
+        }
+    }
+
     private static class PendingTaskRunnable implements Runnable {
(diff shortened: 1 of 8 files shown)


Jeasmine (approved):
Nice!

PR #1345, opened 2021-05-24
Fix flaky tests 2021-05-22

jkasten2:
### One Line Summary
Fix both some individual flaky tests as well as some general test carry over issues.

### Details
* Fix some ad_id field asserts in some testss that didn't get removed.
* Fix focusTimeController carry over from test to test issue.
* Fix test util method `nextSuccessfulRegistrationResponse` not correctly filtering to just player requests.
* Fix InAppMessageIntegrationTests afterEach test not force stopping some threads which sometimes created test carry over.
* A number of flaky tests in InAppMessageIntegrationTests fixed on the individual level, see commit-by-commit for each one.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1345)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -404,7 +404,7 @@ public class OneSignal {
    }
 
    private static OSLogger logger = new OSLogWrapper();
-   private static FocusTimeController focusTimeController = new FocusTimeController(new OSFocusTimeProcessorFactory(), logger);
+   private static FocusTimeController focusTimeController;
    private static OSSessionManager.SessionListener sessionListener = new OSSessionManager.SessionListener() {
          @Override
          public void onSessionEnding(@NonNull List<OSInfluence> lastInfluences) {
@@ -412,7 +412,7 @@ public class OneSignal {
                OneSignal.Log(LOG_LEVEL.WARN, "OneSignal onSessionEnding called before init!");
             if (outcomeEventsController != null)
                outcomeEventsController.cleanOutcomes();
-            focusTimeController.onSessionEnded(lastInfluences);
+            getFocusTimeController().onSessionEnded(lastInfluences);
          }
       };
 
@@ -891,7 +891,7 @@ public class OneSignal {
             activityLifecycleHandler.setNextResumeIsFirstActivity(true);
          }
          OSNotificationRestoreWorkManager.beginEnqueueingWork(context, false);
-         focusTimeController.appForegrounded();
+         getFocusTimeController().appForegrounded();
       } else if (activityLifecycleHandler != null) {
          activityLifecycleHandler.setNextResumeIsFirstActivity(true);
       }
@@ -1289,7 +1289,7 @@ public class OneSignal {
       if (trackAmazonPurchase != null)
          trackAmazonPurchase.checkListener();
 
-      focusTimeController.appBackgrounded();
+      getFocusTimeController().appBackgrounded();
 
       scheduleSyncService();
    }
@@ -1337,7 +1337,7 @@ public class OneSignal {
       if (shouldLogUserPrivacyConsentErrorMessageForMethodName("onAppFocus"))
          return;
 
-      focusTimeController.appForegrounded();
+      getFocusTimeController().appForegrounded();
 
       doSessionInit();
 
@@ -3172,6 +3172,10 @@ public class OneSignal {
    }
 
    static FocusTimeController getFocusTimeController() {
+      if (focusTimeController == null) {
+         focusTimeController = new FocusTimeController(new OSFocusTimeProcessorFactory(), logger);
+      }
+
       return focusTimeController;
    }
    /*
--- a/OneSignalSDK/unittest/src/test/java/com/onesignal/OneSignalPackagePrivateHelper.java
+++ b/OneSignalSDK/unittest/src/test/java/com/onesignal/OneSignalPackagePrivateHelper.java
@@ -30,6 +30,8 @@ public class OneSignalPackagePrivateHelper {
 
    public static final long MIN_ON_SESSION_TIME_MILLIS = com.onesignal.OneSignal.MIN_ON_SESSION_TIME_MILLIS;
 
+   private static final String LOGCAT_TAG = "OS_PACKAGE_HELPER";
+
    private static abstract class RunnableArg<T> {
       abstract void run(T object) throws Exception;
    }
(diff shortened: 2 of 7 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
```diff
@@ -404,15 +404,15 @@ public static String getSdkVersionRaw() {
    }
 
    private static OSLogger logger = new OSLogWrapper();
-   private static FocusTimeController focusTimeController = new FocusTimeController(new OSFocusTimeProcessorFactory(), logger);
```
Jeasmine, line 407: Why changing focusTimeController to lazy init avoid the carry-over 🤔  ? FocusTimeController constructor isn't triggering a functionality and is reset under StaticResetHelper. Am I missing something?

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessageIntegrationTests.java
```diff
@@ -420,7 +420,12 @@ public void testAfterLastInAppTimeIsDisplayedOncePerSession() throws Exception {
                 .pollInterval(new Duration(10, TimeUnit.MILLISECONDS))
                 .untilAsserted(() -> {
                     assertEquals(1, OneSignalPackagePrivateHelper.getInAppMessageDisplayQueue().size());
-                    assertEquals(message1.messageId, OneSignalPackagePrivateHelper.getShowingInAppMessageId());
+                    try {
+                        assertEquals(message1.messageId, OneSignalPackagePrivateHelper.getShowingInAppMessageId());
```
Jeasmine, line 428: Should we assert under the second Awaitility.await() too?

OneSignalSDK/unittest/src/test/java/com/onesignal/OneSignalPackagePrivateHelper.java
```diff
@@ -513,8 +513,12 @@ public OSTestInAppMessageAction(JSONObject json) throws JSONException {
 
    public static void dismissCurrentMessage() {
       com.onesignal.OSInAppMessage message = OneSignal.getInAppMessageController().getCurrentDisplayedInAppMessage();
-      if (message != null)
+      if (message != null) {
          OneSignal.getInAppMessageController().messageWasDismissed(message);
+      }
```
Jeasmine, line 518: since re-formatting code will put the else after the brace, can we keep the same code styling? 

  ```
if (message != null) {         
       OneSignal.getInAppMessageController().messageWasDismissed(message);    
  } else {
     ....
  }
```

File: OneSignalSDK/unittest/src/test/java/com/onesignal/OneSignalPackagePrivateHelper.java [hunk 4]
```diff
@@ -513,8 +513,12 @@ public OSTestInAppMessageAction(JSONObject json) throws JSONException {
...
+      else {
+         System.out.println("No currently displaying IAM to dismiss!");
```
Jeasmine, line 520: what about using use android Log.e ?

Jeasmine (commented):
Nice fixes! 💯

emawby (changes requested):
> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1345)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @jkasten2 and @tanaynigam)

___
*[OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessageIntegrationTests.java, line 153 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1345#-MaUAnvH91826e8R-3_R:-MaUALc3BI9VDyf5I3Bq:b-8zodhs) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/050a8efccd2b5aaa95d342061b0c9ddbb8a36083/OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessageIntegrationTests.java#L153)):*
> ```Java
>         ShadowDynamicTimer.hasScheduledTimer = false;
>         OneSignal.setInAppMessageClickHandler(null);
>         InAppMessagingHelpers.clearTestState();
> ```

Does afterTestCleanup() remove the background threads created in clearTestState or are we just relying on more time being spent between clearTestState() and the start of the next test?

OneSignalSDK/unittest/src/test/java/com/onesignal/OneSignalPackagePrivateHelper.java
jkasten2: Fixed

OneSignalSDK/unittest/src/test/java/com/onesignal/OneSignalPackagePrivateHelper.java
jkasten2: Fixed

OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessageIntegrationTests.java
jkasten2: Oddly enough I never the Awaitility below for the 2nd IAM every fail. However you are right in the fact I think it could so I'll add it there too.
Fixed in commit: 0749a242133fb2dfb0bb5b053fa781050c57a07c

jkasten2 (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1345)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @emawby and @tanaynigam)

___
*[OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessageIntegrationTests.java, line 153 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1345#-MaUAnvH91826e8R-3_R:-MaW61kFAraPA12BvzDS:bpaaypw) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/050a8efccd2b5aaa95d342061b0c9ddbb8a36083/OneSignalSDK/unittest/src/test/java/com/test/onesignal/InAppMessageIntegrationTests.java#L153)):*
<details><summary><i>Previously, emawby (Elliot Mawby) wrote…</i></summary><blockquote>

Does afterTestCleanup() remove the background threads created in clearTestState or are we just relying on more time being spent between clearTestState() and the start of the next test?
</blockquote></details>

`InAppMessagingHelpers.clearTestState();` does run some background work so it is import for `TestHelpers.afterTestCleanup();` to run last to prevent any possible test carry over.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
Jeasmine: Understood, My only concern with the lazy getter is that this goes in the contrary direction with the dependency injection pattern we want to follow.

emawby (approved):
> Reviewed 1 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1345)* status: all files reviewed, 1 unresolved discussion (waiting on @emawby and @tanaynigam)

Jeasmine (approved):

PR #1356, opened 2021-06-07
Add Receive Receipt delay

Jeasmine:
* Add random delay from 0 to 25 seconds
* Add OSDelayTaskController for future delays
* Add MockDelayTaskController for test approaches
* Refactor Notification Work Manager to extends ListenableWorker. This adds future completion. Complete future after sending receive receipts
* Refactor Restoration Work Manager to not init another work manager, since Notification Work Manager nows expects a future when Restoration work is done, it will finish notification work manager that will end on a FutureGarbageCollectedException, avoid this exception by running everything on the same work thread

Note:

HMS and ADM handling always ends calling startNotificationProcessing which ends using OSNotificationWorkManager worker thread

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)

### DIFF
--- a/OneSignalSDK/build.gradle
+++ b/OneSignalSDK/build.gradle
@@ -8,6 +8,7 @@ buildscript {
                 targetSdkVersion: 28
         ]
         androidGradlePluginVersion = '3.6.2'
+        androidConcurrentFutures = '1.1.0'
         googleServicesGradlePluginVersion = '4.3.2'
         huaweiAgconnectVersion = '1.2.1.301'
         huaweiHMSPushVersion = '4.0.2.300'
--- a/OneSignalSDK/onesignal/build.gradle
+++ b/OneSignalSDK/onesignal/build.gradle
@@ -49,6 +49,8 @@ dependencies {
     // Required for GoogleApiAvailability
     implementation 'com.google.android.gms:play-services-base:[17.0.0, 17.6.99]'
 
+    implementation("androidx.concurrent:concurrent-futures:$androidConcurrentFutures")
+
     // firebase-messaging:18.0.0 is the last version before going to AndroidX
     // firebase-messaging:19.0.0 is the first version using AndroidX
     api 'com.google.firebase:firebase-messaging:[19.0.0, 22.0.99]'
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
@@ -347,23 +347,23 @@ class NotificationBundleProcessor {
             return;
         String collapse_id = notificationJob.getJsonPayload().optString("collapse_key");
 
-      OneSignalDbHelper dbHelper = OneSignalDbHelper.getInstance(notificationJob.getContext());
-      Cursor cursor = dbHelper.query(
-              NotificationTable.TABLE_NAME,
-              new String[]{NotificationTable.COLUMN_NAME_ANDROID_NOTIFICATION_ID}, // retColumn
-              NotificationTable.COLUMN_NAME_COLLAPSE_ID + " = ? AND " +
-                      NotificationTable.COLUMN_NAME_DISMISSED + " = 0 AND " +
-                      NotificationTable.COLUMN_NAME_OPENED + " = 0 ",
-              new String[]{collapse_id},
-              null, null, null);
+        OneSignalDbHelper dbHelper = OneSignalDbHelper.getInstance(notificationJob.getContext());
+        Cursor cursor = dbHelper.query(
+                NotificationTable.TABLE_NAME,
+                new String[]{NotificationTable.COLUMN_NAME_ANDROID_NOTIFICATION_ID}, // retColumn
+                NotificationTable.COLUMN_NAME_COLLAPSE_ID + " = ? AND " +
+                        NotificationTable.COLUMN_NAME_DISMISSED + " = 0 AND " +
+                        NotificationTable.COLUMN_NAME_OPENED + " = 0 ",
+                new String[]{collapse_id},
+                null, null, null);
 
-      if (cursor.moveToFirst()) {
-         int androidNotificationId = cursor.getInt(cursor.getColumnIndex(NotificationTable.COLUMN_NAME_ANDROID_NOTIFICATION_ID));
-         notificationJob.setAndroidIdWithoutOverriding(androidNotificationId);
-      }
+        if (cursor.moveToFirst()) {
+            int androidNotificationId = cursor.getInt(cursor.getColumnIndex(NotificationTable.COLUMN_NAME_ANDROID_NOTIFICATION_ID));
+            notificationJob.setAndroidIdWithoutOverriding(androidNotificationId);
+        }
 
-      cursor.close();
-   }
+        cursor.close();
+    }
 
     /**
      * Process bundle passed from FCM / HMS / ADM broadcast receiver
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
@@ -165,6 +165,11 @@ class NotificationBundleProcessor {
             String osNotificationId = OSNotificationFormatHelper.getOSNotificationIdFromJson(notificationController.getNotificationJob().getJsonPayload());
             OSNotificationWorkManager.removeNotificationIdProcessed(osNotificationId);
             OneSignal.handleNotificationReceived(notificationJob);
+        } else {
+            CallbackToFutureAdapter.Completer<ListenableWorker.Result>  callbackCompleter = notificationJob.getCallbackCompleter();
+            OneSignal.Log(OneSignal.LOG_LEVEL.DEBUG, "Process notification restored or IAM with callback completer: " + callbackCompleter);
+            if (callbackCompleter != null)
+                callbackCompleter.set(ListenableWorker.Result.success());
         }
 
         return androidNotificationId;
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
@@ -36,6 +36,8 @@ import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.WorkerThread;
+import androidx.concurrent.futures.CallbackToFutureAdapter;
+import androidx.work.ListenableWorker;
 
 import com.onesignal.OneSignalDbContract.NotificationTable;
 
@@ -97,9 +99,10 @@ class NotificationBundleProcessor {
                             osNotificationId,
                             finalAndroidNotificationId,
                             jsonStrPayload,
-                            isRestoring,
                             shownTimeStamp,
-                            false);
+                            isRestoring,
+                            false,
+                            true);
 
                     // Delay to prevent CPU spikes.
                     // Normally more than one notification is restored at a time.
@@ -131,7 +134,7 @@ class NotificationBundleProcessor {
     }
 
     @WorkerThread
-    static int processJobForDisplay(OSNotificationController notificationController, boolean opened, boolean fromBackgroundLogic) {
+    private static int processJobForDisplay(OSNotificationController notificationController, boolean opened, boolean fromBackgroundLogic) {
         OneSignal.Log(OneSignal.LOG_LEVEL.DEBUG, "Starting processJobForDisplay opened: " + opened + " fromBackgroundLogic: " + fromBackgroundLogic);
         OSNotificationGenerationJob notificationJob = notificationController.getNotificationJob();
 
@@ -176,17 +179,22 @@ class NotificationBundleProcessor {
      */
     static void processNotification(OSNotificationGenerationJob notificationJob, boolean opened, boolean notificationDisplayed) {
         saveNotification(notificationJob, opened);
+        CallbackToFutureAdapter.Completer<ListenableWorker.Result>  callbackCompleter = notificationJob.getCallbackCompleter();
 
         if (!notificationDisplayed) {
             // Notification channel disable or not displayed
             // save notification as dismissed to avoid user re-enabling channel and notification being displayed due to restore
             markNotificationAsDismissed(notificationJob);
+
+            OneSignal.Log(OneSignal.LOG_LEVEL.DEBUG, "Process notification not displayed with callback completer: " + callbackCompleter);
+            if (callbackCompleter != null)
+                callbackCompleter.set(ListenableWorker.Result.success());
             return;
         }
 
         // Logic for when the notification is displayed
         String notificationId = notificationJob.getApiNotificationId();
-        OSReceiveReceiptController.getInstance().sendReceiveReceipt(notificationId);
+        OSReceiveReceiptController.getInstance().sendReceiveReceipt(callbackCompleter, notificationId);
         OneSignal.getSessionManager().onNotificationReceived(notificationId);
     }
 
@@ -425,9 +433,10 @@ class NotificationBundleProcessor {
                         osNotificationId,
                         androidNotificationId,
                         jsonPayload.toString(),
-                        isRestoring,
                         timestamp,
-                        isHighPriority);
+                        isRestoring,
+                        isHighPriority,
+                        true);
 
                 bundleResult.setWorkManagerProcessing(true);
                 notificationProcessingCallback.onResult(true);
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationSummaryManager.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationSummaryManager.java
@@ -143,7 +143,7 @@ class NotificationSummaryManager {
              null
          );
 
-         OSNotificationRestoreWorkManager.showNotificationsFromCursor(context, cursor, 0);
+         OSNotificationRestoreWorkManager.showNotificationsFromCursor(context, cursor, 0, true);
       } catch (Throwable t) {
          OneSignal.Log(OneSignal.LOG_LEVEL.ERROR, "Error restoring notification records! ", t);
       } finally {
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt
@@ -1,0 +1,40 @@
+package com.onesignal
+
+import java.util.concurrent.ScheduledThreadPoolExecutor
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.TimeUnit
+import kotlin.math.roundToInt
+
+open class OSDelayTaskController(private val logger: OSLogger) {
+    private val maxDelay = 25
+    private val minDelay = 0
+
+    private var scheduledThreadPoolExecutor: ScheduledThreadPoolExecutor
+
+    init {
+        scheduledThreadPoolExecutor = ScheduledThreadPoolExecutor(1, JobThreadFactory())
+    }
+
+    protected open fun getRandomNumber(): Int {
+        return (Math.random() * (maxDelay - minDelay + 1) + minDelay).roundToInt()
+    }
+
+    open fun delayTaskByRandom(runnable: Runnable) {
+        val randomNum = getRandomNumber()
+
+        logger.debug("OSDelayTaskController delaying task $randomNum second from thread: ${Thread.currentThread()}")
+        scheduledThreadPoolExecutor.schedule(runnable, randomNum.toLong(), TimeUnit.SECONDS)
+    }
+
+    fun shutdownNow() {
+        scheduledThreadPoolExecutor.shutdownNow()
+    }
+
+    private class JobThreadFactory : ThreadFactory {
+        private val delayThreadName = "ONE_SIGNAL_DELAY"
+
+        override fun newThread(runnable: Runnable): Thread {
+            return Thread(runnable, delayThreadName)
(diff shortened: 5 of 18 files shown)


Dash63:
nice meet

jkasten2 (changes requested):
Most comments i left on specific lines however in addition the test`shouldSendReceivedReceiptWhenEnabled` is [failing on TravisCI](https://travis-ci.com/github/OneSignal/OneSignal-Android-SDK/builds/228260020#L3862-L3878) and needs to be updated for the changes made in this PR.

> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-:-MbevXMk45usR0EW6I21:bpahi0v)* status: all files reviewed, 3 unresolved discussions (waiting on @Jeasmine)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt, line 9 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbeuLKb8A6A5PLGa2bk:-MbeuLKcDMy0ZThoM4xl:b-v8somc) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt#L9)):*
I think we will probably need to use the WorkManager for this implementation. This could be started from an `onRecieved` of a `BroadcastReceiver`. 

See this Android documentation on `BroadcastReceiver` behavior that applies here:
https://developer.android.com/guide/components/broadcasts#security-and-best-practices

> Because a receiver's onReceive(Context, Intent) method runs on the main thread, it should execute and return quickly. If you need to perform long running work, be careful about spawning threads or starting background services because the system can kill the entire process after onReceive() returns. For more information, see Effect on process state To perform long running work, we recommend:
> * Calling goAsync() in your receiver's onReceive() method and passing the BroadcastReceiver.PendingResult to a background thread. This keeps the broadcast active after returning from onReceive(). However, even with this approach the system expects you to finish with the broadcast very quickly (under 10 seconds). It does allow you to move work to another thread to avoid glitching the main thread.
Scheduling a job with the JobScheduler. For more information, see Intelligent Job Scheduling.


Due note that a OneSignal notification is only fully displayed and process from a `onRecieved` only if there isn't a NSE setup and no remote images to pull down. In those cases we already use the WorkManager I believe but we need to ensure we only mark the work done once we send the confirmed delivery network call.
___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java, line 3 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesiVcD-6ciOHASLNb:-MbesiVcD-6ciOHASLNc:b-jwgfyr) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java#L3)):*
> ```Java
> package com.onesignal;
> 
> import org.jetbrains.annotations.NotNull;
> ```

Use AndroidX's `NonNull` instead.
___
*[OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java, line 1353 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesXDd42zrq3jYvTZf:-MbesXDd42zrq3jYvTZg:b-c2js8t) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java#L1353)):*
> ```Java
>    @Test
>    @Config(shadows = { ShadowGenerateNotification.class })
>    public void testNotificationReceivedNoSendReceivedRequest_Delay() throws Exception {
> ```

This new test is failing on TravisCI
https://travis-ci.com/github/OneSignal/OneSignal-Android-SDK/builds/228260020#L2494-L2519

Jeasmine (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @Jeasmine and @jkasten2)

___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java, line 3 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesiVcD-6ciOHASLNb:-MbgRBOn2KBysZzOykpo:bqr9fjr) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java#L3)):*

Use AndroidX's `NonNull` instead.
</blockquote></details>

This object is outside the main and test package it doesn't have android x access, Should we add android x dependency to mock testing files?
___
*[OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java, line 1353 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesXDd42zrq3jYvTZf:-MbgRlgR39qh6zLnQp9o:b-cv0iri) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/test/onesignal/MainOneSignalClassRunner.java#L1353)):*

This new test is failing on TravisCI
https://travis-ci.com/github/OneSignal/OneSignal-Android-SDK/builds/228260020#L2494-L2519
</blockquote></details>

Yeah I was still working on some problem with thee main looper, this is fixed, but will be working on the work manager change

Jeasmine (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @jkasten2)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt, line 9 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbeuLKb8A6A5PLGa2bk:-MbgRsK609CMRv7Rj9Ey:bkm74m1) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt#L9)):*

I think we will probably need to use the WorkManager for this implementation. This could be started from an `onRecieved` of a `BroadcastReceiver`. 

See this Android documentation on `BroadcastReceiver` behavior that applies here:
https://developer.android.com/guide/components/broadcasts#security-and-best-practices

> Because a receiver's onReceive(Context, Intent) method runs on the main thread, it should execute and return quickly. If you need to perform long running work, be careful about spawning threads or starting background services because the system can kill the entire process after onReceive() returns. For more information, see Effect on process state To perform long running work, we recommend:
> * Calling goAsync() in your receiver's onReceive() method and passing the BroadcastReceiver.PendingResult to a background thread. This keeps the broadcast active after returning from onReceive(). However, even with this approach the system expects you to finish with the broadcast very quickly (under 10 seconds). It does allow you to move work to another thread to avoid glitching the main thread.
Scheduling a job with the JobScheduler. For more information, see Intelligent Job Scheduling.


Due note that a OneSignal notification is only fully displayed and process from a `onRecieved` only if there isn't a NSE setup and no remote images to pull down. In those cases we already use the WorkManager I believe but we need to ensure we only mark the work done once we send the confirmed delivery network call.
</blockquote></details>

Totally right!

jkasten2 (changes requested):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: all files reviewed, 3 unresolved discussions (waiting on @Jeasmine and @jkasten2)

___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java, line 3 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesiVcD-6ciOHASLNb:-McNeXJ378kynSeOFLKF:ba39qei) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java#L3)):*
<details><summary><i>Previously, Jeasmine (JNahui) wrote…</i></summary><blockquote>

This object is outside the main and test package it doesn't have android x access, Should we add android x dependency to mock testing files?
</blockquote></details>

`androidx.annotation.NonNull` should work here. This file is in the `unittest` folder which uses the `unittest/build.gradle` which has `implementation(project(':onesignal'))` which will includes all it's dependencies.

jkasten2 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: 6 of 18 files reviewed, 7 unresolved discussions (waiting on @Jeasmine and @jkasten2)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt, line 18 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-Mcrs3TLAA76kYc1cvoC:-Mcrs3TLAA76kYc1cvoD:b-k12eu3) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt#L18)):*
> ```Kotlin
>     }
> 
>     protected open fun getRandomNumber(): Int {
> ```

I think `getRandomDelay` as a method name is more fitting here.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java, line 91 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-McrvAH9C9PYZnOS4YLl:-McrvAHAEHAAXPlB181u:bhr82qn) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java#L91)):*
> ```Java
>             OneSignalDbHelper dbHelper,
>             StringBuilder dbQuerySelection,
>             boolean needsWorkerThread) {
> ```

It looks like we only call this method in one palace and `needsWorkerThread` is always `false` here.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java, line 17 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-McryGgd1Ss1aeFwLKkV:-McryGgd1Ss1aeFwLKkW:b-2bbxya) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java#L17)):*
> ```Java
> import com.google.common.util.concurrent.ListenableFuture;
> 
> import org.jetbrains.annotations.NotNull;
> ```

This should be AndroidX's `NonNull` instead.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java, line 89 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-Mcs01wXDWVJjfcN-9RL:-Mcs01wXDWVJjfcN-9RM:b-i0rdsi) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java#L89)):*
> ```Java
>         OneSignal.Log(OneSignal.LOG_LEVEL.DEBUG, "Receive receipt ending with success callback completer: " + callbackCompleter);
>         if (callbackCompleter != null)
>             callbackCompleter.set(ListenableWorker.Result.success());
> ```

I am not sure I understand moving to a `WorkerListener` from a `Worker` gives us here?
This receive receipt network call before we finish and this worker result could be set before a notification is even displayed. In the case where the random delay time is for the received receipt and the notification has an remote image that takes a while to download.

To solve this we need to wait for both events to complete before we count the `Worker` as complete. (In JS land thinking of received receipt and notification display as two Promises and we need to await for both to complete.) 
___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java, line 3 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesiVcD-6ciOHASLNb:-McryRLZDs8piIOLmGWx:b-jwp2t7) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java#L3)):*

`androidx.annotation.NonNull` should work here. This file is in the `unittest` folder which uses the `unittest/build.gradle` which has `implementation(project(':onesignal'))` which will includes all it's dependencies.
</blockquote></details>

In most projects any references you use directly in a project should be added to it, so you don't rely on a sub-dependency that could change. However since this `unittest` project is a superset of the `onesignal` project it really should always contain at least those dependencies.
Anyway, so I am saying if AndroidX's annotations are being used in the main project they should always be accessible to tests.

Jeasmine (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: 6 of 18 files reviewed, 7 unresolved discussions (waiting on @jkasten2)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt, line 18 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-Mcrs3TLAA76kYc1cvoC:-MdIYw7uDC9Ocsp6kUhb:b-896fix) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSDelayTaskController.kt#L18)):*

I think `getRandomDelay` as a method name is more fitting here.
</blockquote></details>

Done.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java, line 91 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-McrvAH9C9PYZnOS4YLl:-MdIYxF69HRuzifRvfxr:b-6eik5) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationRestoreWorkManager.java#L91)):*

It looks like we only call this method in one palace and `needsWorkerThread` is always `false` here.
</blockquote></details>

Yeah, I wanted to make it parametrized for the future, but it can be simplified for now
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java, line 17 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-McryGgd1Ss1aeFwLKkV:-MdIZc8e25GWIZ8oSJy6:b-896fix) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationWorkManager.java#L17)):*

This should be AndroidX's `NonNull` instead.
</blockquote></details>

Done.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java, line 89 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-Mcs01wXDWVJjfcN-9RL:-MdIdI-q5oIQMBPdWry5:b-vtd4t8) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java#L89)):*
> WorkerListener
Worker Listener is added to ensure that the received receipt call is done and avoid being finished by the operating system. I have checked that the show notification is done before receive receipt is called.  The method processJobForDisplay handles all the notification construction and image pulling before calling processNotification.

processNotification also is in charge of saving the notification on the DB before OSReceiveReceiptController is called. So ReceiveReceip is the last thing called under the notifications building. Are the notification images being handle in another thread? 

The only bitmap networking call I see is the following.

```   private static Bitmap getBitmapFromURL(String location) {
      try {
         return BitmapFactory.decodeStream(new URL(location).openConnection().getInputStream());
      } catch (Throwable t) {
         OneSignal.Log(OneSignal.LOG_LEVEL.WARN, "Could not download image!", t);
      }

      return null;
 ```

And this is handle in the same thread. Let me know If I'm missing something
___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java, line 3 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-MbesiVcD-6ciOHASLNb:-MdI_PyR76u3wvMIslLK:b-3heufr) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/65344012fddd6ba32bf935e1241fe4ca12d27b50/OneSignalSDK/unittest/src/test/java/com/onesignal/MockDelayTaskController.java#L3)):*

In most projects any references you use directly in a project should be added to it, so you don't rely on a sub-dependency that could change. However since this `unittest` project is a superset of the `onesignal` project it really should always contain at least those dependencies.
Anyway, so I am saying if AndroidX's annotations are being used in the main project they should always be accessible to tests.
</blockquote></details>

I was confusing NotNull with Nonnull :facepalm: this is done!

jkasten2 (changes requested):
> Reviewed 3 of 12 files at r3.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: 8 of 18 files reviewed, 3 unresolved discussions (waiting on @Jeasmine and @jkasten2)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java, line 89 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-Mcs01wXDWVJjfcN-9RL:-MdYo6WO0z4_un-SeGRo:b-b94tql) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java#L89)):*
<details><summary><i>Previously, Jeasmine (JNahui) wrote…</i></summary><blockquote>

> WorkerListener
Worker Listener is added to ensure that the received receipt call is done and avoid being finished by the operating system. I have checked that the show notification is done before receive receipt is called.  The method processJobForDisplay handles all the notification construction and image pulling before calling processNotification.

processNotification also is in charge of saving the notification on the DB before OSReceiveReceiptController is called. So ReceiveReceip is the last thing called under the notifications building. Are the notification images being handle in another thread? 

The only bitmap networking call I see is the following.

```   private static Bitmap getBitmapFromURL(String location) {
      try {
         return BitmapFactory.decodeStream(new URL(location).openConnection().getInputStream());
      } catch (Throwable t) {
         OneSignal.Log(OneSignal.LOG_LEVEL.WARN, "Could not download image!", t);
      }

      return null;
 ```

And this is handle in the same thread. Let me know If I'm missing something
</blockquote></details>

ah my mistake, thought the notification generation and / or downloading was done on its own thread.

Just to confirm the sequence here:
1. notification is generated (image is downloaded if needed)
2. `scheduledThreadPoolExecutor` is used to schedule the confirmed delivery 0 to 25 seconds in the future
3. Confirmed Delivery network completes and `callbackCompleter.set()` is called

Jeasmine (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: 8 of 18 files reviewed, 3 unresolved discussions (waiting on @jkasten2)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java, line 89 at r3](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356#-Mcs01wXDWVJjfcN-9RL:-MdYwkso4r4tmHBvyKdp:bso0bfe) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/39cd7b0877de5ec717e85b9ccacbe3ba55f17fe3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSReceiveReceiptController.java#L89)):*

ah my mistake, thought the notification generation and / or downloading was done on its own thread.

Just to confirm the sequence here:
1. notification is generated (image is downloaded if needed)
2. `scheduledThreadPoolExecutor` is used to schedule the confirmed delivery 0 to 25 seconds in the future
3. Confirmed Delivery network completes and `callbackCompleter.set()` is called


</blockquote></details>

That sequence is correct. For cases we don't send receive receipt, the callbackCompleter.set() is also called

jkasten2 (approved):
> Reviewed 5 of 12 files at r3, 3 of 5 files at r4, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: all files reviewed, 3 unresolved discussions (waiting on @jkasten2)

jkasten2 (approved):
> Reviewed 6 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1356)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

Issue #1355, opened 2021-06-04
Use OneSignal in existing notification handler service


keeslinp:
**Description:**

I'm doing an evaluation of OneSignal for my company to transition to after mix panel sunsets its messaging platform and I am having a hard time grokking how to set up android.

Android only lets you have one firebase notification handling service and so other services like OneSignal have the ability to call them from in your existing handler.

For example mix panel does this: https://github.com/mixpanel/mixpanel-android/releases/tag/v5.5.1
and Iterable does this: https://support.iterable.com/hc/en-us/articles/115000331943#handling-firebase-push-messages-and-tokens

I'm not able to find any way in the OneSignal SDK to register a push token change or pass in a notification if it comes from OneSignal. It is totally possible that I am missing something, but it seems like OneSignal expects to be the sole handler of push notifications in the client.

jkasten2:
@keeslinp Thanks for the question, I can confirm there are no extra steps that are needed to use your own `FirebaseMessagingService.onMessageReceived` with OneSignal. Without going into to much detail this is because the OneSignal SDK uses a `BroadcastReceiver` (which allows for multiple receivers) internally instead of using FCM's `FirebaseMessagingService`.

I'll go into however what this means for your app, when you send a notification from:
1. **Onesignal** - The OneSignal SDK will display the message. Also the `FirebaseMessagingService.onMessageReceived` fires as a data message type.
   - OneSignal also has a [Notification Service Extension](https://documentation.onesignal.com/docs/service-extensions#android-notification-service-extension) for OneSignal notifications. This works very similar to iOS's Notification Service Extension if you are familiar with it, allowing you to modify the notification before displaying it.  However it is more powerful as it allows you to omit displaying the notification altogether too. 
2. **FCM or another push vendor** - The OneSignal SDK ignores the notification. Other code fires as expected.

nan-li:
This is a good question, closing this as an issue. Added the documentation label to make this info easier to find in the future.

alisherbu:
Aren't there any restrictions on using a custom handler(BroadcastReceiver) by Google? I also want to manage FCM messages within my library. It streamlines the setup process for users(devs) rather than utilizing FirebaseMessagingService.onMessageReceived

PR #1359, opened 2021-06-11
Fix FirebaseMessagingService migration comment

jkasten2:
* Per the comment in this change we don't want to migrate to using
FirebaseMessagingService.onMessageReceived as it means the OneSignal
SDK would be less compatible with other SDKs and app code.
   - https://github.com/OneSignal/OneSignal-Android-SDK/issues/1355

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1359)


Jeasmine (approved):
Good to know!

PR #1364, opened 2021-06-17
Limit dependencies version ranges to only include stable versions

### LINKED ISSUE #1360: IllegalArgumentException: Comparison method violates its general contract!
**Description:**

Hi there, recently I got velocity alert of an error that the source could be from OneSignal. The information and research are purely on my observation, so please tell me if there's something wrong.

The error is as the title, `IllegalArgumentException`. After looking the stacktrace, the problem could happen because the usage of `WorkManager`, which I didn't use at all in my application and any other third party library I used except OneSignal. It could be happen because of `sort` method in `Array` or because compare in `if` method.

I also looking class that using `WorkManager` in OneSignal lib. I found two class:

1. OSNotificationRestoreWorkManager. I presume this could be the source problem.
2. OSNotificationWorkManager.

I don't exactly know if this error really because of OneSignal or not, please let me know if it's not. 

Thank you.

**Environment**

1. Using OneSignal SDK 4.4.0
2. 99% of affected device are Samsung, you can see the specific device in the attached image
3. 100% affected device OS is Android 11
4. 97% crash happens in background
5. Affecting only 15 user, and its quite consistent, from the first time error appear until now

**Steps to Reproduce Issue:**

Currently, I don't know how to reproduce it, here are some step that I had already try to reproduce the issue, using Android 11 on Note10+ device:

- App in background, send notification, there is no error in Android Studio logcat
- App had been killed, send notification, there is no error in Android Studio logcat

I think there could be some condition to met to reproduce this issue. Either after restarting the application or bluetooth on maybe.

**Anything else:**

```
Fatal Exception: java.lang.IllegalArgumentException: Comparison method violates its general contract!
       at android.os.Parcel.createExceptionOrNull(Parcel.java:2389)
       at android.os.Parcel.createException(Parcel.java:2369)
       at android.os.Parcel.readException(Parcel.java:2352)
       at android.os.Parcel.readException(Parcel.java:2294)
       at android.app.IActivityManager$Stub$Proxy.getHistoricalProcessExitReasons(IActivityManager.java:11243)
       at android.app.ActivityManager.getHistoricalProcessExitReasons(ActivityManager.java:4231)
       at androidx.work.impl.utils.ForceStopRunnable.isForceStopped(ForceStopRunnable.java:171)
       at androidx.work.impl.utils.ForceStopRunnable.forceStopRunnable(ForceStopRunnable.java:210)
       at androidx.work.impl.utils.ForceStopRunnable.run(ForceStopRunnable.java:108)
       at androidx.work.impl.utils.SerialExecutor$Task.run(SerialExecutor.java:91)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:923)
```

And also stack trace below, which not always happen

```
Caused by android.os.RemoteException: Remote stack trace:
	at java.util.TimSort.mergeLo(TimSort.java:777)
	at java.util.TimSort.mergeAt(TimSort.java:514)
	at java.util.TimSort.mergeCollapse(TimSort.java:441)
	at java.util.TimSort.sort(TimSort.java:245)
	at java.util.Arrays.sort(Arrays.java:1492)
```

Device, OS, and Device state info is in image below

Screen Shot 2021-06-16 at 12 35 52

jkasten2:
### One Line Summary
Limit OneSignal's dependencies to know stable versions to prevent alpha and beta version from being automatically used by app projects.

## Details
Set AndroidX, GMS, and FCM max version to know stable versions.
* This fixes an issue where projects can get the latest beta or alpha
versions of dependencies such as work-runtime unless they define their
own version.
* Example: work-runtime:[2.0.0, 2.99.99] would result in 2.7.0-alpha04
as the time of writing.
* Gradle unfortunately does not provide a version range syntax to omit
non-stable versions so the best strategy is to only include a known
stable minor version as the max version in the range.

## Related
* Fixes #1360

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1364)

### DIFF
--- a/OneSignalSDK/onesignal/build.gradle
+++ b/OneSignalSDK/onesignal/build.gradle
@@ -3,8 +3,6 @@ ext {
             minSdkVersion: 16
     ]
 
-    androidXVersion = '[1.0.0, 1.99.99]'
-    androidWorkVersion = '[2.0.0, 2.99.99]'
     firebaseMessagingVersion = '[19.0.0, 22.99.99]'
     googleGMSVersion = '[17.0.0, 17.99.99]'
 }
@@ -37,6 +35,9 @@ android {
 // api || implementation = compile and runtime
 
 // KEEP: version ranges, these get used in the released POM file for maven central
+// NOTE: To prevent consumers from getting alpha or beta version of dependencies listed here
+//         make sure the latest minor value is a known stable version. Using a range for the
+//         patch version is ok, since this allows getting bug fix versions.
 dependencies {
     compileOnly fileTree(dir: 'libs', include: ['*.jar'])
 
@@ -60,11 +61,11 @@ dependencies {
     compileOnly "com.huawei.hms:push:$huaweiHMSPushVersion"
     compileOnly "com.huawei.hms:location:$huaweiHMSLocationVersion"
 
-    api "androidx.cardview:cardview:$androidXVersion"
-    api "androidx.legacy:legacy-support-v4:$androidXVersion"
-    api "androidx.browser:browser:$androidXVersion"
-    api "androidx.appcompat:appcompat:$androidXVersion"
-    api "androidx.work:work-runtime:$androidWorkVersion"
+    api 'androidx.cardview:cardview:[1.0.0, 1.0.99]'
+    api 'androidx.legacy:legacy-support-v4:[1.0.0, 1.0.99]'
+    api 'androidx.browser:browser:[1.0.0, 1.3.99]'
+    api 'androidx.appcompat:appcompat:[1.0.0, 1.3.99]'
+    api 'androidx.work:work-runtime:[2.0.0, 2.5.99]'
 }
 
 apply from: 'maven-push.gradle'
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/build.gradle
```diff
@@ -2,11 +2,6 @@ ext {
     buildVersions = [
             minSdkVersion: 16
     ]
-
-    androidXVersion = '[1.0.0, 1.99.99]'
```
Jeasmine, line 6: Could we have these variables updated instead? Or if there. Any reason behind deleting the variables?

jkasten2: We would have to add new ones for each AndroidX library. I removed them instead and we never override or reassign them and I don't see a reason in the future we would either. It is much easier to read and update versions this way if there isn't a reason to use variables for them.

Jeasmine (approved):

Issue #1357, opened 2021-06-09
OneSignal.java line 3063 com.onesignal.OneSignal.notValidOrDuplicated


JingleBit:
**Description:**

OneSignal notification click crashes my app.

**Environment**

Issue is on Android 9, Android 10 & Android 11

**Steps to Reproduce Issue:**

1. Send OneSignal notification.
2. Click on notification on Android 9 Android 10 or Android 11
3. App is crashed

**Stack trace:**

Fatal Exception: java.lang.RuntimeException: Unable to start receiver com.onesignal.FCMBroadcastReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.onesignal.OSNotificationDataController.notValidOrDuplicated(org.json.JSONObject, com.onesignal.OSNotificationDataController$InvalidOrDuplicateNotificationCallback)' on a null object reference
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:4139)
       at android.app.ActivityThread.access$1600(ActivityThread.java:257)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1993)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:236)
       at android.app.ActivityThread.main(ActivityThread.java:8060)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:656)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:967)

Jeasmine:
This has been released under https://github.com/OneSignal/OneSignal-Android-SDK/releases/tag/4.4.1

PR #1372, opened 2021-06-30
Fix possible NullPointerException under notValidOrDuplicated

Jeasmine:
NullPointerException was happening when calling notValidOrDuplicated under an OSNotificationDataController not inited; this can occur if for some reason OSNotificationDataController got cleaned from memory, or the notification processing doesn't call initWithContext.

Task:
* Init notificationDataController if controller is null
* Add description to notValidOrDuplicated method

Fixes https://github.com/OneSignal/OneSignal-Android-SDK/issues/1357

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1372)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/ADMMessageHandler.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/ADMMessageHandler.java
@@ -93,7 +93,6 @@ public class ADMMessageHandler extends ADMMessageHandlerBase {
          }
       };
       NotificationBundleProcessor.processBundleFromReceiver(context, bundle, bundleReceiverCallback);
-
    }
 
    @Override
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/FCMIntentJobService.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/FCMIntentJobService.java
@@ -26,6 +26,7 @@ public class FCMIntentJobService extends JobIntentService {
         if (bundle == null)
             return;
 
+        OneSignal.initWithContext(this);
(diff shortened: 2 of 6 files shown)


jkasten2 (changes requested):
We should be calling `OneSignal.initWithContext(context)` internally already which initializes `notificationDataController` already. We may have just missed one spot. Possibly just a `Service` as the app's process could have been started from directly in some rare cases.

Adding `OneSignal.initWithContext(context)` will end up call `setupContextListeners` so just adding this where needed should fix the issue. Doing this means we won't have future issues where something else could be missed.

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1372#-:-MdYA2FsCI96QV2eeO-o:b-9y3zl5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @emawby)

Jeasmine (commented):
Yeah, I wanted to avoid searching for every place since we should have one place to handle this, the initWithContext from notification handling. I have seen that most of the notification processing passes through processBundleFromReceiver, so we might want to let that function handle the init. 

I will still leave the null controller check since it will avoid future problems also, and if we have the context available, we should use it.

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1372#-:-MdYevvT3-IM2j-ZMSWq:bc5r7g1)* status: 2 of 6 files reviewed, all discussions resolved (waiting on @emawby and @jkasten2)

jkasten2 (approved):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1372)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @emawby)

PR #1377, opened 2021-07-09
Reverse Activity Trampolining Implementation

jkasten2:
# Description
## One Line Summary
Moved the logic of "should we open the app" from on notification tapped time to notification generation time.

### Pre-existing code & Android 12 restrictions
This PR focuses on providing as direct as possible path from notification open to showing the user an Activity. This change was inspired by the [Notification trampoline restrictions in Android 12](https://developer.android.com/about/versions/12/behavior-changes-12#notification-trampolines) which to quote this Google page was done to "To improve app performance and UX ...".
Technically the OneSignal-Android-SDK 4.0.0 and newer already meets the requirements as the notification open is starting an `Activity`. However this is an invisible `Activity` that conditionally shows another visible Activity after some runtime processing, which in spirit is basically the same as `Broadcast` or `Service` Google is trying to prevent here.

### Limited Android Notification API
OneSignal starts an invisible Activity (NotificationOpenedReceiver) so it can know the user opened the notifications to provide click counts, outcomes, etc on it. This is done so this is automatic and requires zero extra code / configuration by the app developer. Android does not provide a way to attach both an Activity and a background job (Broadcast, Service, etc) to support this without some kind of trampolining.

### Solution implemented - Reverse Activity Trampolining
As noted above, Android does not allow attaching both an `Activity` and a background `Intent`, however it does provide an API to attach multiple Activities with `PendingIntent.getActivities`. This means we can have Android Launch both our invisible `NotificationOpenedReceiver` Activity to track the even while also attaching the visible Activity the app developer wants the user to be shown. You can call this Reverse Activity Trampolining as the OneSignal tracking Activity will be created first but as soon as it is finished Android show the next one in the back stack automatically, hence reverse in the name.

### Changes
`OSNotificationOpenAppSettings` - New class to read global settings for how notification opens behave.
`OSNotificationOpenBehaviorFromPushPayload` - Based on the notification payload and `OSNotificationOpenAppSettings` determine if the notification should open the app or not or if it should open a URL.
`GenerateNotificationOpenIntent` - This logic was moved from `GenerateNotification.java` and has additions to support Reverse Activity Trampolining.
`GenerateNotificationOpenIntentFromPushPayload` - This uses `OSNotificationOpenBehaviorFromPushPayload` to create a `GenerateNotificationOpenIntent` instance. This gets consumed by `GenerateNotification.java`.
`OneSignal.java` - The `startActivity` logic has been pulled out since it has been moved to notification generation time.

### Future
Interduce a OneSignal API to the OneSignal `NotificationServiceExtension` to allow setting an `Activity` `Intent` to control which Activity is shown directly at the notification generation time.
Today if the developer does not want to first show their launcher Activity they have to set `"com.onesignal.NotificationOpened.DEFAULT"` in their `AndroidManifest.xml` and then use `startActivity` from the OneSignal notification open event. Basically the benefits from this PR can't be used when the developer needs to make this customization until OneSignal adds this feature.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1377)


emawby (changes requested):
> Reviewed 6 of 14 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1377)* status: 6 of 14 files reviewed, 2 unresolved discussions (waiting on @jkasten2 and @nan-li)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt, line 77 at r2](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1377#-MgvRYko7jmd_eVdFuQG:-MgvRYko7jmd_eVdFuQH:b-g7er7m) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/238e60c38fef878c59d18ddc8444d82225c96f37/OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt#L77)):*
> ```Kotlin
>             )
> 
>         // Launch desired Activity we want the user to be take to the followed by
> ```

Lets clarify this comment
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationOpenBehaviorFromPushPayload.kt, line 15 at r2](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1377#-MgvOBDp3ySirs8_TcX8:-MgvOBDp3ySirs8_TcX9:bmzctf6) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/238e60c38fef878c59d18ddc8444d82225c96f37/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationOpenBehaviorFromPushPayload.kt#L15)):*
> ```
>             if (!OSNotificationOpenAppSettings.getShouldOpenActivity(context)) return false
>             if (uri != null) return false
>             return true
> ```
Does the following work for this:
`(OSNotificationOpenAppSettings.getShouldOpenActivity(context) && uri == null)`

emawby (approved):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1377)* status: 6 of 14 files reviewed, all discussions resolved (waiting on @nan-li)

PR #1390, opened 2021-07-20
WIP - Move send saved outcomes to foreground

Jeasmine:
* Send saved outcomes when the app is in the foreground, avoid using battery when the app is in background
* Add check to send only once per active session

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1390)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSOutcomeEventsController.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSOutcomeEventsController.java
@@ -32,6 +32,8 @@ class OSOutcomeEventsController {
     @NonNull
     private final OSSessionManager osSessionManager;
 
+    private boolean savedOutcomesSent = false;
+
     public OSOutcomeEventsController(@NonNull OSSessionManager osSessionManager, @NonNull OSOutcomeEventsFactory outcomeEventsFactory) {
         this.osSessionManager = osSessionManager;
         this.outcomeEventsFactory = outcomeEventsFactory;
@@ -79,6 +81,9 @@ class OSOutcomeEventsController {
      * Cached outcomes come from the failure callback of the network request
      */
     void sendSavedOutcomes() {
+        // Only try to send saved outcomes once per session
+        if (savedOutcomesSent)
+            return;
         new Thread(new Runnable() {
             @Override
             public void run() {
@@ -88,6 +93,7 @@ class OSOutcomeEventsController {
                 for (OSOutcomeEventParams event : outcomeEvents) {
                     sendSavedOutcomeEvent(event);
                 }
+                savedOutcomesSent = true;
             }
         }, OS_SEND_SAVED_OUTCOMES).start();
     }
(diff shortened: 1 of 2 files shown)


jkasten2 (changes requested):
If there are outcomes that were never sent (due to connection issues at the time OneSignal.sendOutcome was called) won't we want to send them in the background so they aren't delayed until the next time the app is opened?

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1390#-:-MfB0J9Z25hdHzIOOtMR:b-md9fbo)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @emawby, @Jeasmine, and @tanaynigam)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSOutcomeEventsController.java, line 84 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1390#-MfB-lk6C7Q_nPCgGBHe:-MfB-lk6C7Q_nPCgGBHf:b-a8pzgx) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/1777c13ed263fa7cbfb040d9863caebc2352ab24/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSOutcomeEventsController.java#L84)):*
> ```Java
>      */
>     void sendSavedOutcomes() {
>         // Only try to send saved outcomes once per session
> ```

`savedOutcomesSent` is never set to `false` again so they way the PR is now it is once per app process start instead of session.

Jeasmine (commented):
I see your point, we should retry as fast as possible. But the problem with that is we could end using the battery for this purpose when we should avoid doing too much stuff when in the background and probably working under notification flow. What do you think about that? I will add the session reset check if we agree on this change

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1390#-:-MfYxWDq7-VLZ4AiDhe4:bz2ml3c)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @emawby, @Jeasmine, @nan-li, and @tanaynigam)

jkasten2 (commented):
Outcomes isn't the only things we sync, we also schedule a background updates for any player record data that isn't synced too. So the background data choice we make should effect both.

For more context, we schedule the background job to sync any unsend player data when the app goes out of focus here:
https://github.com/OneSignal/OneSignal-Android-SDK/blob/e0ea2f09f2290824a68a1aabb68843d453629e75/OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java#L173

More details on the background job, we set `setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);` so we don't affect the device's battery when the device is offline for any reason.

Next steps, we probably want to consolidate the code that decides what to sync and when if there are multiple jobs between outcomes and player data. Could make an option to change the sync retry logic if some apps want to customize this is another idea. These are just some ideas we may or may not want to do, open to others as well, and when we think we would want to make such changes.

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1390#-:-MfZW822EdnT17Ym9bg1:be6c5a0)* status: all files reviewed, 1 unresolved discussion (waiting on @emawby, @Jeasmine, @nan-li, and @tanaynigam)

jkasten2:
Since we want to always sync any missing data when we can we don't want to limit this to the foreground. Closing this PR.

PR #1395, opened 2021-07-26
Work in progress: IAM "flickering" when changing activity after clicked

nan-li:
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1395)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/WebViewManager.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/WebViewManager.java
@@ -313,7 +313,7 @@ class WebViewManager extends ActivityLifecycleHandler.ActivityAvailableListener 
             // Navigate to new activity while displaying current IAM
             if (messageView != null)
                 messageView.removeAllViews();
-            showMessageView(lastPageHeight);
+            showMessageView(lastPageHeight); // removing this line prevented quick 2nd re-display in Jon's demo app
         } else {
             // Activity rotated
             calculateHeightAndShowWebViewAfterNewActivity();


OneSignalSDK/onesignal/src/main/java/com/onesignal/WebViewManager.java
```diff
@@ -313,7 +313,7 @@ else if (!lastActivityName.equals(currentActivityName)) {
             // Navigate to new activity while displaying current IAM
             if (messageView != null)
                 messageView.removeAllViews();
-            showMessageView(lastPageHeight);
+            showMessageView(lastPageHeight); // removing this line prevented quick 2nd re-display in Jon's demo app
```
jkasten2, line 316: Check out the git blame for this:
https://github.com/OneSignal/OneSignal-Android-SDK/blame/main/OneSignalSDK/onesignal/src/main/java/com/onesignal/WebViewManager.java#L312

It links to this PR
https://github.com/OneSignal/OneSignal-Android-SDK/pull/1247

Which notes: "If user navigates with backpress while an IAM is being display, IAM should continue displaying until user dismiss it."

Can you test to see if this breaks this functionality?

jkasten2: If the above does break then we probably need to check if the In-App Message is "dismissing" and omit trying to reshow it in this case. That is, the In-App Message is visible but is in the middle of playing it's dismiss animation.

nan-li: Yes removing that line does break the above-mentioned functionality.

nan-li: I found that _either_ **lowering** `ACTIVITY_FINISH_AFTER_DISMISS_DELAY_MS = 600` to something like 200 **_OR_** **raising** `ACTIVITY_INIT_DELAY = 200` to something like 600 seems to work. Don't quite understand the specifics of the functions that use these two variables nor the implications.

https://github.com/OneSignal/OneSignal-Android-SDK/blob/702634d4b3e083f1c0bbfc5c1797d3072bfd623e/OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java#L56-L57

nan-li:
I found that _either_ **lowering** `ACTIVITY_FINISH_AFTER_DISMISS_DELAY_MS = 600` to something like 200 **_OR_** **raising** `ACTIVITY_INIT_DELAY = 200` to something like 600 seems to work. Don't quite understand the specifics of the functions that use these two variables nor the implications.

https://github.com/OneSignal/OneSignal-Android-SDK/blob/702634d4b3e083f1c0bbfc5c1797d3072bfd623e/OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java#L56-L57

jkasten2 (commented):
I looked at the git blame for `ACTIVITY_FINISH_AFTER_DISMISS_DELAY_MS` and `ACTIVITY_INIT_DELAY` but it didn't have good commit know around this. However from the comments we have some details. Based on the name, comments, and what I recall from when I was working on this I can provide more details.

`ACTIVITY_INIT_DELAY` this is a delay used to wait for the new Activity to become "ready". I believe was to ensure any Android transition animations have finished. I think this number was picked based on what visually looked correct based on some device testing. Unrelated to this but there probably is a way we can get this from Android instead of using a static timer.

`ACTIVITY_FINISH_AFTER_DISMISS_DELAY_MS` - This dependent on our IAM dismiss animation. There are two ways a user can dismiss an IAM; 1. tapping on the "X" 2. Swiping down to dismiss.
3. I think these animations also happen when a new Activity is shown. (haven't confirmed this though) I though we did this so that if something change Activity from under the IAM (such as opening a push notification, Activity being reshown due to device rotation, or something else internal to their app) we bring the IAM back tot the foreground. And use use the animations so the transition looks smooth. 

Changing these time values as you noted do fix the issue where an IAM action auto dismisses and also performs an action to opening a new Activity but could also break what is noted above. Also it could break later if we change any of these animations in the future.

If we instead use a "dismissing" state to omit try to reshowing the IAM under any of the situations above then we decouple this fix completely from any animations.

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1395#-:-MfdypyG2Z3NAjO0Yh62:bkcu6tw)* status: all files reviewed, 1 unresolved discussion (waiting on @jkasten2)

jkasten2 (commented):
For reference here is where the animation happens: https://github.com/OneSignal/OneSignal-Android-SDK/blob/e0ea2f09f2290824a68a1aabb68843d453629e75/OneSignalSDK/onesignal/src/main/java/com/onesignal/DraggableRelativeLayout.java#L195

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1395#-:-Mfe54Tn7XQ8sbkrcaFx:b-2o6hh3)* status: all files reviewed, 1 unresolved discussion (waiting on @jkasten2)

PR #1400, opened 2021-07-30
make PendingIntents compatible with Android 12

AndreasBoehm:
Because Android 12 forces mutability to be specified when creating a PendingIntent I got a couple of errors in logcat when starting my app targeting Android SDK 31 (Android 12).
https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability

This PR fixes those issues on OneSignal side. (https://github.com/OneSignal/OneSignal-Android-SDK/issues/1408)
The be fully compatible we would need to upgrade workmanager to at least `2.7.0-alpha05` but I think that we should wait with that until `2.7.0` is stable.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1400)


jkasten2:
@AndreasBoehm Thanks for the PR. There was a recent change in PR #1377 that moved this logic around.  I recreated PR #1411 based on your changes and reference this PR to give you credit.

PR #1404, opened 2021-08-05
Fix `EXTERNAL_USER_ID_AUTH_HASH` not popped and old one sent to backend

nan-li:
* The EXTERNAL_USER_ID_AUTH_HASH is never popped so it remains in the existing UserState syncValues even when the EXTERNAL_USER_ID is set to "" (empty string).

* The sendJson ends up with the new EXTERNAL_USER_ID but with old EXTERNAL_USER_ID_AUTH_HASH

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1404)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/UserState.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/UserState.java
@@ -15,6 +15,7 @@ import java.util.Set;
 
 import static com.onesignal.UserStateSynchronizer.APP_ID;
 import static com.onesignal.UserStateSynchronizer.EMAIL_AUTH_HASH_KEY;
+import static com.onesignal.UserStateSynchronizer.EXTERNAL_USER_ID;
 import static com.onesignal.UserStateSynchronizer.EXTERNAL_USER_ID_AUTH_HASH;
 import static com.onesignal.UserStateSynchronizer.SMS_AUTH_HASH_KEY;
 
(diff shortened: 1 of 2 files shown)


nan-li:
Either just make a check for new auth hash before send off, OR reconsider how the external ID is removed (when external id removed and set to "", auth hash also gets popped if the server validates the auth hash)

jkasten2:
Ah I the change here will fix one of the issues, but I think there are two problems.
1. The cached value overriding the new value (which you just fixed in this PR)
2. Remove external user id doesn't clear it from the cached values when it should, so it show up in `syncValues` again.

Lastly the changes for external user id will probably need to be done to sms and email auth hashes as well.

nan-li:
I spent some time but can't understand how the `dependValues` work differently than the `syncValues`. I do see that, for example, the `SMS Auth Hash` and `Email Auth Hash` are in both `syncValues` and `dependValues`.

I also have not been able to reproduce the incorrect behavior I noted right after making the `persistState()` changes. I think the UserStates were in an incorrect state to start with because of auth_hashes not working correctly, before implementing this change...

emawby (approved):
Great job on the unit tests!

OneSignalSDK/onesignal/src/main/java/com/onesignal/UserState.java
```diff
@@ -317,6 +319,17 @@ private void loadState() {
 
     void persistState() {
         synchronized(LOCK) {
+            // pop the EXTERNAL_USER_ID_AUTH_HASH from the UserState
+            try {
+                if (syncValues.has(EXTERNAL_USER_ID_AUTH_HASH) &&
+                        ((syncValues.has(EXTERNAL_USER_ID) && syncValues.get(EXTERNAL_USER_ID).toString() == "") || !syncValues.has(EXTERNAL_USER_ID))) {
+                    syncValues.remove(EXTERNAL_USER_ID_AUTH_HASH);
+                    // should EXTERNAL_USER_ID also be removed or kept in state as ""?
```
nan-li, line 327: Nit

nan-li, line 327: Remove comment/make it a statement that "" is kept as external ID

jkasten2 (approved):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1404)* status: all files reviewed, 1 unresolved discussion (waiting on @jkasten2)

jkasten2 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1404#-:-MkhUa9S0bokpYTmTWHl:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jkasten2)

PR #1421, opened 2021-08-23
Enable transparent IAM backgrounds

emawby:
We need to set our webview and cardview's background colors to enable a completely transparent background. We can safely always do this because the HTML for the IAM will already set a background color. In the future the HTML will be able to optionally include a semi transparent background image.


nan-li (approved):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1421)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine and @jkasten2)

PR #1424, opened 2021-08-27
Add synchronization to  to fix possible race condition

rgomezp:
Fixes a null pointer exception occurring due to attempt to call `showView` on a null `messageView` object, likely due to it being set to null by a background thread resulting in a race condition. By locking the `messageView` to one thread at a time, we can prevent this issue from occurring.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1424)


OneSignalSDK/onesignal/src/main/java/com/onesignal/WebViewManager.java
```diff
@@ -38,6 +38,7 @@
     private static final String TAG = WebViewManager.class.getCanonicalName();
     private static final int MARGIN_PX_SIZE = dpToPx(24);
     private static final int IN_APP_MESSAGE_INIT_DELAY = 200;
+    static final Object messageViewSyncLock = new Object() {};
```
Jeasmine, line 41: Could this be private?

Jeasmine (approved):

PR #1443, opened 2021-09-28
Remove a key from Shared Prefs if its value is `null`

nan-li:
- This was a problem for `logoutEmail` and `logoutSMSNumber`
- The above methods end up calling `OneSignalPrefs.saveString` to save their channel IDs with the value of `null`
- However, when we `flushBufferToDisk`, the value of null falls through the "if-else if" checks and nothing happens, when it should be removed.

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
@@ -204,6 +204,8 @@ class OneSignalPrefs {
                             editor.putLong(key, (Long)value);
                         else if (value instanceof Set)
                             editor.putStringSet(key, (Set<String>)value);
+                        else if (value == null)
+                            editor.remove(key);
                     }
                     prefHash.clear();
                 }


OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java
```diff
@@ -204,6 +204,8 @@ else if (value instanceof Long)
                             editor.putLong(key, (Long)value);
                         else if (value instanceof Set)
                             editor.putStringSet(key, (Set<String>)value);
+                        else if (value == null)
+                            editor.remove(key);
```
nan-li, line 208: Or perhaps this should be `editor.putString(key, null)`? 
There is some difference with how `remove()` is treated as it is done when `commit()` is called, whereas the other put methods are written once `apply()` or `commit()` is called.

But the current changes is working.

emawby (commented):
Looks good to me. I am not sure about the difference between removeKey and putting null but if this works it is fine with me

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1443#-:-MlBFbCR2Prbs31Z0Zvh:bmf8q0m)* status: all files reviewed, 1 unresolved discussion (waiting on @Jeasmine and @jkasten2)

jkasten2 (approved):
Nice find! LGTM

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1443#-:-MlBZzOIENT_klpbyq7x:b-3rdqhi)* status: all files reviewed, 1 unresolved discussion (waiting on @Jeasmine and @nan-li)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java, line 208 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1443#-MkimDkZkNklDn6Sl5-i-r1-208:-MlBZWxd8RybBBOoya1m:brnhr84) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/6c8906ff936f51dfe4e7319b35df0e5474a06b41/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java#L208)):*

Or perhaps this should be `editor.putString(key, null)`? 
There is some difference with how `remove()` is treated as it is done when `commit()` is called, whereas the other put methods are written once `apply()` or `commit()` is called.

But the current changes is working.
</blockquote></details>

I think `.remove(key)` is more clear, as using `.putString(key, null)` also implies a type of `string`.

PR #1461, opened 2021-10-18
Add `toJSONObject` method to `OSInAppMessage`

nan-li:
# Description
## One Line Summary
Add JSON stringifier to the public `OSInAppMessage` class and refactor JSON stringifier for internal `OSInAppMessageInternal` class to return JSON key of `"messageId"` instead of `"id"`.

### Motivation
We want the public `OSInAppMessage` to have a `toJSONObject` method for the wrappers to use. This JSON Object only has one item of `"messageId"`. Since IAMs have the property `messageId` and our documentation refer to `messageId`, this key was chosen over `"id"`.

The need for this method first arose when adding the InAppMessage Lifecycle Handler to React Native.
We also modified the `OSInAppMessageInternal` class's `toJSONObject` to return a JSON key of `"messageId"` instead of `"id"` for clarity and consistency to the public class.

### Background Context
Previously, the current `OSInAppMessageInternal` class was named `OSInAppMessage` before being refactored into an internal InAppMessage class and a public InAppMessage class. 

> OSInOSInAppMessageInternal extends OSInAppMessage  

Worth noting, `"id"` is expected instead of `"messageId"` when parsing JSON from the backend, so it is used in the constructor.

### Scope
We didn't use any InAppMessage `toJSONObject` in SDK code, except in unit tests (see below) and those are removed now. The existing `OSInAppMessageInternal` `toJSONObject` method was _package-private_ (made _public_ in this PR) so users _couldn't_ have used this method prior to this PR. This means changing the JSON key from `"id"` to `"messageId"` has no public API _changes_ (but introduces a new method to the public).

Currently, all of the `OSInAppMessage` objects are actually `OSInAppMessageInternal` objects. We don't create any superclass OSInAppMessage objects in the existing codebase. This means when a user calls  `OSInAppMessage.toJSONObject()`, they will receive the results of `OSInAppMessageInternal.toJSONObject()`.

# Testing
## Unit testing
Some unit tests were creating an IAM based off the JSON of another IAM by calling its `toJSONObject()`. 
With the above changes, some tests in `InAppMessageIntegrationTests` broke  because IAM constructors expect `"id"` instead of `"messageId"` when parsing the JSON, so  we need to replace that with `"id"` for creating an IAM.

#### Added method: 
`convertIAMtoJSONObject(OSInAppMessageInternal inAppMessage)`
- just replaces key `"messageId"` with `"id"`
- refactored unit test code to use this method instead of `toJSONObject()`

#### Removed a constructor: 
`OSTestInAppMessageInternal(OSInAppMessageInternal inAppMessage)`
- this constructor passed the JSON from `inAppMessage.toJSONObject()`
- instead, use `convertIAMtoJSONObject` and pass this JSON into constructor

#### Added test: 
`testInAppMessageInternalToJSONObject_messageId`
- tests `OSInAppMessageInternal` (not `OSInAppMessage`)
- only tests for `"messageId"`, the primary focus

## Manual testing
✅  Tested by calling `OSInAppMessage.toJSONObject()` in demo app:
- [X] Android 11 Emulator
- [X] Android 12 Emulator

### Example of the JSON
``` json
{
  "messageId": "abe479ed-dc54-4446-898a-a82b67cfd5aa",
  "variants": {"all":{"default":"2a8934de-d3c1-4734-8b7c-796cd691456b"}},
  "displayDuration": 0,
  "redisplay": {"limit":1,"delay":0},"triggers":[],
  "has_liquid": false
}
```

✅  Tested in **react-native-onesignal** via the InAppMessageLifecycleHandler by logging the `message` and `message.messageId`
- When a user logs the `message` object, they receive the entire JSON object above, not just the `messageId`

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [ ] REST API requests
   - [X] Public API changes -  exposes a new method (will update documentation)

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1461)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
@@ -29,7 +29,12 @@ package com.onesignal;
 
 import androidx.annotation.NonNull;
 
+import org.json.JSONException;
+import org.json.JSONObject;
+
 public class OSInAppMessage {
+
+    public static final String IAM_ID = "messageId";
 
     /**
      * The unique identifier for this in-app message
@@ -44,6 +49,18 @@ public class OSInAppMessage {
     @NonNull
     public String getMessageId() {
         return messageId;
+    }
+
+    public JSONObject toJSONObject() {
+        JSONObject mainObj = new JSONObject();
+        try {
+            mainObj.put(IAM_ID, messageId);
+        }
+        catch(JSONException e) {
+            e.printStackTrace();
+        }
+
+        return mainObj;
     }
 
 }
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageInternal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageInternal.java
@@ -145,7 +145,8 @@ class OSInAppMessageInternal extends OSInAppMessage {
         return parsedTriggers;
     }
 
-    JSONObject toJSONObject() {
+    @Override
+    public JSONObject toJSONObject() {
         JSONObject json = new JSONObject();
 
         try {
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageInternal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageInternal.java
@@ -17,7 +17,9 @@ import java.util.Set;
 
 class OSInAppMessageInternal extends OSInAppMessage {
 
-    private static final String IAM_ID = "id";
+    // "id" is expected instead of "messageId" when parsing JSON from the backend
+    private static final String ID = "id";
+    private static final String IAM_ID = "messageId";
     private static final String IAM_VARIANTS = "variants";
     private static final String IAM_TRIGGERS = "triggers";
     private static final String IAM_REDISPLAY_STATS = "redisplay";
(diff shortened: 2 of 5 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java
```diff
@@ -46,4 +49,16 @@ public String getMessageId() {
         return messageId;
     }
 
+    public JSONObject toJSONObject() {
+        JSONObject mainObj = new JSONObject();
+        try {
+            mainObj.put("messageId", messageId);
```
Jeasmine, line 55: I think that in other new classes, all strings are put inside a constant inside the file. Let me know if you agree with putting "messageId" inside a private static final String

jkasten2: ah good catch, I agree with that

nan-li: Ah ok! I have seen that practice but didn't realize it is preferred.

nan-li (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1461)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @Jeasmine and @jkasten2)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java, line 55 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1461#-MmK521OLutvGPH8EdY7-r1-55:-MnSJAFaBaLatVQOrn_T:b-896fja) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/3efad9ccef494ae57f934f510df35cb997c5f7d0/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessage.java#L55)):*

Ah ok! I have seen that practice but didn't realize it is preferred.
</blockquote></details>

Done!

jkasten2 (approved):
:lgtm:

> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1461#-:-MngyViZ0v8ydfxSVehh:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @emawby, @Jeasmine, and @rgomezp)

Issue #1465, opened 2021-10-26
Android 4.4 and older devices fail to register to onesignal.com to due it's TLS1.2+ requirement


jkasten2:
# Description
Android 4.4 and older devices fail to connect to onesignal.com and print the following SSL error in the logcat:
```java
2021-10-25 18:22:29.410 7613-7718/com.onesignal.sdktest W/OneSignal: OneSignalRestClient: null Error thrown from network stack. 
    javax.net.ssl.SSLHandshakeException: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0x55870ae8: Failure in SSL library, usually a protocol error
    error:14077102:SSL routines:SSL23_GET_SERVER_HELLO:unsupported protocol (external/openssl/ssl/s23_clnt.c:714 0x52adcd5c:0x00000000)
        at com.android.org.conscrypt.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:448)
        at com.android.okhttp.Connection.upgradeToTls(Connection.java:146)
        at com.android.okhttp.Connection.connect(Connection.java:107)
        at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:294)
        at com.android.okhttp.internal.http.HttpEngine.sendSocketRequest(HttpEngine.java:255)
        at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:206)
        at com.android.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:345)
        at com.android.okhttp.internal.http.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:296)
        at com.android.okhttp.internal.http.HttpURLConnectionImpl.getResponseCode(HttpURLConnectionImpl.java:503)
        at com.android.okhttp.internal.http.HttpsURLConnectionImpl.getResponseCode(HttpsURLConnectionImpl.java:136)
        at com.onesignal.OneSignalRestClient.startHTTPConnection(OneSignalRestClient.java:183)
        at com.onesignal.OneSignalRestClient.access$100(OneSignalRestClient.java:44)
        at com.onesignal.OneSignalRestClient$4.run(OneSignalRestClient.java:112)
        at java.lang.Thread.run(Thread.java:841)
[...]

PR #1469, opened 2021-10-26
Fix outside package target activities (Launch URL) not opening when notification is tapped on

jkasten2:
# Description
## One Line Summary
Fix notification open with Launch URL not showing any Activities, this was a regression bug introduced in PR #1377.

### Motivation
Notification Launch URL is an active supported feature used by OneSignal customers so ensuring it works correctly is high priority.

### Scope
The main goal is to fix Launch URL not working but the code change in this PR effects all notification opens. The "opens" noted here is scoped to how `Activities` are shown to the user. This includes the `Activity` flags used to start `Activities` and the backstace it generates.
No changes to how OneSignal consumes or generates `Intent` data.

### Why the bug was happening?
The issue was due to an Activity backstack not being built correctly. Both `NotificationOpenedReceiver` and the target `Activity` were launched with `FLAG_ACTIVITY_NEW_TASK` however this means when `NotificationOpenedReceiver` is finished it does not bring up the target since it is on a different task. The Android flag `allowTaskReparenting` allows this to happen, however this can effect the backstack of the main task which we don't want to do.

### How the bug was fixed?
Setting `android:taskAffinity` sets `FLAG_ACTIVITY_NEW_TASK` and `allowTaskReparenting` for us. By setting the value to "" (empty string) it prevents any side effects on existing tasks as it means it has no affinity to associate it with. This allowed us to clean up some runtime logic to add `FLAG_ACTIVITY_NEW_TASK` and others we no longer need now.

A caveat is that on Android 5.1 and older `android:taskAffinity=""` doesn't give us the `allowTaskReparenting` behavior, nor does setting it explicitly work. To account for this an `Activity` named `NotificationOpenedReceiverAndroid22AndOlder` was created without the `android:taskAffinity` parameter.

#### Tasks and the back stack
Basic understand of this topic is recommend
* https://developer.android.com/guide/components/activities/tasks-and-back-stack

*Source: https://developer.android.com/guide/topics/manifest/activity-element#aff*

#### Notification Trampolines
Background on what "Notification Trampolines" are and how a "Reverse Activity Trampoline" works.
* https://onesignal.com/blog/android-12-notification-open-changes/
* [Reverse Activity Trampolining Implementation PR #1377](https://github.com/OneSignal/OneSignal-Android-SDK/pull/1377)

# Testing
## Unit testing
Added test to ensure correct Activity is used for newer vs older versions of Android.
No existing exist unit tests to ensure correct Activity is infocus or the backstack is built correctly. Robolectric doesn't have fakes / mocking to simulate to this level.

## Manual testing
Each scenario was tested with the following devices:
   - ⚠️Can't test older Android versions, due to [know TLS1.2 issue](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1465)
   - ✅Android 5.0 Emulator
   - ✅Android 5.1 Emulator
   - ✅Android 6.0 Emulator
   - ✅Android 7.1 Emulator
   - ✅Android 8.1 Emulator
   - ✅Android 9 LG G7
   - ✅Android 9 Amazon Fire 7 (KFMUWI)
   - ✅Android 10 Emulator
   - ✅Android 11 Emulator
   - ✅Android 12 Emulator

### Default notification
* [X] App Foreground
* [X] App Background - by pressing back button
* [X] App Background - by pressing home button
* [X] App not running - Task still in recent list, resumes app's task correctly
* [X] App not running - Recent list cleared, app cold starts new task

### HTTPS URL notification
* [X] App Foreground
* [X] App Background - by pressing home button
* [X] App not running - Recent list cleared, app cold starts new task
 
### com.onesignal.NotificationOpened.DEFAULT set to DISABLE
* [X] App Foreground
* [X]  App Background - by pressing home button
* [X] App not running - swiped away app

### Data URL scheme - appname://test
Tested by adding the following `<intent-filter>` entries for the following Activities in the example app's `AndroidManifest.xml`.
```xml
<activity
    android:name=".activity.SplashActivity"
    android:exported="true">
  <intent-filter>
      <action android:name="android.intent.action.MAIN" />
      <category android:name="android.intent.category.LAUNCHER" />
  </intent-filter>
  <intent-filter>
      <action android:name="android.intent.action.VIEW"/>
      <category android:name="android.intent.category.DEFAULT"/>
      <category android:name="android.intent.category.BROWSABLE"/>
      <data android:scheme="osexamplesplash"/>
  </intent-filter>
</activity>

<activity
    android:name=".activity.MainActivity"
    android:exported="true">
    <intent-filter>
        <action android:name="android.intent.action.VIEW"/>
        <category android:name="android.intent.category.DEFAULT"/>
        <category android:name="android.intent.category.BROWSABLE"/>
        <data android:scheme="osexample"/>
    </intent-filter>
</activity>
```

Each scenario below includes testing twice, once with `osexamplesplash:// ` another with `osexample://`.

* [X] App Foreground
* [X] App Background - by pressing home button
* [X] App Background - App swiped away
* [X] App not running - Recent list cleared, app cold starts new task

# Affected code checklist
   - [X] Notifications
      - [ ] Display
      - [X] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] I have filled out all REPLACE sections above
   - [X] PR does one thing
   - [X] Public API changes are intended and conform to existing APIs

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1469)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/AndroidManifest.xml
+++ b/OneSignalSDK/onesignal/src/main/AndroidManifest.xml
@@ -140,6 +140,12 @@
           android:theme="@android:style/Theme.Translucent.NoTitleBar"
           android:exported="true" />
 
+        <activity
+            android:name="com.onesignal.NotificationOpenedReceiverAndroid22AndOlder"
+            android:noHistory="true"
+            android:excludeFromRecents="true"
+            android:theme="@android:style/Theme.Translucent.NoTitleBar"
+            android:exported="true" />
     </application>
 
     <!-- NOTE: See release version for tags with placeholders -->
--- a/OneSignalSDK/onesignal/src/main/AndroidManifest.xml
+++ b/OneSignalSDK/onesignal/src/main/AndroidManifest.xml
@@ -136,6 +136,7 @@
           android:name="com.onesignal.NotificationOpenedReceiver"
           android:noHistory="true"
           android:excludeFromRecents="true"
+          android:taskAffinity=""
           android:theme="@android:style/Theme.Translucent.NoTitleBar"
           android:exported="true" />
 
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt
@@ -3,6 +3,7 @@ package com.onesignal
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import androidx.annotation.RequiresApi
 
 class GenerateNotificationOpenIntent(
     private val context: Context,
@@ -35,6 +36,7 @@ class GenerateNotificationOpenIntent(
         )
     }
 
+    @RequiresApi(android.os.Build.VERSION_CODES.M)
     private fun getNewBaseIntentAndroidAPI23Plus(): Intent {
         return Intent(
             context,
@@ -43,6 +45,7 @@ class GenerateNotificationOpenIntent(
     }
 
     // See NotificationOpenedReceiverAndroid22AndOlder.java for details
+    @Deprecated("Use getNewBaseIntentAndroidAPI23Plus instead for Android 6+")
     private fun getNewBaseIntentAndroidAPI22AndOlder(): Intent {
         val intent = Intent(
             context,
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt
@@ -10,26 +10,56 @@ class GenerateNotificationOpenIntent(
     private val startApp: Boolean
 ) {
 
-    private val notificationOpenedClass: Class<*> = NotificationOpenedReceiver::class.java
+    private val notificationOpenedClassAndroid23Plus: Class<*> = NotificationOpenedReceiver::class.java
+    private val notificationOpenedClassAndroid22AndOlder: Class<*> = NotificationOpenedReceiverAndroid22AndOlder::class.java
 
     fun getNewBaseIntent(
         notificationId: Int,
     ): Intent {
-        // We use SINGLE_TOP and CLEAR_TOP as we don't want more than one OneSignal invisible click
-        //   tracking Activity instance around.
-        val intentFlags =
-            Intent.FLAG_ACTIVITY_SINGLE_TOP or
-            Intent.FLAG_ACTIVITY_CLEAR_TOP
+        val intent =
+            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M)
+                getNewBaseIntentAndroidAPI23Plus()
+            else
+                getNewBaseIntentAndroidAPI22AndOlder()
 
-        return Intent(
-            context,
-            notificationOpenedClass
-        )
+        return intent
         .putExtra(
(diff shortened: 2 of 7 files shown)


jkasten2 (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1469)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @emawby, @Jeasmine, and @nan-li)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt, line 11 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1469#-MnC1ViV33UP6StI2Q5V:-MnCp2Au1ID8QxgRFifA:bkmco1w) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/86522908e234da4ddaae40d4f345d64ba4c62a8c/OneSignalSDK/onesignal/src/main/java/com/onesignal/GenerateNotificationOpenIntent.kt#L11)):*
<details><summary><i>Previously, emawby (Elliot Mawby) wrote…</i></summary><blockquote>

Are we using this boolean anymore?
</blockquote></details>

Yes, the only place it gets used now is  at the top of `getIntentAppOpen`.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedReceiverAndroid22AndOlder.kt, line 43 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1469#-Mn2aXZX6W_Tn71hDI2V-r1-43:-MnCpOhnDm3S00K0D7-C:by2b5f9) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/86522908e234da4ddaae40d4f345d64ba4c62a8c/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedReceiverAndroid22AndOlder.kt#L43)):*
<details><summary><i>Previously, Jeasmine (JNahui) wrote…</i></summary><blockquote>

will it make sense to have a single Base NotificationOpenedReceiver and both NotificationOpenedReceiver and NotificationOpenedReceiverAndroid22AndOlder implement it?
</blockquote></details>

ah yes, I'll do that. It will de-dup a few lines of code and prevent missing the other class when updating any of the logic in it.

jkasten2 (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1469)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @emawby, @Jeasmine, and @nan-li)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedReceiverAndroid22AndOlder.kt, line 43 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1469#-Mn2aXZX6W_Tn71hDI2V-r1-43:-MnSrEnH0BEQ-z-RFi46:b-y07wtj) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/86522908e234da4ddaae40d4f345d64ba4c62a8c/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedReceiverAndroid22AndOlder.kt#L43)):*

ah yes, I'll do that. It will de-dup a few lines of code and prevent missing the other class when updating any of the logic in it.
</blockquote></details>

I addressed this in commit 17f63402f9504ee12dbf83518dee70fd726ec722

emawby (approved):

Jeasmine (approved):

Issue #1444, opened 2021-09-30
The app crashes when an exception occurs inside OneSignal in SDK 4.4.1+


ushi3:
**Description:**

In version 4.4.1 and above, when OneSignal receives a push notification, an exception that occurs within OneSignal SDK to  crashes the app.

We have updated the SDK from 4.4.0 to 4.6.0 and released the app. Since then, we've been receiving a lot of crash reports.
After investigating the cause, it was found that it was due to the changes of #1356. And SDK containing it was released in version 4.4.1.

**What's wrong with #1356?**

**TLDR;**
**After 4.4.1, The thread being OneSignal notification process is not catching the exception properly.**

In #1356, implementation of `OSNotificationWorkerManager.NotificationWorker` changed `Worker` to `ListenableWorker`.
`Worker`'s `doWork()` runs on **worker-thread**, while `ListenableWorker`'s `startWork()` runs on **main-thread**.
That difference causes a change in the subsequent processing. (**probably not intended**)

Both processes call `processNotificationData()`, and then reaches `OSNotificationReceivedEvent.complete()`. `OSNotificationReceivedEvent.complete()` is the part that causes this issue.

Here exist code that if call on main-thread, execute process on other thread. [code](https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationReceivedEvent.java#L80-L90)
Below 4.4.1, this if statement is `false`. So, continue processing on current-thread (= worker-thread).
After 4.4.1, this statement is `true`, run in other thread made on instant.

If the method `OSNotificationReceivedEvent.processNotification()` is execute on **worker-thread**, the exception thrown here will be caught by` Worker.startWorker() `. However, if the method is execute on **instant thread**, the exception will not be caught. App crashes when actual exception thrown.

Do you really throw some exception in this method? to anwer YES.
Many of the exceptions are caused by Android NotificationManager API and device platform bugs.
For example, on some devices such as Samsung, LG, the system notification service was killed (reason don't know why), NotificationManager throw DeadSystemException.

**Environment**

Device OS : Android 7.0, 7.1.1, 8.0, 8.1 (at least)
OneSignal SDK : 4.4.1+

**Steps to Reproduce Issue:**

**Confirm that execution-thread has changed around 4.4.1 before-after.**
1. Enabled Debug
2. mark breakpoint`OSNotificationReceivedEvent.complete()#L71~`
[...]

Issue #1419, opened 2021-08-20
Some notifications are duplicated


socialError:
**Description:**

In our news app, lately we have weird problem that some notifications are duplicated. It doesn't happen a lot, but I managed  to get logs with more info and it looks like there is some problem in android sdk. I tried to find out what causes this issue, but i was kind of lost. Based on provided log problem is that notification with notificationID=742dfbd9-6952-470c-a84e-006b90c0ced7 comes to OSRemoteNotificationReceivedHandler remoteNotificationReceived method twice and is also shown twice (when device is restarted etc, also both of the same notifications are restored ). In logs is also my log line with `OSRemoteNotificationReceivedHandler remoteNotificationReceived called, notificationId: 742dfbd9-6952-470c-a84e-006b90c0ced7` from remoteNotificationReceived method so you can see when was that method called.

Here is log:
https://www.dropbox.com/s/ax7tw4ro8kz7z6c/onesignal_duplicated_notification.log?dl=0

Here is my remoteNotificationReceived method. I believe I don't do anything wrong here, which could causes this issue.
```
@Override
    public void remoteNotificationReceived(Context context, OSNotificationReceivedEvent notificationReceivedEvent) {
        OSNotification notification = notificationReceivedEvent.getNotification();
        
        JSONObject data = notification.getAdditionalData();
        OSMutableNotification mutableNotification = notification.mutableCopy();
        
        if(data != null && data.has(OneSignalNotificationOpener.PAYLOAD_TAG)) {
            try {
                String topic = data.getString(OneSignalNotificationOpener.PAYLOAD_TAG);
                mutableNotification.setExtender(builder -> builder.setSubText(topic));
            }
            catch (JSONException e) {
                FirebaseCrashlytics.getInstance().recordException(e);
                e.printStackTrace();
            }
        }

        notificationReceivedEvent.complete(mutableNotification);

        Log.e("onesignal", " OSRemoteNotificationReceivedHandler remoteNotificationReceived called, notificationId: " +  mutableNotification.getNotificationId() + " androidNotificationId: " +  mutableNotification.getAndroidNotificationId() + " body: " + mutableNotification.getBody());
    }
```

If you need any more info from me, feel free to ask. We really need to solve this issue somehow. Thanks a lot

**Environment**

1. 4.4.2
2. OneSignal SDK added using the official docs (mavenCentral)

**Steps to Reproduce Issue:**

I have no steps to reproduce it, it happens completely random for us.

[...]

PR #1484, opened 2021-11-10
Use two Workers to display notification and send receive receipt

nan-li:
# Description
## One Line Summary
Make (1) displaying a notification and (2) sending the receive receipt (with delay) two separate Android `Workers` instead of one `ListenableWorker`.

### Motivation
We were getting reports of **duplicated** notifications such as in this [issue](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1419), as well as a few other reports. I was able to reproduce this and confirm the notification is only sent from the dashboard once.

### Background Context
Prior, the `NotificationWorker` was a `ListenableWorker`. Its work is completed after the receive receipt is sent, for which there is a _delay of 0-25 seconds_. The work is enqueued only once, but if anything happens before the work is done (during the delay before receive receipt is sent), it will re-start its `startWork`, which is where the notification is processed and displayed, leading to displaying a duplicated notification on the device.

_Even prior to that_, the `NotificationWorker` was a `Worker`, and was changed to a `ListenableWorker` for the feature of future completion when the receive receipt delay was implemented in the SDK.

### Scope
Even with testing and trying different scenarios, there can always be edge cases or situations that aren't foreseen and cause a problem.
- Can sometimes the receive receipt not be sent?
- Receive receipts should be retried if app is killed or work is not successful, but potentially could the network request be sent twice?
In terms of receive receipt behavior, I don't believe we are changing the behavior.

Removed the `OSDelayTaskController` and `MockDelayTaskController` as these two classes are not needed anymore.

The reversion back to using `Worker` instead of `ListenableWorker` may also fix https://github.com/OneSignal/OneSignal-Android-SDK/issues/1444 as well.

# Testing
## Unit testing
Three of the existing unit tests for sending receive receipts failed after this change to a delayed `Worker` instead of the `OSDelayTaskController`. The receive receive request is not sent (but is queued), leading to unit test failure.

> Exception: Main looper has queued unexecuted runnables. This might be the cause of the test failure. You might need a shadowOf(getMainLooper()).idle() call.

After trying a few different solutions with unreliable results:
- setting the delay to 1 second and awaiting a long time
- making `shadowOf(getMainLooper()).idle()` calls
- making more `threadAndTaskWait()` calls
- calling `time.advanceSystemAndElapsedTimeBy()`

The conclusion is to **omit the delay**, as thus also not test the delay, in unit tests. I am open to other ideas for addressing this problem that lets us have the delay.

**In addition**, the constraint of `setRequiredNetworkType(NetworkType.CONNECTED)` was causing these same tests to fail as well even though checking `isConnected()` returns `true`. The method `beginEnqueueingWork` is shadowed to not use this constraint, and the `ShadowReceiveReceiptController` is used in all tests related to sending or not sending receive receipts.

#### Removed test: 
`testNotificationReceivedNoSendReceivedRequest_Delay`
- since we are not having the ReceiveReceiptWorker run after a delay, we will not test the delay, and other existing receive receipt tests are sufficient

## Manual Testing
The receive receipt delay is sent after a random delay of 0-25 seconds. For testing only, I bumped up that range to 24-25 seconds.

✅ Ran example app on Pixel Emulator **API 30** 
✅ Ran example app on Pixel Emulator **API 31**

| App State  | NotificationReceivedHandler | ForegroundHandler | Result
| ------------- | ------------- | --- | --- |
| Foreground  | not set  | is set, fires, and calls `complete(null)` | not displayed + no RR sent |
| Foreground or Background | is set, fires, and calls `complete(null)`  | not set | not displayed + no RR sent |
| Background  | is not set  | is set but not fired | displays + RR sent |
| Foreground  | is set, fired, and calls `complete(notif)`  | is set, fired, and calls `complete(notif)` |  displays + RR sent | 
| Foreground  | is set, fired, and calls `complete(notif)`  | is set, fired, and calls `complete(null)` | not displayed + no RR sent | 

Also tested manually on Pixel Emulator **API 28, 30, 31**:
- ✅ Receive notification when app in background, don't click notification, and open and kill app twice during the delay before receive receipt is sent -> receive receipt is sent and shows in dashboard and notification not duplicated.
- ✅ Receive notification when app is in foreground, don't click notification, send a tag, unsubscribe, and then kill app during the delay before receive receipt is sent -> receive receipt is sent and shows in dashboard (which shows the new tag and user as unsubscribed) and notification not duplicated.
- ✅ Turn off data/wifi, send notification, wait a while, turn on data/wifi and receive notification. Then turn off data/wifi shortly after notification comes through. Wait two minutes and turn on data/wifi back on and receive receipt is sent.

# Affected code checklist
   - [X] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [X] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484)


jkasten2 (changes requested):
> ❓ Turn off data/wifi, send notification, wait a while, turn on data/wifi and receive notification. Then turn off data/wifi shortly after notification comes through, receive receipt is sent with failure due to "device is offline" but Worker is completed successfully regardless of response and won't rerun. I think this is ok, because we haven't accounted for this in the previous 2 implementations anyway (on failure, we don't retry).

To account for the device going offline we can add a constraint that device must be online before the received receipt job can fire.
This can be done with the following code.
```java
Constraints constraints = new Constraints.Builder()
                    .setRequiredNetworkType(NetworkType.CONNECTED)
                    .build(); 
OneTimeWorkRequest onetimeJob = new OneTimeWorkRequest.Builder(YourJob.class)
                    .setConstraints(constraints).build(); // or PeriodicWorkRequest
```

> Reviewed 18 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-:-MoAvxvU-3PqDDKOkhPe:b-colno)* status: 18 of 19 files reviewed, 2 unresolved discussions (waiting on @jkasten2 and @nan-li)

___
*[OneSignalSDK/onesignal/build.gradle, line 51 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-MoAztkJ3cGghZqSbF2U:-MoAztkJ3cGghZqSbF2V:b-4fue5l) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/2dd19486dbb57e2b7528bd123c81cb4863b3a961/OneSignalSDK/onesignal/build.gradle#L51)):*
> ```Gradle
>     implementation 'com.google.android.gms:play-services-base:[17.0.0, 17.6.99]'
> 
>     implementation("androidx.concurrent:concurrent-futures:$androidConcurrentFutures")
> ```

Nice, good catch removing a dependency when the code using it was removed.
___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java, line 653 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-MoB1zWV-5epHOpQs3Lc:-MoB1zWV-5epHOpQs3Ld:b-mcd75e) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/5ec456d71a6250525e2547cf893115c488aef3b6/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java#L653)):*
> ```Java
> 
>    static int getRandomDelay(int minDelay, int maxDelay) {
>       return (int) Math.round(Math.random() * (maxDelay - minDelay + 1) + minDelay);
> ```

I believe this has a bug where the value could be 1 higher than max if `Math.random()` is 1 or close enough to it.
Recommend using this instead as it is simpler:
```java
return random.nextInt(max + 1 - min) + min;
```

jkasten2 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-:-MoKNp22-Mz_XiRRFwqu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

jkasten2 (approved):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

jkasten2 (changes requested):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484)* status: all files reviewed, 1 unresolved discussion (waiting on @Jeasmine and @nan-li)

___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowReceiveReceiptController.java, line 27 at r4](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-MoLuaBS81FVLNB9ewqM:-MoLuaBS81FVLNB9ewqN:b-juak5x) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/552f4679d8dd0662930c209f5a0d31d8d10a8d19/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowReceiveReceiptController.java#L27)):*
> ```Java
>     @Implementation
>     public void beginEnqueueingWork(Context context, String osNotificationId) {
>         if (!isReceiveReceiptEnabled()) {
> ```

This `isReceiveReceiptEnabled()` check and the `inputData` below is duplicated with the code in the actual class.
To clean this up we can make another method in `OSReceiveReceiptController` which splits these up so we have to override less.

jkasten2 (approved):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484)* status: all files reviewed, 1 unresolved discussion (waiting on @Jeasmine and @nan-li)

___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowReceiveReceiptController.java, line 27 at r4](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-MoLuaBS81FVLNB9ewqM:-MoLvphb6IhuNAmAfOjr:b8b9kv6) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/552f4679d8dd0662930c209f5a0d31d8d10a8d19/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowReceiveReceiptController.java#L27)):*

This `isReceiveReceiptEnabled()` check and the `inputData` below is duplicated with the code in the actual class.
To clean this up we can make another method in `OSReceiveReceiptController` which splits these up so we have to override less.
</blockquote></details>

Since this is only to clean up the tests we can differ to another PR.

jkasten2 (approved):
:lgtm:

> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-:-MoM1w-nBIGmGNVQryWU:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Jeasmine)

___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowReceiveReceiptController.java, line 27 at r4](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1484#-MoLuaBS81FVLNB9ewqM:-MoM1xWv6JJgaKXZvEwp:barvllc) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/552f4679d8dd0662930c209f5a0d31d8d10a8d19/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowReceiveReceiptController.java#L27)):*

Since this is only to clean up the tests we can differ to another PR.
</blockquote></details>

Nice, abstracted exactly what is different, the constraints!

Issue #1423, opened 2021-08-26
Error Caused by java.lang.ClassNotFoundException: com.huawei.hms.api.HuaweiApiAvailability


fillipeoliveira-hotmart:
**Description:**

After add Dexguard on our project, I start to receive some errors of `NoClassDefFoundError` on `HuaweiApiAvailability`. I search on other issues, like https://github.com/OneSignal/OneSignal-Android-SDK/issues/1194 and https://github.com/OneSignal/OneSignal-Android-SDK/issues/1085 but the only solution is keep everything from OneSignal on dexguard rules.

**Environment**

OneSignal 4.4.2 via Gradle + Dexguard
`implementation ('com.onesignal:OneSignal:4.4.2')`

**Steps to Reproduce Issue:**

None

**Anything else:**
```
Fatal Exception: java.lang.NoClassDefFoundError: Failed resolution of: Lcom/huawei/hms/api/HuaweiApiAvailability;
       at com.onesignal.OSUtils.isHMSCoreInstalledAndEnabled(OSUtils.java:289)
       at com.onesignal.OSUtils.supportsHMS(OSUtils.java:317)
       at com.onesignal.OSUtils.getDeviceType(OSUtils.java:349)
       at com.onesignal.OSUtils.initializationChecker(OSUtils.java:112)
       at com.onesignal.OneSignal.init(OneSignal.java:711)
       at com.onesignal.OneSignal.setAppId(OneSignal.java:625)
       at com.onesignal.OneSignal.reassignDelayedInitParams(OneSignal.java:1028)
       at com.onesignal.OneSignal.onRemoteParamSet(OneSignal.java:759)
       at com.onesignal.OneSignal$5.complete(OneSignal.java:969)
       at com.onesignal.OneSignalRemoteParams.processJson(OneSignalRemoteParams.java:205)
       at com.onesignal.OneSignalRemoteParams.access$100(OneSignalRemoteParams.java:12)
       at com.onesignal.OneSignalRemoteParams$1.onSuccess(OneSignalRemoteParams.java:150)
       at com.onesignal.OneSignalRestClient$5.run(OneSignalRestClient.java:269)
       at java.lang.Thread.run(Thread.java:929)
```
```
Caused by java.lang.ClassNotFoundException: com.huawei.hms.api.HuaweiApiAvailability
       at com.onesignal.OSUtils.isHMSCoreInstalledAndEnabled(OSUtils.java:289)
[...]

PR #1492, opened 2021-11-23
Proguard -assumenosideeffects Compatibility

jkasten2:
# Description
## One Line Summary
Fix compatibility issue with apps that use `-assumenosideeffects` with `java.lang.Class.getName()` with Proguard or Dexguard. _(No issue with R8.)_

### Motivation
To fix issue #1423 which cause a crash on some devices when aggressive minification setting `-assumenosideeffects` with `java.lang.Class.getName()` is used.

### Scope
This effects code paths that check if a Java class exists for feature detection. No logic was changed, just a proxy class was added to prevent Proguard from striping code.

## How this prevents the issue
Addressed this by adding a new private `opaqueHasClass` method and utilized it in all existing class exist helper methods.
This method has the "Keep" annotation on it so proguard will never change it's code and will not make a wrong assumptions about side effects.

### Before code change
Here in the smali output you can see the try-catch was removed by Proguard from `hasHMSAvailabilityLibrary`, even when the rule `-keep class com.onesignal.OSUtils** {*;}` or `@Keep` is added to try to prevent this modification.
```smali
.method private static n()Z
    .registers 1
    const-class v0, Lcom/huawei/hms/api/HuaweiApiAvailability;
    const/4 v0, 0x1
    return v0
.end method
```

### After code change
After you see `opaqueHasClass` is kept as-is. `hasHMSAvailabilityLibrary` is renamed which is ok, the important this is that the try-catch is kept which is required.
```smali
.method private static opaqueHasClass(Ljava/lang/Class;)Z
    .registers 1
    const/4 p0, 0x1
    return p0
.end method

// hasHMSAvailabilityLibrary
.method private static o()Z
    .registers 1

    :try_start_0
        const-class v0, Lcom/huawei/hms/api/HuaweiApiAvailability;
        invoke-static {v0}, Lcom/onesignal/OSUtils;->opaqueHasClass(Ljava/lang/Class;)Z
        move-result v0
    :try_end_6
        .catch Ljava/lang/NoClassDefFoundError; {:try_start_0 .. :try_end_6} :catch_7
        return v0
    :catch_7
        move-exception v0

    const/4 v0, 0x0
    return v0
.end method
```

# Testing
## Unit testing
No unit tests where added since this is a build issue and it has to run on a device to verify.

## Manual testing
Tested on a FireOS tablet and a Android 12 emulator without the FCM library.

Disabled R8 in `gradle.properties`, this forces the Android Gradle Plugin to use Proguard. 
````gradle.properties
android.enableR8 = false
android.enableR8.libraries = false
````

Tested with `classpath 'net.sf.proguard:proguard-gradle:6.1.1'` and with the following `proguard-rules.pro` file contents:
```proguard-rules.pro
-ignorewarnings
-optimizationpasses 1
-dontskipnonpubliclibraryclasses
-optimizations !field/*,!class/merging/*

-assumenosideeffects public class java.lang.Object {
    public final java.lang.Class getClass();
}
-assumenosideeffects public final class java.lang.Class {
    public java.lang.String getName();
}
```

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes
   - [X] Push Token Registration 

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1492)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
@@ -138,36 +138,29 @@ class OSUtils {
       return subscribableStatus;
    }
 
-   // Interim method that prevents proguard from making a wrong assumption that NoClassDefFoundError
-   // catches are not needed when rules include "-assumenosideeffects" with getName().
-   // -assumenosideeffects public final class java.lang.Class {
-   //    public java.lang.String getName();
-   // }
-   // Using Class instead of String as class renames would result incorrectly not finding the class
-   // Using class.getName() as if no method is called then the try-catch would be removed.
-   // This @Keep annotation is also key so this method does not get inlined
+   // Interim method that works around Proguard's overly aggressive assumenosideeffects which
+   // ignores keep rules.
+   // This is specifically designed to address Proguard removing catches for NoClassDefFoundError
+   // when the config has "-assumenosideeffects" with
+   // java.lang.Class.getName() & java.lang.Object.getClass().
+   // This @Keep annotation is key so this method does not get removed / inlined.
+   // Addresses issue https://github.com/OneSignal/OneSignal-Android-SDK/issues/1423
    @Keep
-   private static boolean hasClass(Class<?> clazz) {
-      try {
-         clazz.getName();
-         return true;
-      } catch (NoClassDefFoundError e) {
-         return false;
-      }
+   private static boolean opaqueHasClass(Class<?> _class) {
+      return true;
    }
 
    static boolean hasFCMLibrary() {
       try {
-         return hasClass(com.google.firebase.messaging.FirebaseMessaging.class);
+         return opaqueHasClass(com.google.firebase.messaging.FirebaseMessaging.class);
       } catch (NoClassDefFoundError e) {
          return false;
       }
    }
 
-   @Keep
    static boolean hasGMSLocationLibrary() {
       try {
-         return hasClass(com.google.android.gms.location.LocationListener.class);
+         return opaqueHasClass(com.google.android.gms.location.LocationListener.class);
       } catch (NoClassDefFoundError e) {
          return false;
       }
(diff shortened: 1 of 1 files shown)


adamschlesinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1492)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Jeasmine, @jkasten2, @nan-li, and @rgomezp)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java, line 161 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1492#-MpD4-MNDC0JRmw9r5Tb:-MpD4-MNDC0JRmw9r5Tc:ba1di0m) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/b5bfd597210051ea38d98309989576ef0f801d95/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java#L161)):*
> ```Java
>    static boolean hasFCMLibrary() {
>       try {
>          return hasClass(com.google.firebase.messaging.FirebaseMessaging.class);
> ```

Isn't this (and the additional below) try/catch redundant now?

jkasten2 (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1492)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @adamschlesinger, @Jeasmine, @nan-li, and @rgomezp)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java, line 161 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1492#-MpD4-MNDC0JRmw9r5Tb:-MphvrVv5G8rXoBlWleC:bwnzjex) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/b5bfd597210051ea38d98309989576ef0f801d95/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java#L161)):*
<details><summary><i>Previously, adamschlesinger (Adam Schlesinger) wrote…</i></summary><blockquote>

Isn't this (and the additional below) try/catch redundant now?
</blockquote></details>

I simplified the code in `hasClass` (renamed to `opaqueHasClass`) so the 2nd catch isn't needed. I also updated my PR description to show the smali code differences and the configs I used to test with.

PR #1508, opened 2022-01-11
Speculative fix for null layout params.

brendoncoin:
# Description
## One Line Summary
Checks for null layoutParams when setting IAM height.

### Motivation
Prevent crashes. Crash reports received of NPE on InAppMessageView.java:156.

### Scope
Don't change the height if the layoutParams are null. Wait for the height to be changed when the IAM is displayed.

### OPTIONAL - Other
N/A

# Testing
## Unit testing
None added.

## Manual testing
Unable to reliably reproduce the failure case, so the fix could not be verified, but a manual test fo receiv

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ x ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/InAppMessageView.java
@@ -153,6 +153,13 @@ class InAppMessageView {
                 }
 
                 ViewGroup.LayoutParams layoutParams = webView.getLayoutParams();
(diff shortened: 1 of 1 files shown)


emawby (approved):
Thank you for contributing!

PR #1515, opened 2022-01-14
Replace FocusDelaySyncJobService by Worker

Jeasmine:
# Description
## One Line Summary
A JobService is being used for delaying the on_focus service call, JobServices are not stopped by the system, this makes the app process be alive in the background. Replace FocusDelaySyncJobService by a Worker

### Motivation
Avoid battery draining by the app in the background

# Testing
## Unit testing
Return tests to the previous state before this change https://github.com/OneSignal/OneSignal-Android-SDK/pull/1198

## Manual testing
Tested under Android 11.

1. Ensure that battery saving is off and the app is not restricted when in the background
2. launch the app and then minimize it to the background
3. Run adb shell dumpsys power
4. Check that non PARTIAL_WAKE_LOCK is alive for OneSignal

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [x] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1515)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSFocusHandler.kt
```diff
@@ -0,0 +1,95 @@
...
+            onFocusDoWork()
+            return Result.success()
+        }
+    }
+
+    companion object {
+        private var backgrounded = false
+        private var completed = false
+
+        fun onFocusDoWork() {
```
emawby, line 81: So this should only be called when losing focus not when regaining focus? Can we name it so that it is clear that this is a worker when losing focus?

emawby (changes requested):

emawby (approved):

PR #1479, opened 2021-11-04
Add TTL notification display control

Jeasmine:
# Description
### One-line Summary
Notifications were being shown independent of the TTL due to Android App Standby Buckets 

#### Motivation
App Standby Buckets avoid users being spammed by unused applications. If notification is not set with high priority and the app is marked as not used, the notification will be received by the app, but showed the next time the user opens it, this is not the same as restoring, since they will vibrate and behave as they were just received. Because of that we need to add a TTL check before display logic.

#### Scope
Notification complete logic now checks if notification is between TTL time.
HMS and GCM tests for TTL have been added

#### Background context

# Affected code checklist
   - [x] Notifications
      - [x] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] I have filled out all **REQUIRED** sections above
   - [X] PR does one thing
     - If it is hard to explain how any codes changes are related to each other then it most likely needs to be more than one PR
   - [X] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [X] I have included test coverage for these changes, or explained why they are not needed
   - [X] All automated tests pass, or I explained why that is not possible
   - [X] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [X] Code is as readable as possible.
      - Simplify with less code, followed by splitting up code into well named functions and variables, followed by adding comments to the code.
   - [X] I have reviewed this PR myself, ensuring it meets each checklist item
      - WIP (Work In Progress) is ok, but explain what is still in progress and what you would like feedback on. Start the PR title with "WIP" to indicate this.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1479)

PR #1528, opened 2022-02-03
Fix `ttl` and related current time logic

nan-li:
# Description
## One Line Summary
Add `ttl` and `sentTime` (and `smallIcon`) to an OSNotification constructor and change usages of `currentThreadTimeMillis` (time thread has been active) to `currentTimeMillis` (epoch time) instead.

### Motivation
This PR is related to some features of https://github.com/OneSignal/OneSignal-Android-SDK/pull/1479.

The `protected OSNotification(OSNotification notification)` constructor was missing `ttl` and `sentTime` so that when the NSE is used along with the foreground handler, and the NSE passes on a `mutableCopy()`, the notification that is processed by the foreground handler next has `ttl` and `sentTime` missing, and thus defaulted to `0`, which ends up not displaying due to `isNotificationWithinTTL()` check returning `false` erroneously.

Unrelated, but `smallIcon` was also missing so I added it as well.

While making the above changes, I noticed that the `currentTimeInSeconds` for `isNotificationWithinTTL()` would be close to zero. When we check `isNotificationWithinTTL()` to determine if we should display it, we should use `currentTimeMillis` (milliseconds since epoch) instead of `currentThreadTimeMillis` (the time a thread has been active)  to determine the current time that will be used in the logic. Note that the `google.sent_time` is also in milliseconds since epoch.

While working on this, I looked at other places that used thread time inaccurately and changed to using `currentTimeMillis` instead.

### Scope
I think using `currentThreadTimeMillis` was not a problem in the past because it was only the backup value when `google.sent_time` was not in the payload. 

# Testing
## Unit testing
Added test `testNotificationProcessingAndForegroundHandler_callCompleteWithMutableNotification_displays()` and handler `RemoteNotificationReceivedHandler_notificationReceivedCallCompleteWithMutableNotification`. The test NSE calls `complete()` with a `mutableCopy()` of the notification that is used by the foreground handler later. We test that the notification will display.

The two tests `notShowNotificationPastTTL` and `shouldSetExpireTimeCorrectlyWhenMissingFromPayload` are now failing because they were using thread time. I modified these to use current non-thread time and they now pass.

- [x] All unit tests pass

## Manual testing
Tested using Android emulator Pixel 5 on API 30 with the demo app in the SDK. Sent notification to the device in the foreground and it didn't display before. Now it displays. Also regularly checked values of `sentTime`, `ttl` and `currentTimeInSeconds` throughout testing.

# Affected code checklist
   - [X] Notifications
      - [X] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1528)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationBundleProcessor.java
@@ -236,7 +236,7 @@ class NotificationBundleProcessor {
             values.put(NotificationTable.COLUMN_NAME_MESSAGE, notificationJob.getBody().toString());
 
          // Set expire_time
-         long sentTime = jsonPayload.optLong(OSNotificationController.GOOGLE_SENT_TIME_KEY, OneSignal.getTime().getCurrentThreadTimeMillis()) / 1_000L;
+         long sentTime = jsonPayload.optLong(OSNotificationController.GOOGLE_SENT_TIME_KEY, OneSignal.getTime().getCurrentTimeMillis()) / 1_000L;
          int ttl = jsonPayload.optInt(OSNotificationController.GOOGLE_TTL_KEY, OSNotificationRestoreWorkManager.DEFAULT_TTL_IF_NOT_IN_PAYLOAD);
          long expireTime = sentTime + ttl;
          values.put(NotificationTable.COLUMN_NAME_EXPIRE_TIME, expireTime);
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotification.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotification.java
@@ -146,7 +146,7 @@ public class OSNotification {
          return;
       }
 
-      long currentTime = OneSignal.getTime().getCurrentThreadTimeMillis();
+      long currentTime = OneSignal.getTime().getCurrentTimeMillis();
       if (currentJsonPayload.has(GOOGLE_TTL_KEY)) {
          sentTime = currentJsonPayload.optLong(GOOGLE_SENT_TIME_KEY, currentTime) / 1_000;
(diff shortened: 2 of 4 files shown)


jkasten2 (approved):
Code changes look good to me! Made one comment on a test improvement, feel free to make the fix and merge without another approval.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1528#-:-MvM8p6FDDnxX9pidK24:b-5tpibl)* status: all files reviewed, 1 unresolved discussion (waiting on @nan-li)

___
*[OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java, line 1990 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1528#-MvM79oJ3EcW3Pz9Mrzu:-MvM79oJ3EcW3Pz9Mrzv:bnxop5z) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/528a9085d356ac0839cef77cc3b51867ae474e89/OneSignalSDK/unittest/src/test/java/com/test/onesignal/GenerateNotificationRunner.java#L1990)):*
> ```Java
>    public void testNotificationProcessingAndForegroundHandler_callCompleteWithMutableNotification_displays() throws Exception {
>       // 1. Setup correct notification extension service class
>       startRemoteNotificationReceivedHandlerService("com.test.onesignal.GenerateNotificationRunner$" +
> ```

Recommend using a hard reference to this class so if any code refactor is done in the future this will be a compile error instead of a failed test.
```java
startRemoteNotificationReceivedHandlerService(RemoteNotificationReceivedHandler_notificationReceivedCallCompleteWithMutableNotification.getClass().getCanonicalName());
```

Issue #1514, opened 2022-01-13
Null pointer crash processing rest client response


joerogers:
**Description:**

OneSignal is crashing the app with this stack.  Has happened 9 times in last week since we deployed our app with OneSignal the first time.

```
Fatal Exception: java.lang.NullPointerException
       at java.util.Objects.requireNonNull(Objects.java:203)
       at com.onesignal.OneSignal.init(OneSignal.java:837)
       at com.onesignal.OneSignal.setAppId(OneSignal.java:692)
       at com.onesignal.OneSignal.reassignDelayedInitParams(OneSignal.java:1136)
       at com.onesignal.OneSignal.onRemoteParamSet(OneSignal.java:844)
       at com.onesignal.OneSignal$6.complete(OneSignal.java:1077)
       at com.onesignal.OneSignalRemoteParams.processJson(OneSignalRemoteParams.java:206)
       at com.onesignal.OneSignalRemoteParams.access$100(OneSignalRemoteParams.java:12)
       at com.onesignal.OneSignalRemoteParams$1.onSuccess(OneSignalRemoteParams.java:151)
       at com.onesignal.OneSignalRestClient$5.run(OneSignalRestClient.java:283)
       at java.lang.Thread.run(Thread.java:919)
```

I believe what is happening taking a cursory look at the logic. 

1. OneSignal is initializing.  
2. It requests the remote params are initialized
3. The initial code running the initializer is suspended by the OS (these things happen)
4. The http request completes which also calls init.
5. However, the context is already set so it assumes the full initialization is complete
6. Crashes trying to access the following line because outcomeEventsController is still null

```
outcomeEventsController.sendSavedOutcomes();
```

Wrapping this with a null check should still work as if my theory is right, the original init call will complete and invoke it, but now outcomeEventsController will not be null.

**Environment**

1. What version of the Android SDK are you using?
    4.6.3 (but doesn't look like the code changed in 4.6.5)
    
2. How did you add the SDK to your project (eg. maven)
    Gradle via maven

**Steps to Reproduce Issue:**

Unknown, nothing else in the stack and very little in production logs which implies this is during startup.

**Anything else:**

See above

emawby:
@joerogers Thank you for reporting. Have you had any luck reproducing this crash? If not we can at least try wrapping that controller in a null check like you mention

joerogers:
[...]

Issue #1374, opened 2021-07-01
Background crashes in com.onesignal.OSOutcomeEventsController.sendSavedOutcomes(OSOutcomeEventsController.java:36)


chokokatana:
**Description:**

App has started crashing in the background with the following stacktrace:
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
       at com.onesignal.OSOutcomeEventsController.sendSavedOutcomes(OSOutcomeEventsController.java:36)
       at com.onesignal.OneSignal.init(OneSignal.java:812)
       at com.onesignal.OneSignal.setAppId(OneSignal.java:685)
       at com.onesignal.OneSignal.reassignDelayedInitParams(OneSignal.java:1096)
       at com.onesignal.OneSignal.onRemoteParamSet(OneSignal.java:819)
       at com.onesignal.OneSignal$5.complete(OneSignal.java:1037)
       at com.onesignal.OneSignalRemoteParams.processJson(OneSignalRemoteParams.java:206)
       at com.onesignal.OneSignalRemoteParams.access$100(OneSignalRemoteParams.java:12)
       at com.onesignal.OneSignalRemoteParams$1.onSuccess(OneSignalRemoteParams.java:151)
       at com.onesignal.OneSignalRestClient$5.run(OneSignalRestClient.java:269)
       at java.lang.Thread.run(Thread.java:818)
```
According to crashlytics 50% motorola, 38% Android 6, 100% crashes in the background.

**Environment**
One signal gradle plugin [0.12.10, 0.99.99], library at [4.4.0, 4.99.99].
Target and compile SDK 30, minsdk 17. 
Not using jcenter.

**Steps to Reproduce Issue:**

Don't know what is triggering the problem, no idea how to reproduce. It does not happen on my development devices and only today have I started getting crash reports from the field. Given that 100% crashes are in the background I suspect some push is happening, but nothing from our part as far as I know.

uniquevik:
When can we get an updates on this? I am using latest version and found a crash report for this bug. @Jeasmine

mahmoudElfeel67:
@Jeasmine i also have crashes coming from this is there an update ???

nan-li:
Hi @chokokatana, @android-guru-vikas, @mahmoudElfeel67, thanks for reporting and I apologize for not getting a reply in earlier. 

[...]

PR #1539, opened 2022-03-03
Fix NPE for `outcomeEventsController` during initialization

nan-li:
# Description
## One Line Summary
Lazily initialize and access the `OneSignal.outcomeEventsController`, adding a null check when invoking its methods that wasn't checked before, as it may not have been initialized yet in the critical initialization step of `setupContextListeners()`.

### Motivation
This is a speculative fix for a non-reproducible `NullPointerException` of `outcomeEventsController.sendSavedOutcomes()` during initialization, as reported by users in production [here](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1374) and [here](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1514).

### Context
During initialization, `initWithContext()` and `setAppId()` run. The first method also initializes the `outcomeEventsController` in the method `setupContextListeners()` [see code](https://github.com/OneSignal/OneSignal-Android-SDK/blob/10867586a325eaf9e69292e5d3e288083768ce8a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java#L874).

After the appContext and appId are set by the above methods, the `init()` method runs, which calls `OSOutcomeEventsController.sendSavedOutcomes()` [towards the method's end](https://github.com/OneSignal/OneSignal-Android-SDK/blob/10867586a325eaf9e69292e5d3e288083768ce8a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java#L837). 

It **may** be that in some cases, the `init()` method is called before `initWithContext()` is fully complete, resulting in a null `outcomeEventsController` instance that `init()` relies on.

One GitHub reporter speculated it is due to adding `setLanguage` functionality in `4.4.0`, and there are now more calls to `initWithContext` added in `4.4.1` for more FCM entry points. 

It is possible that multiple threads are initializing and one hangs after setting the `appContext` but hasn't finished `setupContextListeners`.

This is ultimately speculative and it may be some other reason that `outcomeEventsController` is `null`.

### Implementation Details
When we want to access `OneSignal.outcomeEventsController`, use lazy initialization and singleton pattern to access and create it via `getOutcomeEventsController()`. Use synchronization when creating the object.

There are some methods where `outcomeEventsController` is called such as in `sendUniqueOutcome()` where there is already a `null` check and the method returns if `outcomeEventsController == null`. In these methods I didn’t make any changes as it already had its own null check logic.

### Scope
This affects OneSignal initialization.

### Potential Problems
If the root issue is in fact that `setupContextListeners()` isn't completed before `init()` runs, it may be possible that `languageContext` isn't initialized either. In this case, while we move past the NPE for `outcomeEventsController`, we may encounter a NPE for `languageContext` next, such as in `registerUserTask()` or when it is passed as an argument to `getInAppMessageController()`.

However, that may be far enough down the line that it won't pose a problem.

# Testing
## Unit testing
Added unit test to simulate `setupContextListeners()` not completing:

- Add unit test `initWithContext_setupContextListenersNotCompleted_doesNotProduceNPE` to reproduce the NPE of `outcomeEventsController` in `init()`.

- Shadow the `OneSignal.setupContextListeners()` method to simulate it not completing (thus not initializing the `outcomeEventsController`).

- However, the `languageContext` initialization is needed or else it leads to a subsequent NPE for this property in `registerUserTask()`.

- Change access of `languageContext` in OneSignal.java from private to package-private so it can be accessed in unit tests.

- All other unit tests pass

## Manual testing
App built with Android Studio 2020.3.1 with the OneSignal example app on a Pixel 5 with API 30.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing
   - [x] I have personally tested this on my device, but is insufficient as this error is not reproducible.

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1539)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -442,6 +442,22 @@ public class OneSignal {
    @Nullable private static OSOutcomeEventsController outcomeEventsController;
    @Nullable private static OSOutcomeEventsFactory outcomeEventsFactory;
    @Nullable private static OSNotificationDataController notificationDataController;
+   private static final Object outcomeEventsControllerSyncLock = new Object() {};
+
+   static OSOutcomeEventsController getOutcomeEventsController() {
+      if (outcomeEventsController == null) {
+         synchronized(outcomeEventsControllerSyncLock) {
+            if (outcomeEventsController == null) {
+               if (outcomeEventsFactory == null) {
+                  OneSignalDbHelper dbHelper = getDBHelperInstance();
+                  outcomeEventsFactory = new OSOutcomeEventsFactory(logger, apiClient, dbHelper, preferences);
+               }
+               outcomeEventsController = new OSOutcomeEventsController(sessionManager, outcomeEventsFactory);
+            }
+         }
+      }
+      return outcomeEventsController;
+   }
 
    @SuppressWarnings("WeakerAccess")
    public static String sdkType = "native";
@@ -834,7 +850,7 @@ public class OneSignal {
       initDone = true;
       OneSignal.Log(LOG_LEVEL.VERBOSE, "OneSignal SDK initialization done.");
 
-      outcomeEventsController.sendSavedOutcomes();
+      getOutcomeEventsController().sendSavedOutcomes();
 
       // Clean up any pending tasks that were queued up before initialization
       taskRemoteController.startPendingTasks();
@@ -871,8 +887,7 @@ public class OneSignal {
             outcomeEventsFactory = new OSOutcomeEventsFactory(logger, apiClient, dbHelper, preferences);
 
          sessionManager.initSessionFromCache();
-         outcomeEventsController = new OSOutcomeEventsController(sessionManager, outcomeEventsFactory);
-         outcomeEventsController.cleanCachedUniqueOutcomes();
+         getOutcomeEventsController().cleanCachedUniqueOutcomes();
       }
    }
 
@@ -950,7 +965,7 @@ public class OneSignal {
          logger.debug("Starting new session with appEntryState: " + getAppEntryState());
 
          OneSignalStateSynchronizer.setNewSession();
-         outcomeEventsController.cleanOutcomes();
+         getOutcomeEventsController().cleanOutcomes();
          sessionManager.restartSessionIfNeeded(getAppEntryState());
          getInAppMessageController().resetSessionLaunchTime();
          setLastSessionTime(time.getCurrentTimeMillis());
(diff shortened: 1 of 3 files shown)


jkasten2 (changes requested):
Logic changes look good, have a concern about the shadow changes though and it effecting other tests.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1539#-:-MynmvidBM9oj-mIUvY_:bi5bxwl)* status: all files reviewed, 1 unresolved discussion (waiting on @Jeasmine and @nan-li)

___
*[OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOneSignal.java, line 30 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1539#-MynlPTBBhgOIPcgYE68:-MynlPTBBhgOIPcgYE69:b4tzwf2) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/2cfe353184141cab6963d94b9848a412e8ef1413/OneSignalSDK/unittest/src/test/java/com/onesignal/ShadowOneSignal.java#L30)):*
> ```Java
>     */
>    @Implementation
>    public static void setupContextListeners(boolean wasAppContextNull) {
> ```

There are a number of tests using this shadow, this method changes the behavior quite a bit. Can we confirm if this change is needed at all?

If it is could we do one of the following?:
1. By default call the original logic, we could add an if statement to check a flag and only set it in the test that needs this.
2. Create a different class that shadows the OneSignal class and only use it in the test you need.

nan-li:
> There are a number of tests using this shadow, this method changes the behavior quite a bit. Can we confirm if this change is needed at all?
>
> If it is could we do one of the following?:
> 1. By default call the original logic, we could add an if statement to check a flag and only set it in the test that needs this.
> 2. Create a different class that shadows the OneSignal class and only use it in the test you need.

Good catch, I've created a new shadow class `ShadowOneSignalWithMockSetupContextListeners` just for the added unit test that has the modified `setupContextListeners()` method.

jkasten2 (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1539)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @emawby and @Jeasmine)

nan-li:
`shouldSetExpireTimeCorrectlyWhenMissingFromPayload` has been regularly failing here, off by 1 second. I've re-ran the build a few times in the past 1-2 days.

Running the unit tests locally twice, they all pass.

Issue #1543, opened 2022-03-09
[Android 12] Crash when tapping on Notification action button due to use of Intent.ACTION_CLOSE_SYSTEM_DIALOGS


jkasten2:
## Issue
The [`handleDismissFromActionButtonPress`](https://github.com/OneSignal/OneSignal-Android-SDK/blob/4.7.0/OneSignalSDK%2Fonesignal%2Fsrc%2Fmain%2Fjava%2Fcom%2Fonesignal%2FNotificationOpenedProcessor.java#L73) method in this SDK uses [`Intent.ACTION_CLOSE_SYSTEM_DIALOGS`](https://developer.android.com/reference/android/content/Intent#ACTION_CLOSE_SYSTEM_DIALOGS) however in Android API Level 31 (Android 12) this requires a new permission.

## Solution
The logic in `handleDismissFromActionButtonPress` was written in 2016 in PR #18 so it is likely things have changed with Android. We should attempt to remove `handleDismissFromActionButtonPress` to see if it correctly closes the notification shade in Android 12. If it does we can then go back to older versions of Android support by the SDK (Android 4.1 at the time of writing) to see if we still need it as well. If it is still required we can simply omit it if the device is running API level 31 or newer.

## Workaround
Setting your project to use`androidSdkVersion` to `30` would most likely workaround this issue until this is resolved.

vincent-paing:
Will submit a PR to fix this

nan-li:
The crash fix has been released in [version 4.7.1](https://github.com/OneSignal/OneSignal-Android-SDK/releases/tag/4.7.1).

For now, I will still leave this issue open in case users cannot upgrade yet and can use the workaround.

PR #18, opened 2016-02-23
Tapping action button now closes notification shade

jkasten2:
When the app is in focus this does not close on its own.

[Review on Reviewable](https://reviewable.io/reviews/one-signal/onesignal-android-sdk/18)

PR #1548, opened 2022-03-15
Fix  Crash when tapping on Notification action button in Android 12

vincent-paing:
# Description
## One Line Summary
Fixes a crash in Android 12 when interacting with action button due to [notification drawer behavior changes](https://developer.android.com/about/versions/12/behavior-changes-all#close-system-dialogs)

### Motivation
This PR addresses the issue #1543 and fixes the crash.

### Scope
This will impact the functionality of dismiss on action press. By default, it will remove this close drawer feature completely since it's no longer supported starting from Android 12+.  However, if the consumer of the lib uses `overrideLibrary` flag to change the lib's target sdk, the behavior will still works up to Android 11. But on Android 12, it will does a silent no-op and log the error.

# Testing
## Unit testing
Not sure if I need to update unit test for this, let me know if it need changes.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [x] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1548)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
@@ -75,10 +75,9 @@ class NotificationOpenedProcessor {
       if (intent.getBooleanExtra("action_button", false)) {
          NotificationManagerCompat.from(context).cancel(intent.getIntExtra(BUNDLE_KEY_ANDROID_NOTIFICATION_ID, 0));
 
-         //Close notification drawer user override targetSdkVersion to  be lower than Android 11
+         // Only close the notifications shade on Android versions where it is allowed, Android 11 and lower.
          // See https://developer.android.com/about/versions/12/behavior-changes-all#close-system-dialogs
-         int targetSdkVersion = context.getApplicationContext().getApplicationInfo().targetSdkVersion;
-         if (targetSdkVersion < Build.VERSION_CODES.S ) {
+         if (android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.S)  {
             context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
          }
       }
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
@@ -74,7 +74,13 @@ class NotificationOpenedProcessor {
       // Pressed an action button, need to clear the notification and close the notification area manually.
       if (intent.getBooleanExtra("action_button", false)) {
          NotificationManagerCompat.from(context).cancel(intent.getIntExtra(BUNDLE_KEY_ANDROID_NOTIFICATION_ID, 0));
-         context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
+
+         //Close notification drawer user override targetSdkVersion to  be lower than Android 11
+         // See https://developer.android.com/about/versions/12/behavior-changes-all#close-system-dialogs
+         int targetSdkVersion = context.getApplicationContext().getApplicationInfo().targetSdkVersion;
+         if (targetSdkVersion < Build.VERSION_CODES.S ) {
+            context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
+         }
       }
    }
 


OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
```diff
@@ -74,7 +74,13 @@ private static void handleDismissFromActionButtonPress(Context context, Intent i
       // Pressed an action button, need to clear the notification and close the notification area manually.
       if (intent.getBooleanExtra("action_button", false)) {
          NotificationManagerCompat.from(context).cancel(intent.getIntExtra(BUNDLE_KEY_ANDROID_NOTIFICATION_ID, 0));
-         context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
+
+         //Close notification drawer user override targetSdkVersion to  be lower than Android 11
+         // See https://developer.android.com/about/versions/12/behavior-changes-all#close-system-dialogs
+         int targetSdkVersion = context.getApplicationContext().getApplicationInfo().targetSdkVersion;
+         if (targetSdkVersion < Build.VERSION_CODES.S ) {
```
jkasten2, line 81: This is checking for the `targetSdkVersion` which only accounts for what was set in the app, not the device the app is running on. To get this you can use `Build.VERSION.SDK_INT`.

OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
```diff
@@ -74,7 +74,13 @@ private static void handleDismissFromActionButtonPress(Context context, Intent i
       // Pressed an action button, need to clear the notification and close the notification area manually.
       if (intent.getBooleanExtra("action_button", false)) {
          NotificationManagerCompat.from(context).cancel(intent.getIntExtra(BUNDLE_KEY_ANDROID_NOTIFICATION_ID, 0));
-         context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
+
+         //Close notification drawer user override targetSdkVersion to  be lower than Android 11
```
jkasten2, line 78: This comment is difficult to understand. I recommend something like this:
> Only close the notifications shade on Android versions where it is allowed, Android 11 and lower.

jkasten2 (changes requested):
Thanks for the PR I left some comments in-line on the code. Also could you include the Android versions you have tested on? Does the Notification Shade still close when clicking on an action button?

OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationOpenedProcessor.java
vincent-paing: I was under the impression  it depends on OS the device is running on at first. But this behavior change depends only on the `targetSdkVersion`. If you target 12 or higher, regardless of whatever OS you're running, it will throw an `SecurityException`. If you target less than 12, Android 12 will do nothing and log an error whereas it will function as normal on previous devices. Since it's not a crash, we can safely run this code without making an additional check for OS version.

vincent-paing: I think it'd be better to an OS check rather than `targetSdkVersion`. I was under the impression that it only depends on `targetSdkVersion`.

vincent-paing:
@jkasten2 I didn't particularly tested with the sdk but I made a sample repo to replicate the behavior. It will close the notification shade on Android 11 and lower. But it won't on Android 12.

jkasten2 (approved):
@vincent-paing Thanks for the fix ups, sorry for the delay. This looks good to me to fix the crash.

Since you noted the behavior on Android 12 is that the shade does not close when tapping on a notification button this still needs to be addressed, however I consider this a much lower priority than crashing and therefore can be followed up in another PR.

@nan-li or @Jeasmine could you take a quick review as well and merge in if you think it looks good too.

nan-li:
Hi @vincent-paing,
Thank you for submitting this PR!
> It will close the notification shade on Android 11 and lower. But it won't on Android 12.

I did some very basic testing with just this SDK's example app, and for the flows I tried, tapping on an action button does close the notification shade for Android 12 as well. It also appeared that `handleDismissFromActionButtonPress` didn't seem needed either. It might be something different from your setup.

Can you share anything about your sample project or what you tried, that had the behavior of not closing the notification shade on Android 12?

https://user-images.githubusercontent.com/4022291/161454112-65d00e97-667d-43af-ab03-3770a011f7b8.mov

vincent-paing:
@nan-li Are you starting an activity on pressing action? In that case, it will close the shade and launch that activity.

vincent-paing:
I made a demo repo where I reproduce the dismiss behavior on pressing an action (https://github.com/vincent-paing/NotificationDemo). 

I believe in your code, pressing an action sends intent to `NotificationOpenedReceiver`, which in turns called `NotificationOpenedProcessor.processFromContext(this, intent)`, which is responsible for executing `handleDismissFromActionButtonPress`. I basically copied the code from `processFromContext` to replicate in my repo.

Here's a video recording (left is Android 11, right is Android 12), you can observe
- It dismiss the notification shade on Android 11
- It DOES NOT dismiss the notification shade on Android 12
- It dismiss on Android 12 only if there's no more notification in the panel. This is because `NotificationManagerCompat.cancel` function. The OS closes the notification shade if you cancel your notification and it's the only one in the shade.


https://user-images.githubusercontent.com/8557241/161624917-fc6c91c8-f51c-48c0-9155-3606be10ca62.mov

jkasten2:
@nan-li @vincent-paing Thanks for testing!

### Accessibility Service Work around
> @jkasten2 As far as my research goes, there's no way to close the system dialog starting from Android 12 unless you run an accessibility service. I don't think it's possible to address a workaround for the OS changes

I agree, we don't wan to use it as it isn't what it is for. [Google's docs state the following](https://developer.android.com/reference/android/accessibilityservice/AccessibilityService):
> Accessibility services should only be used to assist users with disabilities in using Android devices and apps.

### Details on auto close
This automatic close of the notification shade noted here is interesting, however it states "...showing a window that is on top the notification shade" [in Google's docs](https://developer.android.com/about/versions/12/behavior-changes-all#close-system-dialogs-exceptions).
> * Your app targets Android 11 or lower and is showing a window that is on top of the [notification drawer](https://material.io/design/platform-guidance/android-notifications#behavior).
>>Note: If your app targets Android 12, you don't need to use ACTION_CLOSE_SYSTEM_DIALOGS in this situation. That's because, if your app calls [startActivity()](https://developer.android.com/reference/android/content/Context#startActivity(android.content.Intent)) while a window is on top of the notification drawer, the system closes the notification drawer automatically.

However in your testing @nan-li it seems even though the the `startActivity()` isn't showing on top of the notification shade it still seems to close the shade.

vincent-paing:
If it's helpful, In my test if you replace `PendingIntent` with the one for `Activity` instead of `BroadcastReciever`. It always close the shade regardless of whether the app is active or not.

jkasten2:
@vincent-paing Thanks for confirming!
@nan-li This looks the behavior we want, so I think this PR does cover the crash and the consistently issue to me. Approve the PR if you think this covers it as well.

jkasten2:
> If it's helpful, In my test if you replace `PendingIntent` with the one for `Activity` instead of `BroadcastReciever`. It always close the shade regardless of whether the app is active or not.

Ah, this was probably the whole reason for PR #18 added `ACTION_CLOSE_SYSTEM_DIALOGS` in the first place! The SDK at the time did indeed use `BroadcastReciever` and was switched to an `Activity` in the 4.0.0 release. This means we most likely don't need `ACTION_CLOSE_SYSTEM_DIALOGS` for any version of Android. Since it will take extra testing to confirm this on other Android versions I am still ok with merging this PR in to address the crash.

nan-li (approved):
Appreciate the thoroughness @vincent-paing!

Issue #1553, opened 2022-03-23
ClassCastException


CharlieZheng:
**Description:**

We got a crash, and the method stack below
```
Fatal Exception: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
       at com.onesignal.JSONUtils.generateJsonDiff(JSONUtils.java:66)
       at com.onesignal.UserState.generateJsonDiffFromIntoSyncValued(UserState.java:395)
       at com.onesignal.UserStateSynchronizer.updateDeviceInfo(UserStateSynchronizer.java:509)
       at com.onesignal.OneSignalStateSynchronizer.updateDeviceInfo(OneSignalStateSynchronizer.java:222)
       at com.onesignal.OneSignal.registerUserTask(OneSignal.java:1474)
       at com.onesignal.OneSignal.access$1500(OneSignal.java:84)
       at com.onesignal.OneSignal$9.run(OneSignal.java:1441)
       at java.lang.Thread.run(Thread.java:762)
```

**Environment**

1. We using OneSignal v4.6.3
2. we integrate OneSignal by gradle

**Steps to Reproduce Issue:**

we have no idea to reproduce

**Anything else:**

(crash stacktraces, as well as any other information here)

```
Fatal Exception: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Number
       at com.onesignal.JSONUtils.generateJsonDiff(JSONUtils.java:66)
       at com.onesignal.UserState.generateJsonDiffFromIntoSyncValued(UserState.java:395)
       at com.onesignal.UserStateSynchronizer.updateDeviceInfo(UserStateSynchronizer.java:509)
       at com.onesignal.OneSignalStateSynchronizer.updateDeviceInfo(OneSignalStateSynchronizer.java:222)
       at com.onesignal.OneSignal.registerUserTask(OneSignal.java:1474)
       at com.onesignal.OneSignal.access$1500(OneSignal.java:84)
       at com.onesignal.OneSignal$9.run(OneSignal.java:1441)
       at java.lang.Thread.run(Thread.java:762)
```

shubhamvashisht:
@nan-li I don't have access to any samsung device with Android 7.0. we can see thousands of occurrences in the firebase crashlytics for this.

shubhamvashisht:
@nan-li Hi, Any update?
Thanks

nan-li:
[...]

PR #1571, opened 2022-04-20
Before casting a value to `Number`, check that it can

nan-li:
# Description
## One Line Summary
When generating the JSON diff, check that the new value is an instance of `Number` before we cast it to a `Number`.

### Motivation
PR made to fix https://github.com/OneSignal/OneSignal-Android-SDK/issues/1553, which raises a `ClassCastException` when trying to cast a String to a Number. While I can't think of any keys where this is possible for `deviceInfo` (from where this exception is raised), from crash reports it seems that we could have a situation where the new value is a string while the old value is an integer.

### Scope
If the above case happens, it will just put the new string value into the JSON object.

### Context
To decide whether or not to update each value, we check they are not equal with `!value.equals(curValue)`. The additional check for `((Number)curValue).doubleValue() != ((Number)value).doubleValue()` is for cases where the `curValue = 5` and the new `value = 5.0`. They would pass the first check but we don't need an update since the JSON serializer would turn `5.0` into `5` anyway, so `5.0` is not considered different from `5`.

Made a modification to check the current value is an instance of `Number` instead of `Integer` since it could be that it is a float or double.

# Testing
## Unit testing
- Existing unit tests pass
- This is a minor fix, so doesn't seem to warrant new unit tests.

## Manual testing
- app build with Android Studio 2020.3.1 with a fresh install of the OneSignal example app on a Pixel 5 with Android 11.
- hard coded some `deviceInfo` where the old value is an integer and the new value is a string for testing - got the same exception prior to fix

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1571)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/JSONUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/JSONUtils.java
@@ -62,7 +62,7 @@ class JSONUtils {
                         Object curValue = cur.get(key);
                         if (!value.equals(curValue)) {
                             // Work around for JSON serializer turning doubles/floats into ints since it drops ending 0's
-                            if (curValue instanceof Integer && !"".equals(value)) {
+                            if (curValue instanceof Number && value instanceof Number) {
                                 if ( ((Number)curValue).doubleValue() != ((Number)value).doubleValue())
                                     output.put(key, value);
                             }


jkasten2 (changes requested):
I think we can simply this code with your new logic, more details on the line of code itself.

> This is a minor fix, so doesn't seem to warrant new unit tests.

If we had existing test infrastructure for `generateJsonDiff` this would be a good to add a test for. However the scope of this PR is a minor so it doesn't justify the time it would take to create the testing infrastructure to create an automated test for this at this time.

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1571#-:-N04vcyl3gGVll_3Tn_H:b-nmhj3f)* status: all files reviewed, 1 unresolved discussion (waiting on @nan-li)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/JSONUtils.java, line 65 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1571#-N04uyb7EE5qOimbcF29:-N04uyb7EE5qOimbcF2A:b85wbta) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/55228bacd392f7e9d7daf5b60c0b31d29ccf315b/OneSignalSDK/onesignal/src/main/java/com/onesignal/JSONUtils.java#L65)):*
> ```Java
>                         if (!value.equals(curValue)) {
>                             // Work around for JSON serializer turning doubles/floats into ints since it drops ending 0's
>                             if (curValue instanceof Integer && !"".equals(value) && value instanceof Number) {
> ```

Based on the existing comment and the new `value instanceof Number` check you added I think we can drop the `!"".equals(value)` check. As "" would NOT be an instance of `null` anyway.

jkasten2 (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1571)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nan-li)

___
*[OneSignalSDK/onesignal/src/main/java/com/onesignal/JSONUtils.java, line 65 at r1](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1571#-N04uyb7EE5qOimbcF29:-N07S00u5NUDdtU_6q6z:by3snmp) ([raw file](https://github.com/onesignal/onesignal-android-sdk/blob/55228bacd392f7e9d7daf5b60c0b31d29ccf315b/OneSignalSDK/onesignal/src/main/java/com/onesignal/JSONUtils.java#L65)):*

Based on the existing comment and the new `value instanceof Number` check you added I think we can drop the `!"".equals(value)` check. As "" would NOT be an instance of `null` anyway.
</blockquote></details>

Ah nice! Checking if both are an `instanceof` makes the most sense, since either one could be the int or a floating type.

nan-li:
Test `shouldSetExpireTimeCorrectlyWhenMissingFromPayload` failed, re-running

nan-li (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1571)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nan-li)

PR #1581, opened 2022-05-05
[Change] Notification Activity Trampoline forward instead of reverse

jkasten2:
# Description
## One Line Summary
Reverts [Reverse Activity Trampoline](https://github.com/OneSignal/OneSignal-Android-SDK/pull/1377) back to standard Activity Trampoline, which executes on notification open.

### Motivation
It was discovered after Reverse Activity Trampoline was released in [version 4.5.0](https://github.com/OneSignal/OneSignal-Android-SDK/releases/tag/4.5.0) that it was not compatible with Xiaomi devices and with HMS message type notifications.

For HMS we already went back to using an Activity Trampoline in PR #1533, this was required since the notification is generated by HMS core we don't have the level of control needed for it.
For Xiaomi devices [Reverse Activity Trampoline does not work](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1494) due to [a bug](https://github.com/OneSignal/NotificationActivityBackstackExample) where the back stack is not persevered on notification open.

Reverse Activity Trampoline has limitations on where it can be used. Activity Trampoline is the only option which allows us to have the same implementation for all our features and platforms we support. All the other alternative options (listed below) need to use Activity Trampolining somewhere anyway to fill in gaps to track clicks for all types of notifications, devices, and push vendors.

### Scope
Reverting PR #1377 (Reverse Activity Trampoline) behavior while keeping any intended fixes and refactoring from it.

#### Option 1 - Detect by MIUI version, do Normal Activity Trampling
This might not be possible, even if it is it through this would fragile and extra part to maintain a list of MIUI versions and possibly other manufactures.

#### Option 2 - Detect missing back stack entry on open
Detect if there is a back stack entry when the notification is opened with ActivityManager.getAppTasks(). If there isn't an entry we will need to check something Intent to know if that was really what was intended. If not we need re-parse some of the payload and open the desired Activity like we do for Huawei message notification types.

#### Option 3 - Direct Activity, use ActivityLifecycleCallbacks to get meta data
Opening the Activity directly and the SDK still being able to get the data is possible if we setup the `ActivityLifecycleCallbacks` on or before `Application.onCreate` so we don't mis any events.
The problem with this is if the target was another app (such as opening a URL in Chrome) this approach would not be able to capture the click.
   * For this specific case we could still use Activity Trampling, however this would be a different implementation so don't think it is worth it.
[...]

PR #1583, opened 2022-05-08
[Fix] Network Delayed Activity Trampoline

jkasten2:
# Description
## One Line Summary
Removes a network call dependency on notification open logic which effects foregrounding the app.

### Motivation
* After the changes in PR #1581 this becomes critical this is resolved so the app still foregrounds when tapping on a notification while offline.
* This also resolves a Xiaomi compatibility issue where if the app is swiped away then a notification is opened it won't foreground either. _unless something starts the app process in the background later and is still running when you open the notification oddly enough_

### Scope
Target scope is the fix the bug noted above.
A secondary fix was made to improve the app shutdown process to stop retrying to get remote params when its thread is interrupted. _This fix was only required in this PR to resolve a test carry over issue with the test added in this PR._

### Background
The fixes in this PR fixes a discovered issue in 4.0.0 to 4.4.x, but is not an issue in 4.5.0 to 4.7.2. However this fix needs to be shipped with #1581 as it reverts most of the changes in #1377 that was shipped in 4.5.0.

## Xiaomi - Background Restrictions
The following was observed through testing on a Xiaomi Redmi 6A running Android 9 on MIUI Global 11.0.8 | Stable 11.0.8.0 (PCBMIXM):
* Activity Trampling - Must keep an Activity alive while trampolining
   -  If the app was swiped away AND the app process is not currently running, then a notification is tapped on that has a `PendingIntent` to start an `Activity` and then you trampoline to another `Activity` you MUST call `startActivity` BEFORE finishing the first `Activity`. Otherwise the `startActivity` is ignored.

# Testing
## Unit testing
* A new test was added.
## Manual testing
Tested on:
* Android 12 - Samsung S21 - OneUI 4.1
* Android 9 - Xiaomi Redmi 6A - MIUI 11

Test both offline and online.

# Affected code checklist
   - [X] Notifications
      - [ ] Display
      - [X] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [X] REST API requests
      - [X] Remote Parameters
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1583)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -2395,22 +2395,6 @@ public class OneSignal {
     * Method called when opening a notification
     */
    static void handleNotificationOpen(final Activity context, final JSONArray data, @Nullable final String notificationId) {
-      // Delay call until remote params are set
-      if (taskRemoteController.shouldQueueTaskForInit(OSTaskRemoteController.HANDLE_NOTIFICATION_OPEN)) {
-         logger.error("Waiting for remote params. " +
-                 "Moving " + OSTaskRemoteController.HANDLE_NOTIFICATION_OPEN + " operation to a pending queue.");
-         taskRemoteController.addTaskToQueue(new Runnable() {
-            @Override
-            public void run() {
-               if (appContext != null) {
-                  logger.debug("Running " + OSTaskRemoteController.HANDLE_NOTIFICATION_OPEN + " operation from pending queue.");
-                  handleNotificationOpen(context, data, notificationId);
-               }
-            }
-         });
-         return;
-      }
-
       // If applicable, check if the user provided privacy consent
       if (shouldLogUserPrivacyConsentErrorMessageForMethodName(null))
          return;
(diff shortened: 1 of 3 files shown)


nan-li (approved):
I can understand the phone offline -> tap notification fix.

I can't comprehend the Xiaomi part of the fix. Is it something like `handleNotificationOpen` is added to the task queue, but it's removed before it gets to run?

jkasten2:
> I can't comprehend the Xiaomi part of the fix. Is it something like `handleNotificationOpen` is added to the task queue, but it's removed before it gets to run?

I added a new section " Xiaomi - Background Restrictions" above to explain the Xiaomi limitation in more detail.

nan-li:
>  `startActivity` is called AFTER `finish()` is called in NotificationOpenedReceiverBase.kt

Ahhhh got it, thanks!

jkasten2:
This branch is based on `change/notification_activity_trampoline_forward_instead_of_reverse`, however it was rebased to `--autosquash` a `--fixup` commit. That rebase caused this branch to become out of date, with github showing a merge conflict and commits from both branches showing as the diff. To fix this the following was run:
```
git checkout fix/delayed_activity_trampoline
git rebase --onto change/notification_activity_trampoline_forward_instead_of_reverse origin/change/notification_activity_trampoline_forward_instead_of_reverse@{1}
```

`origin` and `@{1}` are important here as they are the previous known git reflog on github before it was rebased.

PR #1597, opened 2022-06-01
Fix issue where a notification's android ID is not available to OSRemoteNotificationReceivedHandler

brismithers:
# Description
## One Line Summary
Generate a notification's Android ID earlier in the notification received flow to ensure it is available for `OSRemoteNotificationReceivedHandler.remoteNotificationReceived`.

### Motivation
When an `OSRemoteNotificationReceivedHandler` has been [set up](https://documentation.onesignal.com/docs/service-extensions#android-notification-service-extension)  it will receive an `OSNotification` with an `androidNotificationId` that *may* be zero.  When control is given to this handler the expectation is that the `androidNotificationId` will have always been set.

### Scope
It is expected that all scenarios driving `OSRemoteNotificationReceivedHandler.remoteNotificationReceived` to be called will now provide an `OSNotification` (via `OSNotificationReceivedEvent.getNotification()`) that has `androidNotificationId` set to a non-zero value.

### OPTIONAL - Other
Existing logic in `NotificationBundleProcessor` depends on knowing whether the android ID was generated or was previously set (see calls to `OSNotificationGenerationJob.setAndroidIdWithoutOverriding and getAndroidIdWithoutCreate`).  Logic to remember whether an android ID was generated needed to be added as part of the moving of the ID generation logic. 

# Testing
## Unit testing
No unit tests were added for this fix.

## Manual testing
OneSignal Demo App built with Android Studio 2021.2.1 running on Emulator Android v10 and v11 testing the following scenarios:
* Push notification - dismissed
* Push notification - opened
* Restoring push notification after app restart
* Collapsing push notification
* Summary push notification
* Silent push notification

# Affected code checklist
   - [x] Notifications
      - [x] Display
      - [x] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1597)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
```diff
@@ -133,6 +131,21 @@ public OSNotification getNotification() {
     }
 
     public void setNotification(OSNotification notification) {
+        // we determine whether an android notification id has been
+        // generated by us only for the first notification. The assumption
+        // is any subsequent notification updates are derived from the
+        // initial one, therefore the didGenerateId will persist across.
+        if(this.notification == null)
```
jkasten2, line 138: I see where we have a `notDisplayNotificationLogic` method that I don't think will support this logic:
https://github.com/OneSignal/OneSignal-Android-SDK/blob/4.7.3/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationController.java#L108-L109

A suggestion is create some kind of didDisplay flag which might make this simpler. Either way we probably need to update all points accessing this class and account for display / not display logic.

brismithers: @jkasten2 I reworked the fix to include a `isNotificationToDisplay` flag, and adjusted everyone looking for a -1 Android ID to now look for that flag.

jkasten2 (approved):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSNotificationGenerationJob.java
jkasten2: Thanks, looks good to me.

PR #1598, opened 2022-06-02
[Change] After opening a notification with a URL deeplink, the back button will return to the previous Activity

jkasten2:
# Description
## One Line Summary
After opening a notification with a URL deeplink, the back button will return to the previous Activity. Also allow URLs that open the browser to keep task in the task list.

### Motivation
It is a better default user experience if the User taps on a notification that deeplinks into that app and it takes them back to the last screen they were on after pressing back. Also with the removal of the `FLAG_ACTIVITY_NO_HISTORY` the user can resume the app or site they if they background it.

### Scope
Only effects notification opens with URLs.

# Testing
## Unit testing
No existing Unit Test to update. Specific Intent flags are not fully implemented in Robolectric.

## Manual testing
Tested on Android 12 and Android 6 emulators. Tested both a deep link URL and a URL that opens the browser. Tested wit the app backgrounded, swiped away, and kicked out of memory. Pressed the back button in each of these cases to ensure the behavior was correct.

# Affected code checklist
   - [X] Notifications
      - [ ] Display
      - [X] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1598)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
@@ -579,10 +579,7 @@ class OSUtils {
             break;
       }
       intent.addFlags(
-          Intent.FLAG_ACTIVITY_NO_HISTORY |
-          Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET |
-          Intent.FLAG_ACTIVITY_MULTIPLE_TASK |
-          Intent.FLAG_ACTIVITY_NEW_TASK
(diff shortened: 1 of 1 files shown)


nan-li (approved):
Makes sense! LGTM

PR #1633, opened 2022-08-25
[User Model] Migrate Purchase Tracking

brismithers:
* Adjust IStartableService so it is called *after* the appId has been established. This is done to remove the extra step of having startable services register a listener to wait for the appId to be set.
* Create RefreshParamsService as a startable service to pull the remote parameters from the backend. This was previously part of StartupService, separated as a refactor.
* Migrate TrackAmazonPurchase and TrackGooglePurchase which are startable services that track amazon and google purchases (respectively).  When captured these get pushed to the operation repo via TrackPurchaseOperation.  The execution of that operation to-be-implemented.

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-android-sdk/1633)


OneSignalSDK/onesignal/src/main/java/com/onesignal/onesignal/core/internal/OneSignalImp.kt
```diff
@@ -123,7 +121,17 @@ class OneSignalImp() : IOneSignal, IServiceProvider {
             _appId = appId
```
brismithers, line 121: remove setAppId altogether and make parameter to initwithcontext

Issue #1672, opened 2022-10-19
WorkManager is not initialized properly


nan-li:
# Description

We have seen many reported crashes related to this particular WorkManager IllegalStateException: 

> WorkManager is not initialized properly. You have explicitly disabled WorkManagerInitializer in your manifest, have not manually called WorkManager#initialize at this point, and your Application does not implement Configuration.Provider.

However, the OneSignal SDK uses the default `WorkManagerInitializer` and does **not** disable it. We are collecting information to investigate any solutions on our end as well as make a report of this to Google, as this seems to be related to an AndroidX library that OneSignal uses. For reference, these are reports of this and related exceptions to [Google's Issue Tracker](https://issuetracker.google.com/issues?q=%22WorkManager%20is%20not%20initialized%20properly%22).

# Current status
## Google Issue Tracker
We reported this to Google's WorkManager issues [here](https://issuetracker.google.com/issues/258176803).

## Summary

From reported stacktraces, the crash arises from `OSNotificationRestoreWorkManager.beginEnqueueingWork()`, `OSFocusHandler.cancelOnLostFocusWorker()` , and `OSFocusHandler.startOnLostFocusWorker()`.

We know this can be related to multiprocess, but we have also seen many reports of this exception in applications that **don't** use multiprocess. 

This is happening on applications that don't use WorkManager themselves and don't use [Custom WorkManager Configuration and Initialization](https://developer.android.com/topic/libraries/architecture/workmanager/advanced/custom-configuration).

OneSignal maintainers have not been able to reproduce this crash, and users have not reported being able to reproduce this crash either. While there have been reports of some apps encountering this crash 100% of the time, those cases are related to other underlying issues.

This crash is also happening across multiple versions of `androidx.work:work-runtime` including on the latest stable release `2.7.1`. It also appears to affect devices across types, brands, and Android versions.

We speculate it may have to do with the initialization of libraries and components during an app's startup process, and encountering edge cases where OneSignal tries to access the `WorkManager` singleton but it has not yet been initialized by the system. This crash happens when the OneSignal SDK calls `WorkManager.getInstance(Context)` and the `WorkManager` is `null`.

[...]

PR #1721, opened 2023-01-17
Speculative Fix for WorkManager not Initialized Crash

nan-li:
# Description
## One Line Summary
This is a speculative fix for the non-reproducible WorkManager is not Initialized Crash we have been seeing many reports of.

### Motivation
Please see issue this aims to address: https://github.com/OneSignal/OneSignal-Android-SDK/issues/1672.

We don't completely understand this crash, so this fix is speculative and based off a [response from Google Issue Tracker](https://issuetracker.google.com/issues/258176803#comment7), though the responder may not completely understand how OneSignal is used in applications. 

### Scope
* Before using WorkManager, we will check for its existence. Else, in rare cases that were crashing, initialize it ourselves.
* We use a method to check if WorkManager is initialized in the process.
  - This is effectively the `WorkManager.isInitialized()` public method introduced in `androidx.work:work-*:2.8.0-alpha02`.
  - Please see https://android-review.googlesource.com/c/platform/frameworks/support/+/1941186 for the library's implementation

### ⚠️ Concerns ⚠️
We take over and initialize WorkManager if we see it is `nil`, which so far is rare.

Perhaps this crash happens because the default WorkManager initializer hasn't run yet, but is about to... In which case, right after we initialize WorkManager, the default initializer will trigger a `WorkManager is already initialized` exception. But, maybe that's not what is happening and this secondary crash doesn't happen.

Before accessing and initializing WorkManager ourselves, we check if it's already initialized using the same logic that the `work` library provides in a `2.8.0-alpha02` release. However, could we have a race condition such that when we check, it is `nil`, but momentarily after, it becomes initialized, while we go and try to initialize it? This could create issues that didn't exist before. Dig into source code and their `synchronized` blocks.

# Testing
## Unit testing
No unit test was added as it is hard to unit test this. 
- we would have to test double Android library components
- `androidx.work.testing` package didn't have anything for our use case either

## Manual testing
Limited ability to test manually.
- Tested normal app flow in an emulator

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1721)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSFocusHandler.kt
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSFocusHandler.kt
@@ -32,7 +32,6 @@ import androidx.work.Constraints
 import androidx.work.ExistingWorkPolicy
 import androidx.work.NetworkType
 import androidx.work.OneTimeWorkRequest
-import androidx.work.WorkManager
 import androidx.work.Worker
 import androidx.work.WorkerParameters
 import java.util.concurrent.TimeUnit
@@ -87,7 +86,8 @@ class OSFocusHandler {
             .setInitialDelay(delay, TimeUnit.MILLISECONDS)
             .addTag(tag)
             .build()
-        WorkManager.getInstance(context)
+
+        OSWorkManagerHelper.getInstance(context)
             .enqueueUniqueWork(
                 tag,
                 ExistingWorkPolicy.KEEP,
@@ -96,7 +96,7 @@ class OSFocusHandler {
     }
 
     fun cancelOnLostFocusWorker(tag: String, context: Context) {
-        WorkManager.getInstance(context).cancelAllWorkByTag(tag)
+        OSWorkManagerHelper.getInstance(context).cancelAllWorkByTag(tag)
     }
 
     private fun resetStopState() {
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSWorkManagerHelper.kt
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSWorkManagerHelper.kt
@@ -1,0 +1,68 @@
+/**
+ * Modified MIT License
+ * <p>
+ * Copyright 2023 OneSignal
+ * <p>
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * <p>
+ * 1. The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ * <p>
+ * 2. All copies of substantial portions of the Software may only be used in connection
+ * with services provided by OneSignal.
+ * <p>
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.onesignal
+
+import android.annotation.SuppressLint
+import android.content.Context
+import androidx.work.Configuration
+import androidx.work.WorkManager
+import androidx.work.impl.WorkManagerImpl
+
+object OSWorkManagerHelper {
+    /**
+     * Helper method to provide a way to check if WorkManager is initialized in this process.
+     *
+     * This is effectively the `WorkManager.isInitialized()` public method introduced in androidx.work:work-*:2.8.0-alpha02.
+     * Please see https://android-review.googlesource.com/c/platform/frameworks/support/+/1941186.
+     *
+     * @return `true` if WorkManager has been initialized in this process.
+     */
+    @SuppressWarnings("deprecation")
+    @SuppressLint("RestrictedApi")
+    private fun isInitialized(): Boolean {
+        val instance = WorkManagerImpl.getInstance()
+        return instance != null
+    }
+
+    /**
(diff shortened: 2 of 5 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSFocusHandler.kt
```diff
@@ -87,16 +88,27 @@ class OSFocusHandler {
...
             .build()
-        WorkManager.getInstance(context)
-            .enqueueUniqueWork(
-                tag,
-                ExistingWorkPolicy.KEEP,
-                workRequest
-            )
+
+        // Before using WorkManager, check for its existence. Else, initialize it ourselves.
+        if (OSUtils.isWorkManagerInitialized()) {
```
jkasten2, line 93: Recommend making a helper for this if else logic so it doesn't have to be repeated.

Could be something like `WorkerMangerHelper.getInstance(context)` where it contains this logic.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
```diff
@@ -667,4 +669,19 @@ static Throwable getRootCauseThrowable(@NonNull Throwable subjectThrowable) {
...
+    * Provides a way to check if WorkManager is initialized in this process.
+    *
+    * This is effectively the WorkManager.isInitialized() public method introduced in androidx.work:work-*:2.8.0-alpha02.
+    * Please see https://android-review.googlesource.com/c/platform/frameworks/support/+/1941186.
+    *
+    * @return {@code true} if WorkManager has been initialized in this process.
+    */
+   @SuppressWarnings("deprecation")
+   @SuppressLint("RestrictedApi")
+   static boolean isWorkManagerInitialized() {
```
jkasten2, line 683: Would be cleaner to put this into a new WorkManagerHelper class instead of adding to this OSUtil misc class.

jkasten2 (changes requested):

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSFocusHandler.kt
nan-li: Done

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
nan-li: Done

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSWorkManagerHelper.kt
```diff
@@ -0,0 +1,68 @@
...
+     * This is effectively the `WorkManager.isInitialized()` public method introduced in androidx.work:work-*:2.8.0-alpha02.
+     * Please see https://android-review.googlesource.com/c/platform/frameworks/support/+/1941186.
+     *
+     * @return `true` if WorkManager has been initialized in this process.
+     */
+    @SuppressWarnings("deprecation")
+    @SuppressLint("RestrictedApi")
+    private fun isInitialized(): Boolean {
+        val instance = WorkManagerImpl.getInstance()
+        return instance != null
```
jkasten2, line 49: Since the logic here is very simple and `instance` is only used once it would be simpler code if we just make this a one liner.
```kotlin
return WorkManagerImpl.getInstance() != null
```

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSWorkManagerHelper.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * If there is an instance of WorkManager available, use it. Else, in rare cases, initialize it ourselves.
+     *
+     * Calling `WorkManager.getInstance(context)` directly can cause an exception if it is null.
+     *
+     * @return an instance of WorkManager
+     */
+    @JvmStatic
+    fun getInstance(context: Context): WorkManager {
```
jkasten2, line 60: It's possible this could get calls from multiple threads at the same time. To handle this we can use  `@Synchronized` here.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSWorkManagerHelper.kt
```diff
@@ -0,0 +1,68 @@
...
+        return if (isInitialized()) {
```
jkasten2, line 61: I think it would simplify the code a bit written like this so we don't repeat `WorkManager.getInstance(context)`.
```kotlin
if (!isInitialized()) {
   WorkManager.initialize(context, Configuration.Builder().build())
}

return WorkManager.getInstance(context);

```

jkasten2 (approved):
I think this PR is safe to put into main and a normal release.

This won't introduce any new crashes, the concern of "WorkManager is already initialized" wouldn't happen with the default work manager initialization. This is because they use a Provider which always runs before `Application.onCreate` [per Android's docs](https://developer.android.com/reference/android/content/ContentProvider#onCreate()) and what we have observed in our own testing. However, we have seen a report of [one expectation](https://github.com/OneSignal/OneSignal-Unity-SDK/issues/507#issuecomment-1336822621) to this where it can be completely skipped and never runs the Provider. This PR should fix this, and hopefully is this is the same root issue for the other crash reports.

If the app is initializing WorkManager themselves there is a very small possibility "WorkManager is already initialized" could happen instead of "WorkManager not Initialized". However this could only happen if there was a pre-existing race condition with OneSignal and the app's custom initializer anyway.

nan-li:
Consider if we will override an app's custom initializer if there is one.

jkasten2:
> Consider if we will override an app's custom initializer if there is one.

I believe our `isInitialized()` would detect that, then we would skip `WorkManager.initialize` so we won't overwrite theirs.

nan-li:
I was thinking about `WorkManager.getInstance(context)` ([this code here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:work/work-runtime/src/main/java/androidx/work/impl/WorkManagerImpl.java;l=165)). If it returned `null` and there is a custom initializer, the code would have created a WorkManager with that configuration.

jkasten2:
True, but if they didn't run their custom initializer soon enough today they would be seeing "WorkManager not Initialized" being thrown by the OneSIgnal SDK today.

I think this is a compatibility issue we should try to improve on in it's own PR, but we need to come up with a strategy and / or an API first.

OneSignalSDK/onesignal/src/main/java/com/onesignal/OSWorkManagerHelper.kt
nan-li: ah ok, Android library split up these two lines. Resolved.

jkasten2 (approved):

Martin-Gonzalez90:
Any ETA for this merge?

jennantilla (approved):
Interesting potential for conflict with a custom initializer not running soon enough. Very interested in observing future strategizing on this compatibility issue.

PR #1750, opened 2023-03-20
[User Model] Don't show "Missing Google Project number!" error message until remote config loaded.

brismithers:
# Description
## One Line Summary
Don't show "Missing Google Project number!" error message until remote config loaded.

### Motivation
Retrieving the FCM push token occurs on initialization. The first time the SDK is initialized this *might* happen before the remote config has been pulled down, causing error message "Missing Google Project number!" to be logged.  Once the configuration has been pulled down the FCM push token is again attempted to be retrieved and (assuming no other issues) will succeed, so no long term underlying issue.  Rather than attempt to synchronize token retrieval to always be after config retrieval, we continue to allow this as a possibility but fail early on token retrieval when there is not yet a config.

### Scope
This change affects when an error message is logged due to a bad google project number.  We no longer consider "Has not yet retrieved the google project number" as reason to log an error. 

# Testing
## Manual testing
Using the example app, tested SDK initialization on a new and existing app and manually inspected state/logs to determine correct behavior.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
@@ -51,6 +51,10 @@ internal abstract class PushRegistratorAbstractGoogle(
     abstract suspend fun getToken(senderId: String): String
 
     override suspend fun registerForPush(): IPushRegistrator.RegisterResult {
+        if(!_configModelStore.model.isInitializedWithRemote) {
+            return IPushRegistrator.RegisterResult(null, SubscriptionStatus.FIREBASE_FCM_INIT_ERROR)
+        }
+
         if (!_deviceService.hasFCMLibrary) {
             Logging.fatal("The Firebase FCM library is missing! Please make sure to include it in your project.")
             return IPushRegistrator.RegisterResult(null, SubscriptionStatus.MISSING_FIREBASE_FCM_LIBRARY)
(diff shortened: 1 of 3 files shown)


jennantilla:
This is a really helpful update! I know on 4.x.x a similar log shows when evaluating whether the sender ID is a valid format. I didn't notice anywhere we were making the remote config check (but might've missed it?). If not, are there any plans to add this to the current SDK, or just for 5.0.0 onwards?

brismithers:
@jennantilla was only planning on making this for 5.0.0 onwards, but we can look into lower versions too

nan-li (approved):
LGTM

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
```diff
@@ -51,6 +51,10 @@ internal abstract class PushRegistratorAbstractGoogle(
     abstract suspend fun getToken(senderId: String): String
 
     override suspend fun registerForPush(): IPushRegistrator.RegisterResult {
+        if(!_configModelStore.model.isInitializedWithRemote) {
```
jkasten2, line 54: nit, missing space, should be `if (`

jkasten2 (approved):

PR #1775, opened 2023-05-10
Prevent ANRs when backgrounding app

jkasten2:
# Description
## One Line Summary
Prevent ANRs when app is backgrounded due to some disk I/O with session logic and AndroidX Worker usage.

### Motivation
The time spent on the main thread in this scenario is significantly contributing to ANRs. The Google Play store penalizes apps with ANRs with lower search rankings and possibly some other negatives.  

### Scope
Simply move code to a new background thread. No logic changes. Longer term in the 5.0.0 release we can use a thread pool or Kotlin suspend instead.

### Full Details
Move calls to methods that do disk I/O from the main thread to a new background thread.

There are reports where this can cause a significant amount of time on main thread. Any disk I/O is a risk on the main thread and should not be done as there is no upper bond of how long it can take. ANRs trigger after 5 seconds of the main thread being busy.

The stacktrace of reported ANRs confirm that appStopped then calls getSessionInfluences which can result in waiting on disk I/O. We have also seen stacktraces with scheduling a AndroidX worker so that was moved into this background thread as well. I isn't known if it does any disk I/O but it was significant enough to show up in some ANRs.

# Testing
## Unit testing
No unit testing changes.

## Manual testing
Confirmed code still fires when the app is backgrounded on an Android 13 emulator.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [X] Outcomes
   - [X] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1775)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/ActivityLifecycleHandler.java
@@ -179,8 +179,14 @@ class ActivityLifecycleHandler implements OSSystemConditionController.OSSystemCo
         if (focusHandler == null || focusHandler.hasBackgrounded() && !focusHandler.hasCompleted())
             return;
 
-        OneSignal.getFocusTimeController().appStopped();
-        focusHandler.startOnLostFocusWorker(FOCUS_LOST_WORKER_TAG, SYNC_AFTER_BG_DELAY_MS, OneSignal.appContext);
+        new Thread() {
+            public void run() {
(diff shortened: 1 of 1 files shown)


emawby (approved):
LGTM!

PR #1799, opened 2023-08-09
[5.0.0] Fix IAM display & dismiss bugs

jennantilla:
# Description
## One Line Summary
This PR addresses two In App Message bugs:

1. IAM with a custom trigger not displaying after dismissing previous IAM.
`InAppMessagesManager.attemptToShowInAppMessage: There is an IAM currently showing!`
2. `messageWasDismissed` returning early if `!_lifecycleCallback.hasSubscribers`

### Motivation
**Issue 1** 
Issue is reproducible following these steps:

1. Create IAM_1 that is triggered On App Open
2. Create IAM_2 that has a custom trigger
3. Run app
4. IAM_1 displays
5. Dismiss IAM_1
6. Trigger IAM_2 with `addTrigger` method
7. IAM_2 does not display, console logs `InAppMessagesManager.attemptToShowInAppMessage: There is an IAM currently showing!`

setting `inAppMessageIdShowing` back to `null` in `attemptToShowInAppMessage` allows IAM_2 to successfully display.

**Issue 2**
This issue was caught during a pair programming session where the early return would interfere with `_lifecycleCallback.fireOnMain` being called.

### Scope
Changes affect displaying and dismissing IAMs, specifically `attemptToShowInAppMessage` & `messageWasDismissed` in InAppMessagesManager.

### OPTIONAL - Other

# Testing
## Manual testing
Tested Issue 1 to ensure that second IAM  successfully displays.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1799)

### DIFF
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
@@ -393,11 +393,9 @@ internal class InAppMessagesManager(
         }
 
         // fire the external callback
-        if (!_lifecycleCallback.hasSubscribers) {
-            Logging.verbose("InAppMessagesManager.messageWasDismissed: inAppMessageLifecycleHandler is null")
-            return
+        if (_lifecycleCallback.hasSubscribers) {
+            _lifecycleCallback.fireOnMain { it.onDidDismiss(InAppMessageLifecycleEvent(message)) }
         }
-        _lifecycleCallback.fireOnMain { it.onDidDismiss(InAppMessageLifecycleEvent(message)) }
 
         _state.inAppMessageIdShowing = null
 


OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -394,10 +395,11 @@ internal class InAppMessagesManager(
 
         // fire the external callback
         if (!_lifecycleCallback.hasSubscribers) {
-            Logging.verbose("InAppMessagesManager.messageWasDismissed: inAppMessageLifecycleHandler is null")
+            _lifecycleCallback.fireOnMain { it.onDidDismiss(InAppMessageLifecycleEvent(message)) }
```
emawby, line 398: This seems backwards we are firing the event if we don't have subscribers. We should be firing if we do have subscribers and we shouldn't be returning in the else case

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -350,6 +350,7 @@ internal class InAppMessagesManager(
         // If there was a message dequeued, display it
         if (messageToDisplay != null) {
             var result = _displayer.displayMessage(messageToDisplay!!)
+            _state.inAppMessageIdShowing = null
```
emawby, line 353: If result isn't null or false that means that we are displaying a message correct? In that case setting the id to null seems wrong.

emawby (changes requested):

emawby (approved):

PR #1842, opened 2023-09-18
Fix timezone_id not being set

### LINKED ISSUE #1829: [Bug]: Timed Push Notifications are not received at the correct hour
### What happened?

Since upgrading my apps to OneSignal Flutter 5.0.0, my messages that are supposed to be delivered in each user's timezone at 7pm are being delivered hours early. My Android device in Central Time is receiving the notification at 2pm instead of 7pm.

### Steps to reproduce?

```Markdown
1. Send a timed notification for 7pm in the user's timezone.
```

### What did you expect to happen?

Notification should be received at 7pm. Instead it is received at 2pm.

### OneSignal Android SDK version

5.0.0

### Android version

12

### Specific Android models

```Markdown
All
```

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

emawby:
# Description
## One Line Summary
This fixes an issue where timezone_id was not being set as a property on users.

## Details
This issue resulted in scheduled notifications being received at the wrong time.
In order to send the `timezone_id` we are adding a properties map in the `createUser` call. The properties set there (only `timezone_id`) will be hydrated to the properties model when the response is received.
fixes #1829 

### Motivation
Fix scheduled notifications

### Scope
user properties during createUser

# Testing
## Unit testing
Added properties object to existing unit tests

## Manual testing
Tested scheduled notifications on a physical device.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1842)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
@@ -1,11 +1,7 @@
 package com.onesignal.user.internal.operations.impl.executors
 
 import android.os.Build
-import com.onesignal.common.AndroidUtils
-import com.onesignal.common.DeviceUtils
-import com.onesignal.common.NetworkUtils
-import com.onesignal.common.OneSignalUtils
-import com.onesignal.common.RootToolsInternalMethods
+import com.onesignal.common.*
 import com.onesignal.common.exceptions.BackendException
 import com.onesignal.common.modeling.ModelChangeTags
 import com.onesignal.core.internal.application.IApplicationService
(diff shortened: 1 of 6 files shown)


nan-li (approved):
LGTM, noting we don't save the timezone_id to the properties model locally until the next session via hydration from a fetch user.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
```diff
@@ -99,6 +93,8 @@ internal class LoginUserOperationExecutor(
     private suspend fun createUser(createUserOperation: LoginUserOperation, operations: List<Operation>): ExecutionResponse {
         var identities = mapOf<String, String>()
         var subscriptions = mapOf<String, SubscriptionObject>()
+        val properties = mutableMapOf<String, String>()
+        properties["timezone_id"] = TimeUtils.getTimeZoneId()!!
```
nan-li, line 97: I would prefer to remove `!!` and change `TimeUtils.getTimeZoneId()` to return a non-optional String. That method seems like it should always return a String.

nan-li (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
emawby: Fixed! Ya that method always returns a string so I changed its return type to be non-optional

nan-li (approved):

PR #1846, opened 2023-09-21
Fix In-App Message auto-dismiss crash

jennantilla:
# Description
## One Line Summary
Fix In-App Message auto-dismiss crash.

## Details
Manually closing an auto-dismiss In-App Message before the timer runs out resulted in a NullPointerException crash. Replacing `messageView` non-null assertion `!!` with safe call operator `?` prevents crash.

### Motivation
Prevent NullPointerException crash when auto-dismiss IAM is manually closed.

### Scope
Non-null assertion `!!` replacement with safe call operator `?` only applied to the `messageView` variable within `showMessageView` method. Similar safe call operator replacements may be made to other code blocks in the future.

# Testing
## Manual testing
Tested manually dismissing the auto-dismiss IAM before the timer runs out on Pixel 6 emulator running Android 13 as well as Pixel 7 emulator running Android 14.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/WebViewManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/WebViewManager.kt
@@ -279,13 +279,14 @@ internal class WebViewManager(
             }
             Logging.debug("In app message, showing first one with height: $newHeight")
 
-            messageView!!.setWebView(webView!!)
+            messageView?.setWebView(webView!!)
             if (newHeight != null) {
                 lastPageHeight = newHeight
-                messageView!!.updateHeight(newHeight)
+                messageView?.updateHeight(newHeight)
             }
-            messageView!!.showView(activity)
-            messageView!!.checkIfShouldDismiss()
+            messageView?.showView(activity)
+            // Executed in the same thread
+            messageView?.checkIfShouldDismiss()
         }
     }
 


OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/WebViewManager.kt
```diff
@@ -279,13 +279,14 @@ internal class WebViewManager(
...
+            messageView?.setWebView(webView!!)
             if (newHeight != null) {
                 lastPageHeight = newHeight
-                messageView!!.updateHeight(newHeight)
+                messageView?.updateHeight(newHeight)
             }
-            messageView!!.showView(activity)
-            messageView!!.checkIfShouldDismiss()
+            messageView?.showView(activity)
+            // Executed in the same thread
```
jkasten2, line 288: I don't understand the importance of this comment. Isn't everything in this function running in the same thread?

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/WebViewManager.kt
jennantilla: Updated comment for clarity!

jkasten2 (approved):

Issue #1843, opened 2023-09-20
[Bug]: same language code("en") for all users


zion3w:
### What happened?

start using latest sdk, and there is always "en" language code for all users. had test on a device with "ru" lang, and have "en" in onesignal dashboard.

### Steps to reproduce?

```Markdown
1. install on a device with not "en" language and check dashboard
```

### What did you expect to happen?

to see real device language tag in dashboard) how I'm going to send messages on a different language users if it's "en" only ?)

### OneSignal Android SDK version

5.0.0

### Android version

13, 12, 11, 10, 9, 8, 7.1, 7.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

whyamsx:
https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/src/main/res/values/strings.xml

Also this messages need translate to method **promptForPushNotificationsWithUserResponse**
I want set another language to this messages, but not have possibility

jennantilla:
Hi @zion3w thanks for reaching out! We're working on a fix for this and will let you know when it's released!

jennantilla:
@zion3w thanks for your patience! This fix has been included in our latest [release 5.0.2](https://github.com/OneSignal/OneSignal-Android-SDK/releases/tag/5.0.2). Let us know if you have any other questions or concerns.

PR #1847, opened 2023-09-21
Fix language not being set

jennantilla:
# Description
## One Line Summary
This fixes an issue where language was not being set as a user property when a new user is created.

## Details
This issue resulted in language being set as "en" by default despite device settings.
In order to send the language we are building upon the properties map added in #1842. An additional property set there (language) will be hydrated to the properties model when the response is received.

This fixes https://github.com/OneSignal/OneSignal-Android-SDK/issues/1843

### Motivation
Ensure language matches device.

### Scope
user properties during createUser

# Testing
## Unit testing
Added language to properties object in unit tests. Ran/passed unit tests. 

## Manual testing
Tested a fresh install with a Pixel 7 emulator running Android 13 with device language set to French. User is created with language code 'fr' and reflected in dashboard.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1847)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
@@ -26,6 +26,7 @@ import com.onesignal.user.internal.properties.PropertiesModelStore
 import com.onesignal.user.internal.subscriptions.SubscriptionModel
 import com.onesignal.user.internal.subscriptions.SubscriptionModelStore
 import com.onesignal.user.internal.subscriptions.SubscriptionType
+import java.util.Locale
 
 internal class LoginUserOperationExecutor(
     private val _identityOperationExecutor: IdentityOperationExecutor,
@@ -95,6 +96,7 @@ internal class LoginUserOperationExecutor(
         var subscriptions = mapOf<String, SubscriptionObject>()
         val properties = mutableMapOf<String, String>()
         properties["timezone_id"] = TimeUtils.getTimeZoneId()!!
+        properties["language"] = Locale.getDefault().getLanguage()
 
         if (createUserOperation.externalId != null) {
             val mutableIdentities = identities.toMutableMap()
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
```diff
@@ -99,6 +94,9 @@ internal class LoginUserOperationExecutor(
     private suspend fun createUser(createUserOperation: LoginUserOperation, operations: List<Operation>): ExecutionResponse {
         var identities = mapOf<String, String>()
         var subscriptions = mapOf<String, SubscriptionObject>()
+        val properties = mutableMapOf<String, String>()
+        properties["timezone_id"] = TimeUtils.getTimeZoneId()!!
+        properties["language"] = Locale.getDefault().getLanguage()
```
jkasten2, line 99: We will want to use our `ILanguageContext` instead of using this directly. This is because there may have been a call to `OneSignal.User.setLanguage()` to change the language that this needs to pick up.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
jennantilla: Updated!

jkasten2 (approved):

jennantilla:
@nan-li noticed we needed to update LoginUserOperationExecutor tests to take in the additional parameter of MockHelper.languageContext(). See [d49858c](https://github.com/OneSignal/OneSignal-Android-SDK/pull/1847/commits/d49858cc61b26731e4dd02ffea05803db0f00635)

cc @jkasten2

nan-li (approved):

PR #1848, opened 2023-09-22
fix: NullPointerException getTargetSdkVersion

kyunkakata:
# Description
Sometime, the application starts but the context was not initialized in OneSignal module soon enough and it will be null. This happens sometime with React Native application. And the block of above code will be executed when NotificationPermissionController is initialized --> Context is null. So it should be moved to block of code of the function prompt below. So when we call from JS promptForPushNotificationsWithUserResponse later, the context should be initialized.

### Motivation
Fix the crash https://github.com/OneSignal/react-native-onesignal/issues/1516

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1848)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationPermissionController.kt
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationPermissionController.kt
@@ -43,9 +43,10 @@ object NotificationPermissionController : PermissionsActivity.PermissionCallback
     }
 
     @ChecksSdkIntAtLeast(api = 33)
-    val supportsNativePrompt =
+    val supportsNativePrompt: Boolean by lazy = {
         Build.VERSION.SDK_INT > 32 &&
             OSUtils.getTargetSdkVersion(OneSignal.appContext) > 32
+    }
 
     fun prompt(
         fallbackToSettings: Boolean,


OneSignalSDK/onesignal/src/main/java/com/onesignal/NotificationPermissionController.kt
```diff
@@ -42,15 +42,15 @@ object NotificationPermissionController : PermissionsActivity.PermissionCallback
...
-    val supportsNativePrompt =
-        Build.VERSION.SDK_INT > 32 &&
-            OSUtils.getTargetSdkVersion(OneSignal.appContext) > 32
-
     fun prompt(
         fallbackToSettings: Boolean,
         callback: OneSignal.PromptForPushNotificationPermissionResponseHandler?
     ) {
+
+        val supportsNativePrompt =
```
jkasten2, line 50: This make this code imperative instead of declarative which makes this harder to read.

I would recommend converting this a [Kotlin Lazy property](https://kotlinlang.org/docs/delegated-properties.html#lazy-properties)

jkasten2 (changes requested):
@kyunkakata thanks for the PR! I left some feedback that need to be addressed before we can accept this PR.

jkasten2 (changes requested):
@kyunkakata Thanks for addressing the feedback. Could you clean up your commit history to be a single commit? You can use git rebase on your fork to do so or make a new fork if that is easier.

jkasten2 (approved):
Thanks for the fixups! Approved and merging it into main.

joaquinvaz:
@jkasten2 can we get a release on this please? we are having this issue on production and need a fix asap please

nan-li:
I'm encountering an issue running the DevApp after this PR merged. Will investigate further.
SDKVersionChange

Issue #1858, opened 2023-10-03
[Bug]: EventProducer.fire


ayushi-sk:
### What happened?

java.util.ConcurrentModificationException

### Steps to reproduce?

```Markdown
Sometimes on app start, sometimes on notification receive
```

### What did you expect to happen?

Not crash the app

### OneSignal Android SDK version

5.0.2

### Android version

13

### Relevant log output

```Shell
Fatal Exception: java.util.ConcurrentModificationException:
       at java.util.ArrayList$Itr.next(ArrayList.java:860)
       at com.onesignal.common.events.EventProducer.fire(EventProducer.kt:43)
       at com.onesignal.core.internal.application.impl.ApplicationService.setCurrent(ApplicationService.kt:50)
       at com.onesignal.core.internal.application.impl.ApplicationService.onActivityStarted(ApplicationService.kt:145)
       at android.app.Application.dispatchActivityStarted(Application.java:413)
       at android.app.Activity.dispatchActivityStarted(Activity.java:1483)
       at android.app.Activity.onStart(Activity.java:2017)
       at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1566)
       at android.app.Activity.performStart(Activity.java:8572)
       at android.app.ActivityThread.handleStartActivity(ActivityThread.java:3934)
       at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:221)
       at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:201)
       at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:173)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2415)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:211)
       at android.os.Looper.loop(Looper.java:300)
       at android.app.ActivityThread.main(ActivityThread.java:8410)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:559)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:954)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

[...]

Issue #1857, opened 2023-10-03
[Bug]: Fatal Exception: java.util.ConcurrentModificationException


ayushi-sk:
### What happened?

```
Fatal Exception: java.util.ConcurrentModificationException:
       at java.util.ArrayList$Itr.next(ArrayList.java:860)
       at com.onesignal.common.modeling.ModelStore.persist(ModelStore.kt:140)
       at com.onesignal.common.modeling.ModelStore.addItem(ModelStore.kt:108)
       at com.onesignal.common.modeling.ModelStore.addItem$default(ModelStore.kt:98)
       at com.onesignal.common.modeling.ModelStore.add(ModelStore.kt:46)
       at com.onesignal.common.modeling.IModelStore$DefaultImpls.add$default(IModelStore.kt:37)
       at com.onesignal.core.internal.operations.impl.OperationRepo.internalEnqueue(OperationRepo.kt:76)
       at com.onesignal.core.internal.operations.impl.OperationRepo.enqueue(OperationRepo.kt:60)
       at com.onesignal.core.internal.operations.IOperationRepo$DefaultImpls.enqueue$default(IOperationRepo.kt:16)
       at com.onesignal.core.internal.operations.listeners.SingletonModelStoreListener.onModelUpdated(SingletonModelStoreListener.kt:51)
       at com.onesignal.common.modeling.SingletonModelStore$onModelUpdated$1.invoke(SingletonModelStore.kt:50)
       at com.onesignal.common.modeling.SingletonModelStore$onModelUpdated$1.invoke(SingletonModelStore.kt:50)
       at com.onesignal.common.events.EventProducer.fire(EventProducer.kt:44)
       at com.onesignal.common.modeling.SingletonModelStore.onModelUpdated(SingletonModelStore.kt:50)
       at com.onesignal.common.modeling.ModelStore$onChanged$1.invoke(ModelStore.kt:74)
       at com.onesignal.common.modeling.ModelStore$onChanged$1.invoke(ModelStore.kt:74)
       at com.onesignal.common.events.EventProducer.fire(EventProducer.kt:44)
       at com.onesignal.common.modeling.ModelStore.onChanged(ModelStore.kt:74)
       at com.onesignal.common.modeling.Model$notifyChanged$1.invoke(Model.kt:296)
       at com.onesignal.common.modeling.Model$notifyChanged$1.invoke(Model.kt:296)
[...]

Issue #1856, opened 2023-10-03
[Bug]: Fatal Exception: java.util.ConcurrentModificationException


ayushi-sk:
### What happened?
```
Fatal Exception: java.util.ConcurrentModificationException:
       at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:757)
       at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:790)
       at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:788)
       at com.onesignal.common.modeling.Model.toJSON(Model.kt:312)
       at com.onesignal.common.modeling.ModelStore.persist(ModelStore.kt:141)
       at com.onesignal.common.modeling.SingletonModelStore.replace(SingletonModelStore.kt:36)
       at com.onesignal.core.internal.config.impl.ConfigModelStoreListener$fetchParams$1.invokeSuspend(ConfigModelStoreListener.kt:101)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:32)
       at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:102)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
       at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:280)
       at kotlinx.coroutines.BlockingCoroutine.joinBlocking(BlockingCoroutine.java:85)
       at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(BuildersKt__Builders.kt:59)
       at kotlinx.coroutines.BuildersKt.runBlocking(Builders.kt:1)
       at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(BuildersKt__Builders.kt:38)
       at kotlinx.coroutines.BuildersKt.runBlocking$default(Builders.kt:1)
       at com.onesignal.common.threading.ThreadUtilsKt$suspendifyOnThread$1.invoke(ThreadUtils.kt:60)
[...]

Issue #1850, opened 2023-09-24
[Bug]: ConcurrentModificationException at initializeFromModel in production


fanwgwg:
### What happened?

From user crash logs, we're seeing **quite a few**  ConcurrentModificationException from OneSignal's `initializeFromModel` method

Stacktrace:
```
Fatal Exception: java.util.ConcurrentModificationException:
       at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:757)
       at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:790)
       at java.util.LinkedHashMap$LinkedEntryIterator.next(LinkedHashMap.java:788)
       at com.onesignal.common.modeling.Model.initializeFromModel(Model.kt:123)
       at com.onesignal.core.internal.config.impl.ConfigModelStoreListener$fetchParams$1.invokeSuspend(ConfigModelStoreListener.kt:69)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:33)
       at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:102)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:284)
       at kotlinx.coroutines.BlockingCoroutine.joinBlocking(BlockingCoroutine.java:85)
       at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(BuildersKt__Builders.kt:59)
       at kotlinx.coroutines.BuildersKt.runBlocking(Builders.kt:1)
       at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(BuildersKt__Builders.kt:38)
       at kotlinx.coroutines.BuildersKt.runBlocking$default(Builders.kt:1)
       at com.onesignal.common.threading.ThreadUtilsKt$suspendifyOnThread$1.invoke(ThreadUtils.kt:60)
       at com.onesignal.common.threading.ThreadUtilsKt$suspendifyOnThread$1.invoke(ThreadUtils.kt:59)
[...]

PR #1876, opened 2023-10-19
Fix: Add synchronized blocks to prevent ConcurrentModificationException

jennantilla:
# Description
## One Line Summary
Add synchronized blocks to prevent ConcurrentModificationException

### Motivation
Reports of ConcurrentModificationException occurring in various areas of the SDK:

**ModelStore.kt**
- ModelStore.add - #1820 
- ModelStore.persist - #1857 

**Model.kt**
- Model.toJSON - #1856 
- Model.initializeFromModel - #1850 #1819 

**EventProducer.kt**
EventProducer.fire - #1858 

### Scope
The methods reported in the various stack traces were wrapped in synchronized blocks to make them thread-safe.

Additional blocks may need to be added to other areas of the SDK if additional crashes are reported in the future.

# Testing
## Manual testing
The exact conditions needed to trigger these exceptions are unclear--the only way I was able to reproduce a similar crash was from some custom code. Adding synchronized blocks to this code prevented the crash from occurring on Pixel 7 emulator running Android 33. If reproduction steps can be verified, eventual unit tests may be added in the future to help check for these exceptions.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1876)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt
@@ -17,19 +17,25 @@ open class EventProducer<THandler> : IEventNotifier<THandler> {
     private val _subscribers: MutableList<THandler> = Collections.synchronizedList(mutableListOf())
 
     override fun subscribe(handler: THandler) {
-        _subscribers.add(handler)
+        synchronized(_subscribers) {
+            _subscribers.add(handler)
+        }
     }
 
     override fun unsubscribe(handler: THandler) {
-        _subscribers.remove(handler)
+        synchronized(_subscribers) {
+            _subscribers.remove(handler)
+        }
     }
 
     /**
      * Subscribe all from an existing producer to this subscriber.
      */
     fun subscribeAll(from: EventProducer<THandler>) {
-        for (s in from._subscribers) {
-            subscribe(s)
+        synchronized(_subscribers) {
+            for (s in from._subscribers) {
+                subscribe(s)
+            }
         }
     }
 
@@ -40,8 +46,10 @@ open class EventProducer<THandler> : IEventNotifier<THandler> {
      * @param callback The callback will be invoked for each subscribed handler, allowing you to call the handler.
      */
     fun fire(callback: (THandler) -> Unit) {
-        for (s in _subscribers) {
-            callback(s)
+        synchronized(_subscribers) {
+            for (s in _subscribers) {
+                callback(s)
+            }
         }
     }
 
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
@@ -122,20 +122,25 @@ open class Model(
      * @param model The model to initialize this model from.
      */
     fun initializeFromModel(id: String?, model: Model) {
-        synchronized(initializationLock) {
-            for (item in model.data) {
-                if (item.value is Model) {
-                    val childModel = item.value as Model
-                    childModel._parentModel = this
-                    data[item.key] = childModel
-                } else {
-                    data[item.key] = item.value
-                }
-            }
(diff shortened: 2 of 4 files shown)


arthurgeron-work:
Thank you for digging into this issue

emawby (commented):
Lets try using the @synchronized class/function annotations to make this read more cleanly.

nan-li:
> Lets try using the @synchronized class/function annotations to make this read more cleanly.

`@synchronized` methods protect methods only and allow one thread to execute a particular method at a time, but different methods can execute at the same time. So it would not be for our purposes. I think we still have to use synchronized blocks.

jennantilla:
@brismithers some additions here based on your analysis in Asana, if you wouldn't mind taking a quick look to ensure I got it all!

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt` line 125 at r2](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1876#-NiBBwBM4Z8HM_tawt1c:-NiBBwBM4Z8HM_tawt1d:blzudxu) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/e8bf0e89859cb8ddc9b2b85517691f471db3c188/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt#L125)):*
> ```Kotlin
>      */
>     fun initializeFromModel(id: String?, model: Model) {
>         synchronized(initializationLock) {
> ```

I believe the `data.clear()` is needed, if the incoming model doesn't have a particular property set the current model will not have that property removed.  I wonder if it might be safer to build up a local `MutableMap` then at the end replace `data` with the new instance?

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt` line 127 at r3](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1876#-NiFCvla3iXuh-4HNgI9:-NiFCvla3iXuh-4HNgIA:bbi0q58) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/c742bbf20b194004a9de80497972cd70e5079f45/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt#L127)):*
> ```Kotlin
>         model: Model,
>     ) {
>         val newData = Collections.synchronizedMap(mutableMapOf<String, Any?>())
> ```

nit: this doesn't need to by synchronized since it is only used by this method as a local variable.

jennantilla (commented):
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1876)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @brismithers, @emawby, @jkasten2, @nan-li, and @shepherd-l)

___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt` line 125 at r2](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1876#-NiBBwBM4Z8HM_tawt1c:-NiH--YM6mibozgnXeIS:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/e8bf0e89859cb8ddc9b2b85517691f471db3c188/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt#L125)):*

I believe the `data.clear()` is needed, if the incoming model doesn't have a particular property set the current model will not have that property removed.  I wonder if it might be safer to build up a local `MutableMap` then at the end replace `data` with the new instance?
</blockquote></details>

Done.

emawby (approved):

PR #1901, opened 2023-11-06
Fix issue with migration from v4 to v5 when obfuscation is used by the app

brismithers:
# Description
## One Line Summary
Fix issue with migration from v4 to v5 when obfuscation is used by the app

## Details
The v5 SDK will read/write to a preference store called "OneSignal", it is [defined as a string ](https://github.com/OneSignal/OneSignal-Android-SDK/blob/user-model/main/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/IPreferencesService.kt#L155) so it is always expected a file at /data/data/<app_id>/shared_prefs/OneSignal.xml .  The v4 SDK however will read/write to a preference store defined as OneSignal.class.getSimpleName(); (see [here](https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java#L59)).

When an app is using ProGuard (or similar) the OneSignal class name can be renamed, as a result the underlying preference store for v4 can be named anything.  So when going from v4 -> v5 (or even different builds of v4 to v4),  on startup the SDK does not find the preference store and believes it is starting "fresh", resulting in a new subscription for the device.

Since we don't know what name was used, the underlying fix is to scan the existing preference stores to find the correct one and rename it to "OneSignal.xml", which is the static name as of v5.

### Motivation
Desire to only have 1 subscription per device across app builds.

### Scope
This fix is a startup/initialization fix targeting devices that have a new app install migrating from SDK v4 to SDK v5.

# Testing

## Manual testing
1. On an emulator, run as a fresh install the OneSignal Example app on v4, but modifying the [PREFS_ONESIGNAL](https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignalPrefs.java#L59) value to a static string like "g3" (pick any string).  A new subscriber should be created when started.
2. On the same emulator and without uninstalling the v4 version, run the OneSignal example app on this version of the SDK.
3. When `OneSignal.initWithContext` is called the preference file should be appropriately renamed, and the subscriber created from the v4 SDK version is used by the v5 SDK.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1901)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/PreferenceStoreFix.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/PreferenceStoreFix.kt
@@ -1,0 +1,49 @@
+package com.onesignal.core.internal.preferences
+
+import android.content.Context
+import android.os.Build
+import com.onesignal.debug.LogLevel
+import com.onesignal.debug.internal.logging.Logging
+import java.io.File
+
+object PreferenceStoreFix {
+
+    /**
+     * Ensure the OneSignal preference store is not using the v4 obfuscated version, if one
+     * exists.
+     */
+    fun ensureNoObfuscatedPrefStore(context: Context) {
+        try {
+            // In the v4 version the OneSignal shared preference name was based on the OneSignal
+            // class name, which might be minimized/obfuscated if the app is using ProGuard or
+            // similar.  In order for a device to successfully migrate from v4 to v5 picking
+            // up the subscription, we need to copy the shared preferences from the obfuscated
+            // version to the static "OneSignal" preference name.  We only do this
+            // if there isn't already a "OneSignal" preference store.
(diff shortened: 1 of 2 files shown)


shepherd-l:
I noticed that the linting check fails. The linting fix has been merged and it should be working now

Could you fix the linting errors? I can also do it as well

brismithers:
Done

jkasten2 (approved):

emawby (approved):

PR #1902, opened 2023-11-07
Catch All Exception to fix DeadSystemException

puderty:
# Description
## One Line Summary
On some device like Huawei, there's a lot of DeadSystemException when calling getPackageInfo, we can just ignore it.

### Motivation
We can all ignore that exception, by merge this PR, we don't need to compile this library by ourself anymore.

### Scope

# Testing
## Unit testing
Not answered

## Manual testing
Not answered

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1902)


jkasten2 (commented):
Adding some technical background here for reference. Android's documentation noted that the [DeadSystemException](https://developer.android.com/reference/android/os/DeadSystemException) is:
> The core Android system has died and is going through a runtime restart. All running apps will be promptly killed.


Since this exception is not the fault of the app (or SDK) there isn't anything the app (or SDK) can do to recover from such a state. Catching and ignoring these won't improve the end-user experience since the app is already being killed, however doing so will clean up the noise of exceptions that can't be handled.

jkasten2 (changes requested):
Catching `Exception` (or `Throwable`) can hide other issues which could result in broken states that make finding root issues difficult. Can you add catches for `DeadSystemException` here instead of changing all these to catch all exceptions?

jkasten2:
@puderty Thanks for pointing out the places where `DeadSystemException` can be thrown. A PR was created based on this:
https://github.com/OneSignal/OneSignal-Android-SDK/pull/1937

It has been merged in, and will be in the next 4.x.x release.

PR #1903, opened 2023-11-07
initWithContext synchronization fix

brismithers:
# Description
## One Line Summary
Synchronize initWithContext so initialization occurs once, and prevent login/logout from being called before `initWithContext`.

## Details
It is possible for an app to call `initWithContext` on different threads at the same time, causing the initialization logic to run more than once.  If this were to happen, the internal state of the SDK will not be correct.  For example, the initialization code drives calling `registerActivityLifecycleCallbacks`, we don't want to have 2 of those!

This change synchronizes the initialization logic so the SDK can only go through `initWithContext` once.  Additionally, the `login` and `logout` functions now throw exceptions rather than log an error, if called before `initWithContext`.  This is done to raise programming errors to the caller, rather than swallow them in the logs.

### Motivation
No specific issues to point to, although the belief is this is a concurrency issue that needs to be resolved.

### Scope
This change is limited to initialization and login/logout flows.

# Testing
## Unit testing
2 additional unit tests were added to demonstrate the expected behavior of calling login/logout before calling `initWithContext`.

## Manual testing
Run the OneSignal Example App on an emulator to ensure initialization of the SDK continues to operate successfully.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1903)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -250,10 +250,11 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
             startupService!!.bootstrap()
 
             if (forceCreateUser || !identityModelStore!!.model.hasProperty(IdentityConstants.ONESIGNAL_ID)) {
-                val legacyPlayerId = preferencesService!!.getString(
-                    PreferenceStores.ONESIGNAL,
-                    PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID
-                )
+                val legacyPlayerId =
+                    preferencesService!!.getString(
+                        PreferenceStores.ONESIGNAL,
+                        PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
+                    )
                 if (legacyPlayerId == null) {
                     Logging.debug("initWithContext: creating new device-scoped user")
                     createAndSwitchToNewUser()
@@ -293,7 +294,7 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                         configModel!!.pushSubscriptionId = legacyPlayerId
                         subscriptionModelStore!!.add(
                             pushSubscriptionModel,
-                            ModelChangeTags.NO_PROPOGATE
+                            ModelChangeTags.NO_PROPOGATE,
                         )
                         suppressBackendOperation = true
                     }
@@ -304,13 +305,13 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                         LoginUserFromSubscriptionOperation(
                             configModel!!.appId,
                             identityModelStore!!.model.onesignalId,
-                            legacyPlayerId
+                            legacyPlayerId,
                         ),
                     )
                     preferencesService!!.saveString(
                         PreferenceStores.ONESIGNAL,
                         PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
-                        null
+                        null,
                     )
                 }
             } else {
@@ -318,8 +319,8 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                 operationRepo!!.enqueue(
                     RefreshUserOperation(
                         configModel!!.appId,
-                        identityModelStore!!.model.onesignalId
-                    )
+                        identityModelStore!!.model.onesignalId,
+                    ),
                 )
             }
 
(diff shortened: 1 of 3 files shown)


shepherd-l:
Could you fix the linting errors in OneSignalImplTests?

brismithers:
Done, also renamed OneSignalImplTests to OneSignalImpTests to align with OneSignalImp

jennantilla (approved):
Makes sense to add additional synchronization here, great work!

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -172,126 +173,160 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
 
         // do not do this again if already initialized
         if (isInitialized) {
```
jkasten2, line 175: A  `if (isInitialized)` check was added in this PR in the new `synchronized` block, so we can remove this one.

jkasten2 (changes requested):

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt` line 175 at r2](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1903#gh-1388531783:-Nj8J30xDi47oY88LdL7:blnekw) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/986adfa4f6f150c77f77dc102492a38dba94a109/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt#L175)):*

A  `if (isInitialized)` check was added in this PR in the new `synchronized` block, so we can remove this one.
</blockquote></details>

The check is done twice intentionally. Most post-initialization callers will hit this if block and can do so without lock contention.  We need to check `isInitialized` again after obtaining the lock to ensure there wasn't someone else (that had the lock, but didn't finish initialization).

jkasten2 (approved):
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1903)* status: all files reviewed, 1 unresolved discussion (waiting on @brismithers, @emawby, @nan-li, and @shepherd-l)

___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt` line 175 at r2](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1903#gh-1388531783:-NjA8D9v1u8-1rqpld3t:b-utacj8) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/986adfa4f6f150c77f77dc102492a38dba94a109/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt#L175)):*

The check is done twice intentionally. Most post-initialization callers will hit this if block and can do so without lock contention.  We need to check `isInitialized` again after obtaining the lock to ensure there wasn't someone else (that had the lock, but didn't finish initialization).
</blockquote></details>

Gotcha, I have seen this before, it trades performance (in-most cases) over less code. Since init isn't normally called more than not sure if I like that trade off. However it's more of a nit so I'll unblock the PR and leave that up to you.

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt` line 175 at r2](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1903#gh-1388531783:-NjDLAW7B56CdZhrbgiu:bw237qq) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/986adfa4f6f150c77f77dc102492a38dba94a109/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt#L175)):*

Gotcha, I have seen this before, it trades performance (in-most cases) over less code. Since init isn't normally called more than not sure if I like that trade off. However it's more of a nit so I'll unblock the PR and leave that up to you.
</blockquote></details>

that makes sense. I've removed the extra check before obtaining the lock.  If we notice high contention cases this can always be added in the future!

PR #1909, opened 2023-11-14
Add getter for onesignalId and UserStateObserver

jinliu9508:
# Description
Add getters for onesignal ID and external ID, and a  UserStateObserver to listen on User State changes to know when these values are changed.

### Motivation
From developer feedback and to support integration partners, we are exposing the onesignal ID and external ID with getters.

We also want to grant developers the ability to add observers that can be called when there is a change in user state.

### Scope
UserState contains onesignalId and externalId (both can be empty strings if not available or not set).

# Testing
## Unit testing
** need to figure out a good testing plan here

## Manual testing
App built

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
```diff
@@ -1,6 +1,7 @@
 package com.onesignal.user
 
 import com.onesignal.OneSignal
+import com.onesignal.user.internal.state.IUserStateObserver
```
brismithers, line 4: since `IUserManager` is part of the external interface, it should not reference anything under an `internal` package.  I recommend you move `IUserStateObserver`, `UserChangedState`, and `UserState` under the `com.onesignal.user` package as they will also be part of the external API.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/state/UserState.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import org.json.JSONObject
+
+/**
+ * A user state.
+ */
+class UserState (
+        /**
+         * The unique identifier for OneSignal account. This will be an empty string
+         * until the user has been successfully logged in on the backend and
```
brismithers, line 11: If it will be an empty string when there is no onesignalId, what does a null `onesignalId` represent?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/state/UserState.kt
```diff
@@ -0,0 +1,27 @@
+package com.onesignal.user.internal.state
+
+import org.json.JSONObject
+
+/**
+ * A user state.
+ */
+class UserState (
```
brismithers, line 8: The majority of the external API is defined through interfaces, even for simple POJO types like this ([example](https://github.com/OneSignal/OneSignal-Android-SDK/blob/user-model/main/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/IActionButton.kt)), to solidify an understanding of what is intended as an external interface.

I think it's okay to define a class as part of the external interface, but keep in mind anything that goes in here could be interpreted to be part of the API.

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ISingletonModelStoreChangeHandler.kt` line 15 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#-NktzmhuDt3SmZV3tiPk:-NktzmhuDt3SmZV3tiPl:b-t9zmjh) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ISingletonModelStoreChangeHandler.kt#L15)):*
> ```Kotlin
>      */
>     fun onModelReplaced(
>         model: ModelReplacedArgs<TModel>,
> ```

nit: rename `model` to `args` as it now represents more than the model.

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt` line 252 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#-Nku-z5FAbMUr4JYNVWE:-Nku-z5FAbMUr4JYNVWF:b-wofsl4) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt#L252)):*
> ```Kotlin
>         tag: String,
>     ) {
>         val oldUserState = UserState(args.oldModel[IdentityConstants.ONESIGNAL_ID], args.oldModel.externalId)
> ```

just curious, couldn't you use `args.oldModel.onesignalId` here instead?

brismithers:
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt` line 271 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#-Nku0WJo4WVaYMrSdpfA:-Nku0WJo4WVaYMrSdpfB:blswdkp) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt#L271)):*
> ```Kotlin
>         tag: String,
>     ) {
>         if (args.property.equals(IdentityConstants.ONESIGNAL_ID)) {
> ```

I think you also need to check and fire if the externalId has changed here as well?

brismithers (commented):
> Reviewed 3 of 7 files at r2, 1 of 6 files at r6, 3 of 7 files at r7, 21 of 21 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909)* status: all files reviewed, 5 unresolved discussions (waiting on @emawby, @jennantilla, @jinliu9508, @jkasten2, @nan-li, and @shepherd-l)

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ISingletonModelStoreChangeHandler.kt
```diff
@@ -12,7 +12,7 @@ interface ISingletonModelStoreChangeHandler<TModel> where TModel : Model {
      * @param tag The tag which identifies how/why the model was replaced.
      */
     fun onModelReplaced(
-        model: TModel,
+        model: ModelReplacedArgs<TModel>,
```
nan-li, line 15: args

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/SingletonModelStore.kt
```diff
@@ -38,10 +38,12 @@ open class SingletonModelStore<TModel>(
         tag: String,
     ) {
         synchronized(replaceLock) {
+            val oldModel = this.store.create() ?: throw Exception("Unable to initialize model from store $store")
+            oldModel.initializeFromModel(singletonId, this.model)
```
nan-li, line 42: Hmm, this is a bit convoluted workaround and do prefer to check if there is even an old model

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/listeners/SingletonModelStoreListener.kt
```diff
@@ -62,7 +63,7 @@ internal abstract class SingletonModelStoreListener<TModel>(
      *
      * @return The operation to enqueue when the model has been replaced, or null if no operation should be enqueued.
      */
-    abstract fun getReplaceOperation(model: TModel): Operation?
+    abstract fun getReplaceOperation(model: ModelReplacedArgs<TModel>): Operation?
```
nan-li, line 66: args

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
```diff
@@ -138,4 +138,15 @@ interface IUserManager {
      * Return a copy of all local tags from the current user.
      */
     fun getTags(): Map<String, String>
+
+    /**
```
nan-li, line 142: Getters for onesignal ID and external ID needs to be exposed through this interface for public consumption

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserState.kt
```diff
@@ -0,0 +1,39 @@
+package com.onesignal.user
+
+import com.onesignal.common.IDManager
+import org.json.JSONObject
+
+/**
+ * A user state.
+ */
+class UserState {
```
nan-li, line 9: Consider making non-optional and default to `""` to match push subscription state and simplify `toJSONObject()` method.

jinliu9508 (commented):
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909)* status: 12 of 29 files reviewed, 13 unresolved discussions (waiting on @brismithers, @emawby, @jennantilla, @jkasten2, @nan-li, and @shepherd-l)

___
*[`Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/model/MainActivityViewModel.java` line 31 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1411424804:-NkwKbgDE4NtkBrFOFOu:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/model/MainActivityViewModel.java#L31)):*

Since the rest of the dev app uses `android.util.Log`, let's use that instead of the SDK's internal logging system.
</blockquote></details>

Done.
___
*[`Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/model/MainActivityViewModel.java` line 70 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1411426763:-NkwKdUqBZZ7cx5dwEiV:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/model/MainActivityViewModel.java#L70)):*

Since the user state observer is not affecting anything in ActivityViewModel, I recommend to put into MainApplication. You can see some examples there already of listeners that are added like Notification Click Listener.
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt` line 142 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1416263074:-NkwKgqOCmxCzIwgun5Y:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt#L142)):*

Getters for onesignal ID and external ID needs to be exposed through this interface for public consumption
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserState.kt` line 9 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1416267538:-NkwKjG14JJNwme4SHye:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserState.kt#L9)):*

Consider making onesignalId and externalId non-optional and default to `""` to match push subscription state and look nice in `toJSONObject()` method.
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt` line 252 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#-Nku-z5FAbMUr4JYNVWE:-NkwKlZoCLngSb9ySfn8:bf91r9l) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt#L252)):*

just curious, couldn't you use `args.oldModel.onesignalId` here instead?
</blockquote></details>

onesignalId might not be exist in the model so we couldn't directly get onesignalId that way. We have decided to not carrying over the old model at this point so I have removed this part of code. 
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt` line 271 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#-Nku0WJo4WVaYMrSdpfA:-Nkwqvhm7FCX8V_g6bVr:brrtlb9) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt#L271)):*

I think you also need to check and fire if the externalId has changed here as well?
</blockquote></details>

Good question. In the team meeting we have gone through all possible scenarios when we need to fire up the callback, we found that change of externalId only is not enough to determine whether we should fire the callback. So only the change of onesignalId is under the watch here.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt` line 47 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1411434116:-NkwrX1V0NfpHbNKRqDL:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt#L47)):*

to self: why made open? Come back and check at end of review. Is it because this is called from UserManager.kt?
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/IModelChangedHandler.kt` line 55 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1411443402:-NkwrXes2lrOuw6bpwmx:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/IModelChangedHandler.kt#L55)):*

this class should go in ISingletonModelStoreChangeHandler.kt and the comment updated to reflect that it is used in ISingletonModelStoreChangeHandler.
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ISingletonModelStoreChangeHandler.kt` line 15 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1416258956:-Nkwrhjj60nUhCsmPVyN:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ISingletonModelStoreChangeHandler.kt#L15)):*

args
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/SingletonModelStore.kt` line 42 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1416260156:-Nkwrl4eEEDxl2sSUOdd:b-34yqwp) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/SingletonModelStore.kt#L42)):*

Hmm, this is a bit convoluted workaround and do prefer to check if there is even an old model
</blockquote></details>

We just went through this and the old model is no longer stored.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/listeners/SingletonModelStoreListener.kt` line 66 at r8](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1416260956:-NkwruDhAJq51BNt46fj:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/689b2b9429d31b5ef3071c215f7f4caeedd982b1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/listeners/SingletonModelStoreListener.kt#L66)):*

args
</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/state/UserState.kt` line 8 at r1](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1392708588:-NkwsMqP6GMFanpLSa2Q:b-896fix) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/148c1a30c1e7b45542e79672cbd6b37969cfd9e1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/state/UserState.kt#L8)):*

The majority of the external API is defined through interfaces, even for simple POJO types like this ([example](https://github.com/OneSignal/OneSignal-Android-SDK/blob/user-model/main/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/IActionButton.kt)), to solidify an understanding of what is intended as an external interface.

I think it's okay to define a class as part of the external interface, but keep in mind anything that goes in here could be interpreted to be part of the API.


</blockquote></details>

Done.
___
*[`OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/state/UserState.kt` line 11 at r1](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1909#gh-1392686820:-Nkwrw9u6YwMSLLd4f9H:bll4830) ([raw file](https://github.com/OneSignal/OneSignal-Android-SDK/blob/148c1a30c1e7b45542e79672cbd6b37969cfd9e1/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/state/UserState.kt#L11)):*

If it will be an empty string when there is no onesignalId, what does a null `onesignalId` represent?
</blockquote></details>

The code has changed so now that it is no longer an optional field. When onesignalId is not available, it will be an empty string "" so that it is consistent and more friendly to JsonObject across wrappers.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
```diff
@@ -219,4 +235,30 @@ internal open class UserManager(
...
+        args: ModelChangedArgs,
+        tag: String,
+    ) {
+        if (args.property == IdentityConstants.ONESIGNAL_ID) {
+            val newUserState = UserState(args.newValue.toString(), externalId)
+            this.changeHandlersNotifier.fire {
+                it.onUserStateChange(UserChangedState(newUserState))
+            }
+        }
+        Logging.debug(args.property)
```
nan-li, line 262: I think this log prints just "onesignal_id" and was left over from testing? Let's remove it. I don't think it's necessary to replace with an informative log here either.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserState.kt
```diff
@@ -0,0 +1,28 @@
...
+     * The unique identifier for your OneSignal account. This will be an empty string until the
+     * user has been successfully logged in on the backend and assigned an ID.
+     * Use [addObserver] to be notified when the [onesignalId] has been successfully assigned.
+     */
+    val onesignalId: String,
+    /**
+     * The external identifier that you use to identify users. This will be an empty string
+     * until the user has been successfully logged in on the backend and assigned an ID.
+     * Use [addObserver] to be notified when the [externalId] has
+     * been successfully assigned.
```
nan-li, line 19: This sentence "This will be an empty string _until_ the user has been successfully logged in on the backend and assigned an ID" could be confusing for app developers. Perhaps something like:

> The external identifier that you use to identify users. Use [addObserver] to be notified when the [externalId] has been successfully assigned. This will be an empty string if no external identifier has been assigned to the associated [onesignalId].

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserStateObserver.kt
```diff
@@ -0,0 +1,15 @@
+package com.onesignal.user
+
+/**
+ * A user state changed handler. Implement this interface and provide the implementation
```
nan-li, line 4: > A user state changed **observer**. Implement this interface and provide the implementation to be notified when the user **state** has changed.

Under the hood, the user has already changed in the SDK so this is a bit more correct to say it is the state object that changed.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserStateObserver.kt
```diff
@@ -0,0 +1,15 @@
...
+ * to be notified when the user has changed.
+ */
+interface IUserStateObserver {
```
nan-li, line 7: We should call out somewhere in the KDoc that app developers should be checking **both** the `onesignalId` **and** the `externalId` in the `UserState` to make sure they are grabbing the correct `onesignalId`.

For example, a new app install that calls `login("ext1")` will trigger the callback twice. App developers may just grab the very first callback and save the `onesignalId`. However, that first callback will contain an `""` `externalId`, and it is the second callback that contains the actual `onesignalId` they want and an `externalId` of `ext1`.

I'm not sure where best to put this information since it can go in the description of `IUserStateObserver` or `fun onUserStateChange`, `UserChangedState`, or `UserState`. 

I need to do the equivalent for iOS as well.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserStateObserver.kt
```diff
@@ -0,0 +1,15 @@
+package com.onesignal.user
```
nan-li, line 1: What do you think about putting the 3 new user state related files into a subpackage `package com.onesignal.user.state`? That way the high-level `com.onesignal.user` package just contains the interface still.

nan-li:
Not sure how important this is, a behavior I noticed during testing is if internet is turned off:

1. login("someNewUser") and add tags
2. call logout
3. Turn internet back on
4. The new user is created on server and tags are attached to that user correctly
5. The user state observer is not fired to notify the app developer of that user's onesignalId

App developers may always be waiting for `onesignalId` every time they login to send to their own server. If the user is already logged out or changed by the time the response is received in the SDK, the SDK really shouldn't receive anything for that user anyway. However, the app developer may have some integration they still want to attach the `onesignalId` to that user on their own server.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
```diff
@@ -25,6 +26,16 @@ interface IUserManager {
      */
     val pushSubscription: IPushSubscription
 
+    /**
+     * onesignalId, empty if this is current unavailable
+     */
```
nan-li, line 31: Just a few more suggestions. Since we are documenting this getter, let's tell them what this ID is:

> "The UUID generated by OneSignal to represent a user, empty if this is currently unavailable" 


(I pulled from [our docs](https://documentation.onesignal.com/docs/aliases-external-id#onesignal-id))

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
```diff
@@ -25,6 +26,16 @@ interface IUserManager {
...
+    val onesignalId: String
+
+    /**
+     * externalId, empty if this is current unavailable
+     */
```
nan-li, line 36: Similar to above, let's say

> The External ID is OneSignal's default and recommended alias label. This should be the main identifier you use to identify users. It is set when calling the [OneSignal.login] method.

> This is empty if the External ID has not been set.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
```diff
@@ -138,4 +149,15 @@ interface IUserManager {
      * Return a copy of all local tags from the current user.
      */
     fun getTags(): Map<String, String>
+
+    /**
+     * Add an observer to the user state, allowing the provider to be
+     * notified whenever the user state has changed.
+     */
```
nan-li, line 156: Let's add a callout to developers here on the `addObserver` method documentation:

> Important: When using the observer to retrieve the `onesignalId`, check the `externalId` as well to confirm the values are associated with the expected user.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
jinliu9508: Good point! I have modified all the comments to make them clear and consistent to our docs.

nan-li (approved):

nan-li:
We should update the migration guide? I forgot about it.

Issue #1821, opened 2023-08-26
[Bug]: IllegalStateException GoogleApiClient is not connected yet.


tajchert:
### What happened?

Crash is affecting some of my users (7 of them).

### Steps to reproduce?

```Markdown
-
```

### What did you expect to happen?

No crash.

### OneSignal Android SDK version

5.0.0

### Android version

13, 12, 11

### Relevant log output

```Shell
Fatal Exception: java.lang.IllegalStateException: GoogleApiClient is not connected yet.
       at com.google.android.gms.common.api.internal.zaax.zab(com.google.android.gms:play-services-base@@18.1.0:1)
       at com.google.android.gms.common.api.internal.zabi.zaf(com.google.android.gms:play-services-base@@18.1.0:2)
       at com.google.android.gms.common.api.internal.zabe.execute(com.google.android.gms:play-services-base@@18.1.0:12)
       at com.google.android.gms.internal.location.zzau.removeLocationUpdates(com.google.android.gms:play-services-location@@21.0.1:3)
       at com.onesignal.location.internal.controller.impl.GmsLocationController$FusedLocationApiWrapper.cancelLocationUpdates(GmsLocationController.kt:224)
       at com.onesignal.location.internal.controller.impl.GmsLocationController$LocationUpdateListener.close(GmsLocationController.kt:186)
       at com.onesignal.location.internal.controller.impl.GmsLocationController.stop(GmsLocationController.kt:105)
       at com.onesignal.location.internal.controller.impl.GmsLocationController$GoogleApiClientListener$onConnectionFailed$1.invokeSuspend(GmsLocationController.kt:150)
       at com.onesignal.location.internal.controller.impl.GmsLocationController$GoogleApiClientListener$onConnectionFailed$1.invoke(GmsLocationController.kt:2)
       at com.onesignal.location.internal.controller.impl.GmsLocationController$GoogleApiClientListener$onConnectionFailed$1.invoke(GmsLocationController.kt:1)
       at com.onesignal.common.threading.ThreadUtilsKt$suspendifyOnThread$1$1.invokeSuspend(ThreadUtils.kt:61)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:284)
[...]

PR #1940, opened 2023-12-13
Fix: Catch IllegalStateException on call to Google's location library

jennantilla:
# Description
## One Line Summary
Catch IllegalStateException on call to Google's location library to avoid crash.

### Motivation
IllegalStateException was reported in #1821, potentially due to and issue with a device's "Google Play services" app. Catching these exceptions and logging them will avoid crashes on calls to Google's location library.

### OPTIONAL - Other
While the original crash was not reproducible, adding this try/catch block to log the exception should prevent the app from crashing.

## Manual testing
Since crash was not reproducible, just tested running app to ensure that change does not impact location sharing on Android.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1940)

### DIFF
--- a/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/GmsLocationController.kt
+++ b/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/GmsLocationController.kt
@@ -231,10 +231,10 @@ internal class GmsLocationController(
             googleApiClient: GoogleApiClient,
             locationListener: LocationListener,
         ) {
-            try {
+            if (googleApiClient.isConnected) {
                 LocationServices.FusedLocationApi.removeLocationUpdates(googleApiClient, locationListener)
-            } catch (e: IllegalStateException) {
-                Logging.warn("Caught IllegalStateException in cancelLocationUpdates", e)
+            } else {
+                Logging.warn("GoogleApiClient is not connected. Unable to cancel location updates.")
             }
         }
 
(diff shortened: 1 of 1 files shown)


jkasten2 (approved):

nan-li:
Are there other reports of `IllegalStateException` aside from `GoogleApiClient is not connected yet`?

I think it may be clearer to check for `googleApiClient.isConnected` instead of using a try-catch, such as in the `getLastLocation()` or `refreshRequest()` method.

```kotlin
        fun getLastLocation(googleApiClient: GoogleApiClient): Location? {
            if (googleApiClient.isConnected) {
                return LocationServices.FusedLocationApi.getLastLocation(googleApiClient)
            }
            return null
        }
```


```kotlin
        private fun refreshRequest() {
            if (!googleApiClient.isConnected) {
                Logging.warn("Attempt to refresh location request but not currently connected!")
                return
            }
            // ... stuff
```

jennantilla:
I've updated the method per @nan-li's suggestion. Let me know if this looks good! cc @jkasten2

jkasten2 (approved):

nan-li (approved):

Issue #1910, opened 2023-11-14
[Bug]: disabling location sharing not working


JNaehringApploft:
### What happened?

The latest sdk shares the geo location when it is available in the app, although location sharing is disabled when initializing the one signal sdk.
  
```kotlin
OneSignal.initWithContext(this, ONESIGNAL_APP_ID)
OneSignal.Location.isShared = false
``` 

### Steps to reproduce?

```Markdown
1. have an app that uses geo location and integrate one signal sdk v5.0.4 with location sharing disabled
2. install and start the app
3. grant location permission (e.g. through custom ui)
4. make sure that custom app code accesses the location
5. close and restart the app at least once
6. open the onesignal dashboard for this project and look for a new subscription
7. see that the new subscription shows the location point
```

### What did you expect to happen?

I would expect that the geo location is not read and send to one signal servers, when location sharing is disabled.

### OneSignal Android SDK version

5.0.4

### Android version

13

### Specific Android models

```Markdown
Poco x4 Pro - Android 13
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jennantilla:
Hello @JNaehringApploft thank you for reaching out! I have reproduce this issue and am investigating a solution. We will let you know progress here.

johanlantz:
@jennantilla imho, two weeks is too long to address this. Customers using 5.0.x with location are likely in violation of their privacy policies, gdpr and disclousures with AppStore and Google Play. If they have said that they are not sharing information with third-parties, then presumably, uploading highly sensitive information such as gps coordinates to onesignal without user consent is really bad.

jennantilla:
@johanlantz we appreciate your patience! This will be included in the next release of the SDK.

jennantilla:
Hi everyone, thanks for your patience! Just an update that a fix for this has been added in our latest release, [5.1.0](https://github.com/OneSignal/OneSignal-Android-SDK/releases/tag/5.1.0).

PR #1942, opened 2023-12-13
Fix: Update startGetLocation to only run if location is shared

jennantilla:
# Description
## One Line Summary
Update startGetLocation to only run if location is shared

### Motivation
In #1910 it was reported that when `OneSignal.Location.setShared(false)` but location permission is granted through a custom UI, OneSignal is still accessing and recording the device's location. Adding a conditional before `startGetLocation` to check whether `isShared` is true will prevent OneSignal from accessing the device's location when `setShared` is false.

# Testing
## Manual testing
The following steps were followed on and Android 14 emulator to recreate and test the fix:

1. Add [native code](https://developer.android.com/develop/sensors-and-location/location/permissions#java) to request/grant location permission.
2. Set `OneSignal.Location.setShared(false)`
3. Restart app and observe that location is shared in OneSignal dashboard
4. Add conditional to `startGetLocation`
5. On a new device, request location permission natively
6. Restart app and observe that location is not shared in OneSignal dashboard

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [X] Location
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1942)

### DIFF
--- a/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
+++ b/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
@@ -4,6 +4,7 @@ import android.os.Build
 import com.onesignal.common.AndroidUtils
 import com.onesignal.common.threading.suspendifyOnThread
 import com.onesignal.core.internal.application.IApplicationService
+import com.onesignal.core.internal.preferences.IPreferencesService
 import com.onesignal.core.internal.startup.IStartableService
 import com.onesignal.debug.LogLevel
 import com.onesignal.debug.internal.logging.Logging
@@ -22,12 +23,14 @@ internal class LocationManager(
     private val _capturer: ILocationCapturer,
     private val _locationController: ILocationController,
     private val _locationPermissionController: LocationPermissionController,
+    private val _prefs: IPreferencesService,
 ) : ILocationManager, IStartableService, ILocationPermissionChangedHandler {
-    private var _isShared: Boolean = false
+    private var _isShared: Boolean = _prefs.getBool("OneSignal", "PREFS_OS_LOCATION_SHARED", false)!!
     override var isShared
         get() = _isShared
         set(value) {
             Logging.debug("LocationManager.setIsShared(value: $value)")
+            _prefs.saveBool("OneSignal", "PREFS_OS_LOCATION_SHARED", value)
             _isShared = value
         }
 
--- a/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
+++ b/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
@@ -179,13 +179,18 @@ internal class LocationManager(
 
     // Started from this class or PermissionActivity
     private suspend fun startGetLocation() {
-        Logging.debug("LocationManager.startGetLocation()") // with lastLocation: " + lastLocation)
-        try {
-            if (!_locationController.start()) {
-                Logging.warn("LocationManager.startGetLocation: not possible, no location dependency found")
+        if (isShared) {
+            Logging.debug("LocationManager.startGetLocation()") // with lastLocation: " + lastLocation)
+            try {
+                if (!_locationController.start()) {
+                    Logging.warn("LocationManager.startGetLocation: not possible, no location dependency found")
+                }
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
```diff
@@ -179,13 +179,18 @@ internal class LocationManager(
 
     // Started from this class or PermissionActivity
     private suspend fun startGetLocation() {
-        Logging.debug("LocationManager.startGetLocation()") // with lastLocation: " + lastLocation)
-        try {
-            if (!_locationController.start()) {
-                Logging.warn("LocationManager.startGetLocation: not possible, no location dependency found")
+        if (isShared) {
```
jkasten2, line 182: Recommend returning early instead if `isShared` is `false`, this keeps the code from getting to nested.

nan-li:
On every cold start, `isShared` defaults to `false`.

General question as I am not sure: what is the expected behavior when the app developer wants location to be shared with OneSignal?

They must set `Location.isShared = true` whenever they initialize OneSignal? Or is it expected to be a cached setting that persists across cold starts?

Also, from our [docs](https://documentation.onesignal.com/docs/mobile-sdk#getting-shared-location-status): "Whether the location is currently shared with OneSignal. (defaults to true if your app has location permission)" suggests the default is `true` unless _explicitly_ set to `false`. I think this PR may change that behavior? I did not test, but do you know what it looks like when you tested the scenario that shares location with OneSignal?

jennantilla:
@nan-li really good insight! While I did test the scenarios of setting `setShared` explicitly to true as well as omitting the method completely (it appears everything behaves as expected), I did not account for the fact that if `setShared` is set true outside of initialization (e.g., on a button click) that the value will not persist to a cold start. 

I'm also not sure what the expected behavior for the app developer should be, but it appears that the current behavior is that `setShared` should be set whenever they initialize OneSignal.

I'll think about this one some more!

jennantilla:
@nan-li I've pushed two additional commit that account for the following requirements:
- Update default value for `isShared` to false and to be accessed via preferences.
- Allow for mid-Session updates by calling `onLocationPermissionChange` when set method is called (I'm not sure if there is a better way to do this or if I'm missing any gotchas).

Insight appreciated!

OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
```diff
@@ -22,13 +23,17 @@ internal class LocationManager(
     private val _capturer: ILocationCapturer,
     private val _locationController: ILocationController,
     private val _locationPermissionController: LocationPermissionController,
+    private val _prefs: IPreferencesService,
 ) : ILocationManager, IStartableService, ILocationPermissionChangedHandler {
-    private var _isShared: Boolean = false
+    private var _isShared: Boolean = _prefs.getBool("OneSignal", "PREFS_OS_LOCATION_SHARED", false)!!
```
nan-li, line 28: Let's use the `PREFS_OS_LOCATION_SHARED` constant you created instead of hardcoded strings:
```
_prefs.getBool(PreferenceStores.ONESIGNAL, PreferenceOneSignalKeys.PREFS_OS_LOCATION_SHARED, false)!!
```

OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/LocationManager.kt
```diff
@@ -22,13 +23,17 @@ internal class LocationManager(
...
     override var isShared
         get() = _isShared
         set(value) {
             Logging.debug("LocationManager.setIsShared(value: $value)")
+            _prefs.saveBool("OneSignal", "PREFS_OS_LOCATION_SHARED", value)
```
nan-li, line 33: same as above to use the constants

nan-li (commented):
Some nits, but everything looks like they are working as expected!

nan-li (approved):

Issue #1827, opened 2023-08-31
[Bug]: No Activity found to handle Intent for invalid url


osmanowadim:
### What happened?

After sending push notifications with invalid link - "ttps://google.com" app started crashing.
Maybe need handle invalid url?

### Steps to reproduce?

```Markdown
1. Send push with invalid url
2. In application click on Push Notification
3. Receive crash of application.
```

### What did you expect to happen?

Handle invalid url like in iOS SDK. Just open app in case of invalid url

### OneSignal Android SDK version

Release 4.8.6

### Android version

13, 12, 11, 10, 9, 8, 7.1, 7.0

### Relevant log output

```Shell
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.***/com.onesignal.NotificationOpenedReceiver}: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=ttps://google.com flg=0x10000000 }
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3606)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3778)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2210)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:236)
       at android.app.ActivityThread.main(ActivityThread.java:8097)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:620)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1011)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jennantilla:
Thanks for bringing this to our attention, @osmanowadim! We'll work on handling invalid urls to prevent a crash.

ilkancelikMA:
Hello, do we have any ETA on the fix for this issue?

jinliu9508:
[...]

PR #1954, opened 2024-01-05
CRASH - When the launch url is invalide

MichaelOliveiraBx:
# Description
Catch an crash when the url is invalide and add an callback in case of this error happen 

### Motivation
FIX this [issue](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1827#issue-1875268837)

### Scope
When we start the destination activity more than `JSONException` we catch `ActivityNotFoundException` and we call the listener 

## Manual testing
Tested opening a notification with an invalide url , app build with Android Studio 2023.1.1 with our app (Omada) on a Pixel 5 with Android 13.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [x] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java
@@ -31,6 +31,7 @@ import android.app.Activity;
 import android.app.AlertDialog;
 import android.app.Application;
 import android.app.NotificationManager;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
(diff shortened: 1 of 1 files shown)


MichaelOliveiraBx:
My bad I forget to add the setter for the listener, I have add it on the next commit. 
I will squash after your review.

jinliu9508:
@MichaelOliveiraBx Thank you for reporting this and uploading your work! We are currently reviewing this issue. Out of curiosity, what plans do you have in mind for handling the event when an ActivityNotFoundException is thrown?

MichaelOliveiraBx:
To be honest, log the stack trace of the exception will be enough from. But I tell that this is an information that we might want to know because it is linked to a setup problem. And so we may want to put an alert when the person has set up a bad URL on the BO. 
But if you think it is too much, I can remove it easily

nan-li:
Hi @MichaelOliveiraBx, thanks again for your PR.

An error-level log would be ideal, so we don't introduce another interface, since we don't have callbacks for other type of errors.

abdulraqeeb33:
closing this PR and solving in the below PR

https://github.com/OneSignal/OneSignal-Android-SDK/pull/2508

Issue #1837, opened 2023-09-14
[Bug]: OptOut then OptIn doesn't work, user can no longer receive notifications


fanwgwg:
### What happened?

Using 5.0.1 release.

When a user is already subscribed, calling `OneSignal.getUser().getPushSubscription().optOut()` does opt the user out as reflected on the dashboard. However, immediately calling `OneSignal.getUser().getPushSubscription().optIn()` doesn't change the opt-in status, user is still opt-out on the dashboard and cannot receive notifications.

There is a similar issue on the flutter sdk https://github.com/OneSignal/OneSignal-Flutter-SDK/issues/729.

### Steps to reproduce?

```Markdown
Described above
```

### What did you expect to happen?

User should be subscribed after calling optIn

### OneSignal Android SDK version

5.0.1

### Android version

13

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

emawby:
@fanwgwg Hello could you get logs for this by turning on verbose logging? I am not able to reproduce this behavior

emawby:
In the logs I am seeing that after optIn we do update the subscription but the enabled value is still false. In my test app the value is true. Just to confirm before calling optOut is your subscription enabled (user is able to get push)? I am wondering if there is something else about the subscription (maybe the push permission?) that is causing it to stay false

emawby:
The enabled value is calculated based off a few factors one of which is optOut/In being called. The device needs to have a valid push token, push permissions enabled, and optIn true. Since in the log optIn was set to true and the device has a token then the SDK must not recognize that push permissions are enabled. I will try requesting permissions outside of the SDK and see if that is the issue!

emawby:
@fanwgwg Got it I will try that

Munzey:
experiencing the same with `5.0.2` - app permissions are definitely on.

On a fresh install, after calling `initWithContext` on app create and I login through our login flow and call `OneSignal.login(myLoginUserId)` I can push notifications fine.

If I try `optOut()` a user on my logout flow, and then try login again and call `optIn()`, the subscription status never seems to update, and the user cannot receive pushes.
l only tried adding these after finding that when calling `OneSignal.logout()` and `OneSignal.login(myUserId)` in my respective flows wasn't working.

The only way I can get it to trigger again for the same user is to uninstall the app and reinstall, going through the initial flow again.

[...]

Issue #1824, opened 2023-08-30
[question]: Permission not granted after optIn


MarianoManzi:
### How can we help?

In my app i have a switch to enable and disable notifications the following is the listner on the switch:

```kotlin
switch.setOnCheckedChangeListener { _, isChecked ->
  if (OneSignal.Notifications.permission) {
    if (isChecked) OneSignal.User.pushSubscription.optIn() else OneSignal.User.pushSubscription.optOut()
  } else CoroutineScope(Dispatchers.IO).launch { OneSignal.Notifications.requestPermission(true) }
}
``` 

The problem is that when i call `OneSignal.User.pushSubscription.optIn()`, even if `OneSignal.Notifications.permission` is true the OneSignal console tells me thet the user hasn't granted the permission.

The `OneSignal.User.pushSubscription.optOut()` work as expected.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jennantilla:
Hi @MarianoManzi thanks for reaching out! I have not been able to reproduce this issue. Would you be able to provide the [logs](https://documentation.onesignal.com/docs/mobile-sdk#setting-the-log-level) from the device so we can take a closer look?

Thanks!

MarianoManzi:
Hi @jennantilla, thanks for the answer.

I try to give you more information. So when I call the optIn function I get the error:  `Subscription can't change notification_types to positive integer while disabled`, but as I said befor the `OneSignal.Notifications.permission` is correctly true (checking from the devices settings the notification permission are granted) and when, from the system dialog, I accept the notification permission the `onNotificationPermissionChange` is triggered with the right value.

`[DefaultDispatcher-worker-5] HttpClient: PATCH RECEIVED JSON: {"errors":[{"code":"subscription-10","title":"Subscription can't change notification_types to positive integer while disabled"}]}`

`Operation execution failed without retry: [{"name":"update-subscription","appId":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx","onesignalId":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx","subscriptionId":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx","type":"PUSH","enabled":false,"address":"","status":"SUBSCRIBED","id":"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"}]`

[...]

PR #1957, opened 2024-01-09
Fix: Method `optIn()` not working after `optOut()` has been called

nan-li:
# Description
## One Line Summary
Fix method `optIn()` not working after `optOut()` has been called, by maintaining the local push subscription status.

### Motivation

- Fixes https://github.com/OneSignal/react-native-onesignal/issues/1590
- Fixes https://github.com/OneSignal/OneSignal-Android-SDK/issues/1824
- Fixes https://github.com/OneSignal/OneSignal-Flutter-SDK/issues/729
- Fixes https://github.com/OneSignal/OneSignal-Android-SDK/issues/1837
- Fixes https://github.com/OneSignal/OneSignal-Flutter-SDK/issues/811

**Existing Problem:**
- If on a previous session, `optOut()` was called, that means we sent notification_types of `UNSUBSCRIBE(-2)` to the server.
- We will hydrate the push subscription with the status of `UNSUBSCRIBE(-2)` in the response on the next session.
- However, we lose the true device-level subscription status in doing so.
- Then, when `optIn()` is called, we were using `UNSUBSCRIBE(-2)` along with `optedIn` to calculate the data to send in the Update Subscription request.
- This resulted in incorrect values being sent and reports of `optIn()` method not working.

**Solution:**
- We can still hydrate the push subscription model's `optedIn` property if we see either `UNSUBSCRIBE(-2)` or `DISABLED_FROM_REST_API_DEFAULT_REASON(-30)` in the response.
- However, we will not hydrate the `model.status` from the server. When the subscription model store is replacing all of the subscriptions in the refresh user operation, keep the existing push subscription `status`.
- This way, when `optIn()` is called again, we can use this _device-level status_ with the `optedIn` boolean to calculate values to send to the server.

### Scope
- Also added parsing of `DISABLED_FROM_REST_API_DEFAULT_REASON(-30)` from the server to update local `optedIn` state
- The subscription values we send to the server are calculated by the helper method `getSubscriptionEnabledAndStatus` ([here](https://github.com/OneSignal/OneSignal-Android-SDK/blob/52987ab505816e3f42fe837f195a53ef817f94e7/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/listeners/SubscriptionModelStoreListener.kt#L58)) so the model's own properties aren't directly sent.

# Testing
## Unit testing
No unit test added, can consider adding.

## Manual testing
Android emulator API 33
- tested scenarios of `optIn()` and `optOut()` with different permissions, new session, cold starts, and logging into different users

Android emulator API 30
- tested scenarios of `optIn()` and `optOut()` with new session, cold starts, and logging into different users

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1957)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
@@ -24,6 +24,7 @@ open class SubscriptionModelStore(prefs: IPreferencesService) : SimpleModelStore
                         model.deviceOS = existingPushModel.deviceOS
                         model.carrier = existingPushModel.carrier
                         model.appVersion = existingPushModel.appVersion
+                        model.status = existingPushModel.status
                     }
                     break
                 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
@@ -105,7 +105,7 @@ internal class RefreshUserOperationExecutor(
                             SubscriptionType.PUSH
                         }
                     }
-                subscriptionModel.optedIn = subscriptionModel.status != SubscriptionStatus.UNSUBSCRIBE
+                subscriptionModel.optedIn = subscriptionModel.status != SubscriptionStatus.UNSUBSCRIBE && subscriptionModel.status != SubscriptionStatus.DISABLED_FROM_REST_API_DEFAULT_REASON
                 subscriptionModel.sdk = subscription.sdk ?: ""
                 subscriptionModel.deviceOS = subscription.deviceOS ?: ""
                 subscriptionModel.carrier = subscription.carrier ?: ""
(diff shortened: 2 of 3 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
```diff
@@ -24,6 +24,8 @@ open class SubscriptionModelStore(prefs: IPreferencesService) : SimpleModelStore
                         model.deviceOS = existingPushModel.deviceOS
                         model.carrier = existingPushModel.carrier
                         model.appVersion = existingPushModel.appVersion
+                        // Don't hydrate device-level subscription status
```
jkasten2, line 27: This comment seem redundant with `// When hydrating an existing push subscription model, use existing device properties` above.

jkasten2 (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
nan-li: That's very good point, I agree

nan-li:
Noting that our Test/build setup Android step is failing, but previous runs of this PR did succeed.
Merging while we are still investigating that fix.

PR #1970, opened 2024-01-23
Concurrent modification in event producer

jinliu9508:
# Description
Instead of iterating the subscribers list when an event.fire() is called, iterate through a copy of the list to avoid concurrent modification.

### Motivation
Fixes a bug that causes concurrent modification when calling EventProducer.fire()

# Testing
I am unable to reproduce the exception consistently.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1970)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt
@@ -60,7 +60,8 @@ open class EventProducer<THandler> : IEventNotifier<THandler> {
      */
     fun fireOnMain(callback: (THandler) -> Unit) {
         suspendifyOnMain {
-            for (s in subscribers) {
+            val localList = subscribers.toList()
+            for (s in localList) {
                 callback(s)
             }
         }
@@ -73,7 +74,8 @@ open class EventProducer<THandler> : IEventNotifier<THandler> {
      * @param callback The callback will be invoked for each subscribed handler, allowing you to call the handler.
      */
     suspend fun suspendingFire(callback: suspend (THandler) -> Unit) {
-        for (s in subscribers) {
+        val localList = subscribers.toList()
+        for (s in localList) {
             callback(s)
         }
     }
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt
```diff
@@ -45,10 +45,9 @@ open class EventProducer<THandler> : IEventNotifier<THandler> {
      * @param callback The callback will be invoked for each subscribed handler, allowing you to call the handler.
      */
     fun fire(callback: (THandler) -> Unit) {
-        synchronized(subscribers) {
-            for (s in subscribers) {
-                callback(s)
-            }
+        val localList = subscribers.toList()
```
jkasten2, line 48: Is `.toList()` thread safe? I looked into it a bit and it seems like this might not be due to the way Koltin makes a copy of it.

jinliu9508: @jkasten2 Good point! Technically, .toList() is not thread-safe, but the variable 'subscribers' is a synchronized list, preventing two threads from accessing it simultaneously. The only way to trigger the concurrent modification exception is to access the variable while it is being iterated over in a loop. Based on your comment, I've added a unit test that can throw a concurrent modification exception if any user subscribes to a handler that unsubscribes itself during the event. In this case, they are in the same thread, but unsubscribing is called while "subscribers" is iterating.

jkasten2: Awesome, good test to have going forward in our unit test suite!

jkasten2 (approved):

PR #1984, opened 2024-02-07
[4.X.X] Add setting to prefer HMS over FCM

emawby:
# Description
## One Line Summary
Read a manifest metadata boolean to allow apps to prefer HMS over FCM for devices that can receive both.

## Details
Apps can add the following to their `AndroidManifest.xml`
`<meta-data android:name="com.onesignal.preferHMS"
            android:value="true"/>`

This setting will allow devices with both HMS and FCM capabilities to use HMS rather than the default of FCM.

**I have left in the testing commits for review, but will rebase prior to merge. The only changed file is `OSUtils.java`**
### Motivation
Allows apps to message users who move in and out of zones reachable by FCM, but can always receive HMS.

### Scope
subscription type

# Testing
## Unit testing
n/a

## Manual testing
Tested on Huawei device. I have left in 3 testing commits if reviewers would like to test. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1984)

### DIFF
--- a/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
+++ b/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
@@ -369,11 +369,19 @@ class OSUtils {
       if (supportsADM())
          return UserState.DEVICE_TYPE_FIREOS;
 
-      if (supportsGooglePush())
+      boolean preferHMS = false;
+      boolean supportsHMS = supportsHMS();
+      boolean supportsFCM = supportsGooglePush();
+      
+      if (supportsFCM && supportsHMS) {
+         return preferHMS ? UserState.DEVICE_TYPE_HUAWEI: UserState.DEVICE_TYPE_ANDROID;
+      }
+
+      if (supportsFCM)
          return UserState.DEVICE_TYPE_ANDROID;
 
       // Some Huawei devices have both FCM & HMS support, but prefer FCM (Google push) over HMS
-      if (supportsHMS())
+      if (supportsHMS)
          return UserState.DEVICE_TYPE_HUAWEI;
 
       // Start - Fallback logic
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/src/main/java/com/onesignal/OSUtils.java
```diff
@@ -369,11 +370,20 @@ int getDeviceType() {
       if (supportsADM())
          return UserState.DEVICE_TYPE_FIREOS;
 
-      if (supportsGooglePush())
+      Context context = OneSignal.appContext;
+      boolean preferHMS = context != null && OSUtils.getManifestMetaBoolean(context, PREFER_HMS_METADATA_NAME);
```
jkasten2, line 374: nit, this will be a bit more readable if we move these two lines into the `if (supportsFCM && supportsHMS) {` block.

jkasten2 (commented):
What happens if the device already has a player with a FCM token before this change?
I would expect we will update the player record with the new `identifier` (pushToken) and `device_type` but is this something we have tested that works?

emawby:
> Member

The token is updated, but the device type is not updated properly so the device cannot receive push. This will need a fix.

emawby:
> > Member
> 
> The token is updated, but the device type is not updated properly so the device cannot receive push. This will need a fix.

This issue has been addressed now!

jkasten2 (approved):

PR #1987, opened 2024-02-09
Update firebase-messaging version to 23.4.0

jinliu9508:
# Description
## One Line Summary
Update prefer version of firebase-messaging to 23.4.0

### Motivation
Several users have reported that updating the version of firebase-messaging resolves the 'Subscription can't change notification_types to positive integer while disabled' issue. Although a more precise cause is still under investigation, this change doesn't seem to break any existing functionality.

# Testing
## Manual testing
I am unable to reproduce the issue, as it seems to occur only for certain users, and the environment data is unclear at this moment.
I have tested push notification sent from the portal, push notification with the app closed, push notification on new install, and push notification click. The test was conducted on pixel 7 emulator with api 33. Everything works as expected. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/1987)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/build.gradle
+++ b/OneSignalSDK/onesignal/notifications/build.gradle
@@ -74,8 +74,8 @@ dependencies {
     // firebase-messaging:23.0.0 incoporates fix for SecurityException: Not allowed to bind to service
     api('com.google.firebase:firebase-messaging') {
         version {
-            require '[19.0.0, 23.0.99]'
-            prefer '23.0.8'
+            require '[19.0.0, 23.4.99]'
+            prefer '23.4.0'
         }
     }
 


OneSignalSDK/onesignal/notifications/build.gradle
```diff
@@ -74,8 +74,8 @@ dependencies {
     // firebase-messaging:23.0.0 incoporates fix for SecurityException: Not allowed to bind to service
     api('com.google.firebase:firebase-messaging') {
         version {
-            require '[19.0.0, 23.0.99]'
-            prefer '23.0.8'
+            require '[19.0.0, 23.99.99]'
```
jkasten2, line 77: I would recommend bumping only the minor version number in this version range. We have seen Firebase and Google break things in minor version bumps in the past.  So thinking this should be `[19.0.0, 23.4.99]`.

jkasten2 (changes requested):
Left a comment on the `require` version range.

Also can you note what was tested and on what device / android version in the PR?
* Testing registering on a new install and that notifications display is normally all we need to test when updating the FCM library.

OneSignalSDK/onesignal/notifications/build.gradle
jinliu9508: That makes sense, I pushed the change just now.

jinliu9508:
> Left a comment on the `require` version range.
> 
> Also can you note what was tested and on what device / android version in the PR?
> 
> * Testing registering on a new install and that notifications display is normally all we need to test when updating the FCM library.

Ok, I have updated the PR description with the testing environment and the testing units I have done for this PR.

jkasten2 (approved):

PR #2003, opened 2024-02-23
[Abandoned][Tests] Allow JSONObject Kotlin extension functions to work in tests

jkasten2:
# Description
## One Line Summary
Allow tests to utilize JSONObject Kotlin extension functions defined in our `JSONObjectExtensions.kt`.

### Motivation
We have a number of useful Kotlin extension functions `JSONObjectExtensions.kt` and it would be helpful to use them in tests.

### Scope
Only fixes issue with JSONObject in tests.

# Testing
## Unit testing
Added unit test to ensure our `JSONObject` extension functions will work in other tests.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] I have filled out all **REQUIRED** sections above
   - [X] PR does one thing
   - [X] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
   - [X] I have included test coverage for these changes, or explained why they are not needed
   - [X] All automated tests pass, or I explained why that is not possible
   - [X] I have personally tested this on my device, or explained why that is not possible

## Final pass
   - [X] Code is as readable as possible.
   - [X] I have reviewed this PR myself, ensuring it meets each checklist item

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2003)

PR #2004, opened 2024-02-24
Fix JSON Kotlin Extension Functions

jkasten2:
# Description
## One Line Summary
Fixes our JSON Kotlin extension functions and added tests to ensure our JSON implementation in tests stay accurate.

## Details
This PR sloves 3 issues with JSONArray / JSONArray:
1. Tests were running with a different implementation than Android's
   * Swapping out `org.json:json` for `com.tdunning:json` fixed this.
      - Yes this is confusing, but `org.json:json` is NOT the same library as Android's `org.json` namespace!
2. Our `JSONObject.toMap()` Kotlin extension function didn't support nested JSON.
   * It isn't clear if anything in our SDK was affected by this today, but can bite us in the future.
3. Could not use JSONObject Kotlin extension functions defined in src in tests using Robolectric
   * Was only an issue with `org.json:json` for some reason, so swapping `com.tdunning:json` also fixed this issue.

This PR is based on some of the attempted work from PR #2003.

### Motivation
`JSONObject` and `JSONArray` are two APIs we use a lot in this SDK. It is important they are working fully and we are testing with the same implementation in tests.

### Scope
Effects code that uses `JSONObject.toMap()`. And tests that used `org.json:json`.

### History 
Fun fact, Google rolled [their own org.json](https://cs.android.com/android/platform/superproject/main/+/main:libcore/json/src/main/java/org/json/JSONObject.java;l=34?q=jsonobject) for Android. They probably did this due to the silly "The Software shall be used for Good, not Evil." line in [the orignal org.json licence](https://www.json.org/license.html).
* The original [org.json](https://github.com/stleary/JSON-java/issues/756) is now public domain.

# Testing
## Unit testing
New [JSONObjectEnvTest.kt](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2004/files#diff-5e066b1919d5f409aca7a1d25c687ac65a15972f75c62e871c978f49bd73d045) and [JSONObjectExtensionsTest.kt ](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2004/files#diff-b81646a50d3df8a1625e919a0c46d32ce5cbc6610e319ad1a0037f1016792efd) test files were added to cover the behavior changes and ensure test env does not regress.

## Manual testing
Tested on an Android 14 emulator ensuring the device tag updates are sent.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2004)

### DIFF
--- a/OneSignalSDK/build.gradle
+++ b/OneSignalSDK/build.gradle
@@ -18,6 +18,8 @@ buildscript {
         ktlintPluginVersion = '11.6.1'
         ktlintVersion = '1.0.1'
         junitVersion = '4.13.2'
+        // DO NOT upgrade for tests, using an old version so it matches AOSP
+        tdunningJsonForTest = '1.0'
     }
 
     repositories {
--- a/OneSignalSDK/onesignal/core/build.gradle
+++ b/OneSignalSDK/onesignal/core/build.gradle
@@ -88,8 +88,10 @@ dependencies {
     testImplementation("androidx.test:core-ktx:1.4.0")
     testImplementation("androidx.test:core:1.4.0")
     testImplementation("io.mockk:mockk:1.13.2")
-    testImplementation("org.json:json:20180813")
     testImplementation("org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion")
+
+    // com.tdunning:json is needed for non-Robolectric tests.
+    testImplementation("com.tdunning:json:$tdunningJsonForTest")
 }
 
 ktlint {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/JSONObjectExtensions.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/JSONObjectExtensions.kt
@@ -2,6 +2,7 @@ package com.onesignal.common
 
 import org.json.JSONArray
 import org.json.JSONObject
+import org.json.JSONObject.NULL
 
 /**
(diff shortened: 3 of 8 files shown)


nan-li (approved):
How did you know to make the changes in this PR? 
Was something failing, or did you experience some strange behavior during testing?

jkasten2:
The first thing that got me started in looking into this was I tried to use `JSONObject.toMap()` in a test to make it easier to read the test.However it would fail when running the test with it not being able to find it `.toMap()`

I then uncovered these problems:
1. The `org.json:json` library causing Kotlin Extension methods on JSONObject to not be found, but only if the test uses Robolectric 
   * I attempted to fix this by not `org.json:json` when Robolectric is used (in PR  #2003). However I discovered that some tests failed due to behavior differences in the implementation of some JSONObject methods, specifically `JSONObject.getString` can throw in `org.json:json`, but AOSP just toString it for you if it is an nest JSONObject.
2. Our `JSONObject.toMap()` wasn't implemented correctly, as it didn't supported nested JSONObject and JSONArray like other libraries do.

Issue #1995, opened 2024-02-15
[Bug]: 5.1.4 - NullPointerException: null cannot be cast to non-null type kotlin.String crashes on Startup


CarlBarks:
### What happened?

We experienced a NullPointer on initialisation which causes the app to crash on startup. We are on `5.1.4`

Might be related to https://github.com/OneSignal/OneSignal-Android-SDK/issues/1990 but the logoutput there suggests a crash at a different place

Seems more related to https://github.com/OneSignal/OneSignal-Android-SDK/issues/1745 however, that was closed in March 2023.

### Steps to reproduce?

```Markdown
Not easily reproducible 

1. OneSignal Android SDK `5.1.4`
2. `initWithContext` 
3. Experience crash on Crashlytics
```

### What did you expect to happen?

OneSignal to get initialised without issues across all users and devices

### OneSignal Android SDK version

5.1.4

### Android version

9

### Relevant log output

```
Fatal Exception: java.lang.RuntimeException
Unable to create application xxx: java.lang.NullPointerException: null cannot be cast to non-null type kotlin.String
```

```
Caused by java.lang.NullPointerException
null cannot be cast to non-null type kotlin.String
com.onesignal.common.modeling.Model.getStringProperty (Model.kt:491)
com.onesignal.common.modeling.Model.getStringProperty$default (Model.kt:488)
com.onesignal.user.internal.identity.IdentityModel.getOnesignalId (IdentityModel.kt:19)
com.onesignal.user.internal.operations.impl.listeners.SubscriptionModelStoreListener.getUpdateOperation (SubscriptionModelStoreListener.kt:48)
com.onesignal.user.internal.operations.impl.listeners.SubscriptionModelStoreListener.getUpdateOperation (SubscriptionModelStoreListener.kt:15)
com.onesignal.core.internal.operations.listeners.ModelStoreListener.onModelUpdated (ModelStoreListener.kt:52)
com.onesignal.common.modeling.ModelStore$onChanged$1.invoke (ModelStore.kt:91)
com.onesignal.common.modeling.ModelStore$onChanged$1.invoke (ModelStore.kt:91)
com.onesignal.common.events.EventProducer.fire (EventProducer.kt:50)
com.onesignal.common.modeling.ModelStore.onChanged (ModelStore.kt:91)
com.onesignal.common.modeling.Model$notifyChanged$1.invoke (Model.kt:666)
[...]

PR #2009, opened 2024-02-26
Migration: Missing onesignalId during initWithContext 

jinliu9508:
# Description
## One Line Summary
Add default subscription status if legacy user info was previously saved on the device.

### Motivation
We've received crash reports from this thread https://github.com/OneSignal/OneSignal-Android-SDK/issues/1995, indicating a crash caused by a null pointer exception when calling onesignal_id during the initWithText process. After some research, I discovered that this only occurs when the push subscription status is missing during a migration from v4 to v5. The getter of the subscription status will attempt to make a change to the model, which will then be queued in the model update operation. This operation requires onesignal_id to be completed, but the id is null at the early stage. To prevent the access of onesignal_id when it is null in the early init process, this fix will add a default subscription status in case of its absence under the NO_PROPOGATE tag.

# Testing
## Manual testing
Step to reproduce:
  1. Perform a fresh install of v4 on an Android device. (I am using Pixel 3a API 34)
  2. Make sure the player is created and shown on the OneSignal dashboard.
  3. Check out main and comment out the code that adds push subscription status when notification_types is null.
  4. Build and run the app on the same device. Notice that a null pointer exception is reported due to onesignal Id being null.

After the fix, the crash no longer exists. Some normal OneSignal operations such as logging in, logging out, sending and receiving push notifications are all working properly.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2009)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -289,6 +289,8 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                             legacyUserSyncJSON.safeString("identifier") ?: ""
                         if (notificationTypes != null) {
                             pushSubscriptionModel.status = SubscriptionStatus.fromInt(notificationTypes) ?: SubscriptionStatus.NO_PERMISSION
+                        } else {
+                            pushSubscriptionModel.status = SubscriptionStatus.SUBSCRIBED
                         }
 
                         pushSubscriptionModel.sdk = OneSignalUtils.SDK_VERSION


nan-li:
Can you add more context / background in the PR, and what the stacktrace is, or linking to the GH issue related to this fix.

- What was the root cause of this crash and how did this PR fix it? 
- How does making sure there is a defalut subscription status prevent the onesignal ID from being null?
- Is the onesignal_id now no longer `null` at the place where it used to crash, or is it still null but no longer accessed?
- etc.

jinliu9508:
> Can you add more context / background in the PR, what the stacktrace is, linking to the GH issue related to this fix.
> 
> * What was the root cause of this crash and how did this PR fix it?
> * How does making sure there is a default subscription status prevent the onesignal ID from being null?
> * Is the onesignal_id now no longer `null` at the place where it used to crash, or is it still null but no longer accessed?
> * etc.

Good questions! I have updated the description with better explanations for why the error occurs and how the fix helps to avoid it. I hope this provides a better understanding of the PR.

nan-li (approved):

PR #2017, opened 2024-03-05
[Feat] Add backoff to OperationRepo when retrying network calls

jkasten2:
# Description
## One Line Summary
Add a delay to retries with a backoff to slow down network calls from `OperationRepo`.

### Motivation
Retrying too quickly:
* puts unnecessary load on the device and OneSignal's backend.
* may cause multiple OneSignal Users being created
    - some of them being in a bad state with the same `external_id` but different `onesignal_id` values due to this fast retrying.
    - Mostly only an issue when `OperationRepo.force` is true, which causes the 5 second batching to delay to be skipped, which normally get set when calling `OneSignal.login`

### Scope
Only affects adding a delay to retries on `OperationRepo`.

# Testing
## Unit testing
To ensure this works as expected we [updated the retrying test to confirm the delay is called](https://github.com/OneSignal/OneSignal-Android-SDK/compare/feat/add_backoff_to_operation_retries?expand=1#diff-7fa744405d396861944bf74935126dad924f9dfd24d3dda580a8d8a6461687d9R105).

## Manual testing
Also to ensure this commit works as intended I ran this on an Android 14 emu and ensured it waits the right amount of time when there is `500` response. Also after we stop getting a `500` error the queue successfully resumes and all pending tags are even added too. While it's waiting for the next retry I ensured it didn't lock main thread as well.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [X] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2017)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -25,6 +25,7 @@ internal class OperationRepo(
     private class OperationQueueItem(
         val operation: Operation,
         val waiter: WaiterWithValue<Boolean>? = null,
+        var retries: Int = 0,
     )
 
     private val executorsMap: Map<String, IOperationExecutor>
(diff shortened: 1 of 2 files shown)


nan-li (approved):
LGTM

- In my testing, without this PR, retries happen every 5 seconds, so they weren't so immediate as is.
- I do think there are some code paths that do because I have seen my logcat spamming retries in the past.

jkasten2:
@nan-li Thanks for the review!

> - In my testing, without this PR, retries happen every 5 seconds, so they weren't so immediate as is.
> - I do think there are some code paths that do because I have seen my logcat spamming retries in the past.

In my testing there are cases where `OperationRepo.force` gets set to `true` and this skips the 5 second batching logic. This normally happens when OneSignal.login is called.

PR #2033, opened 2024-03-20
[Improvement] Prevent OperationRepo from continuously pulling when empty

jkasten2:
# Description
## One Line Summary
Prevent `OperationRepo` from continuously pulling when empty, which might be putting extra load on the device.

### Motivation
The pulling might be causing some battery drain due to the app's process never fully sleeping or CPU continuing to be woken up. However the theory is this may only be noticable factor on some devices (probably more noticeable on watches running Android). 

### Scope
Only changes to the `OperationRepo` polling logic.

# Testing
## Unit testing
Added a new test to confirm it continuously pulls when the operations list is empty.

## Manual testing
Test setup
* Device = LGL58VL
* Android Version = 6.0.1
* Phone mode: Airplane mode on, with WiFi on

Tested the battery drain by unplugging the device for a long time then obserserving how much power it takes to rechanged the device to full again. Recharge was measured with a USB multiple meter, getting the Wh it took to recharge the device until it stopped accepting power.

We test 3 scenarios, baseline to see how much power the device loses over time without OneSignal, and then two more comparing before after. Results are as follows:

**Baseline (OneSignal app not installed on device)**
* Unplugged for ~1 hour, recharged 0.12Wh in 11:17 = Drain of .12Wh
* Unplugged for ~19 hours, recharged 0.77Wh in 57:56 = Drain of 0.0405Wh

**Test 1 - 5.1.6**
* Unplugged for ~1 hour, recharged.14Wh in 20:06 = Drain of .14Wh
* Unplugged for ~1 hour, recharged .09Wh in 8:50 = Drain of .09Wh
* Unplugged for ~50 minutes, recharged .10Wh in 12:57 = Drain of .12Wh
* Unplugged for ~23 hours, recharged .85Wh in 54:41 = Drain of 0.0370Wh

**Test 2 - 5.1.6 (with OperationRepo fix)**
* Unplugged for ~1 hour recharged .05Wh in 8:03 = Drain of .05Wh
* Unplugged for ~1 hour, recharged 0.12Wh in 9:38 = Drain of .12Wh
* Unplugged for ~6 hours, recharged .20Wh in 22:32 = Drain of .0333Wh
* Unplugged for ~84 hours, recharged 1.99Wh in 1:24:29  = Drain of .023Wh

**Summary of testing**
The data is inconclusive, unfortunately this method of testing is resulting in wildly different numbers between tests, even tests that last 24 hours. There stand to be some theory improvement still from this PR, but a different testing strategy and/or different devices would have to be tested.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2033)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -97,53 +97,37 @@ internal class OperationRepo(
      * dedicated thread.
      */
     private suspend fun processQueueForever() {
-        var lastSyncTime = _time.currentTimeMillis
-        var force = false
-
-        // This runs forever, until the application is destroyed.
+        var force = waitForWake()
         while (true) {
             if (paused) {
                 Logging.debug("OperationRepo is paused")
                 return
             }
-            try {
-                val ops = getNextOps()
 
-                // if the queue is empty at this point, we are no longer in force flush mode. We
-                // check this now so if the execution is unsuccessful with retry, we don't find ourselves
-                // continuously retrying without delaying.
-                if (queue.isEmpty()) {
-                    force = false
-                }
+            val ops = getNextOps()
+            Logging.debug("processQueueForever:force:$force, ops:$ops")
 
-                if (ops != null) {
-                    executeOperations(ops!!)
-                }
-
-                if (!force) {
-                    // potentially delay to prevent this from constant IO if a bunch of
-                    // operations are set sequentially.
-                    val newTime = _time.currentTimeMillis
-
-                    val delay = (lastSyncTime - newTime) + _configModelStore.model.opRepoExecutionInterval
-                    lastSyncTime = newTime
-                    if (delay > 0) {
-                        withTimeoutOrNull(delay) {
-                            // wait to be woken up for the next pass
-                            force = waiter.waitForWake()
-                        }
-
-                        // This secondary delay allows for any subsequent operations (beyond the first one
-                        // that woke us) to be enqueued before we pull from the queue.
-                        delay(_configModelStore.model.opRepoPostWakeDelay)
-
-                        lastSyncTime = _time.currentTimeMillis
-                    }
-                }
-            } catch (e: Throwable) {
-                Logging.log(LogLevel.ERROR, "Error occurred with Operation work loop", e)
+            if (ops != null) {
+                executeOperations(ops)
+                // Allows for any subsequent operations (beyond the first one
+                // that woke us) to be enqueued before we pull from the queue.
+                delay(_configModelStore.model.opRepoPostWakeDelay)
+            } else if (!force) {
+                force = waitForWake()
+            } else {
+                force = false
             }
         }
+    }
+
+    private suspend fun waitForWake(): Boolean {
+        var force = waiter.waitForWake()
+        if (!force) {
+            withTimeoutOrNull(_configModelStore.model.opRepoExecutionInterval) {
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -97,61 +97,47 @@ internal class OperationRepo(
...
+     *  amount of time afterwards, so operations can be grouped/batched.
+     */
+    private suspend fun waitForNewOperationAndExecutionInterval() {
+        // 1. Wait for an operation to be enqueued
+        var force = waiter.waitForWake()
+
+        // 2. Wait at least the time defined in opRepoExecutionInterval
+        //    so operations can be grouped, unless one of them used
+        //    flush=true (AKA force)
+        val startTime = _time.currentTimeMillis
```
nan-li, line 132: Wondering if I missed something about this logic, technically `startTime` is not necessary since it is never updated?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -97,61 +97,47 @@ internal class OperationRepo(
...
+                waitForNewOperationAndExecutionInterval()
+            }
+        }
+    }
 
-                        lastSyncTime = _time.currentTimeMillis
-                    }
-                }
-            } catch (e: Throwable) {
-                Logging.log(LogLevel.ERROR, "Error occurred with Operation work loop", e)
```
nan-li, line 153: This try-catch is removed because it does not seem to have a specific purpose or thing it is meant to catch right?

nan-li (approved):

nan-li:
> A longer multiple day test is being done to confirm the results, will update PR afterwards with the results.
Was there an outcome to this?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
jkasten2: ah good point, I removed it in a `--fixup` commit just now. I think I refactored this a few times and forgot I didn't need `startTime` any more.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
jkasten2: Ya, there isn't anything that I can see that could throw, and if it does we want to know instead of silently moving on. Also note I left the try-catch on `executeOperations` since it wasn't in scope of this PR.

jkasten2:
> > A longer multiple day test is being done to confirm the results, will update PR afterwards with the results.

> Was there an outcome to this?

After some longer tests it seems inconclusive now, this testing strategy vary to much between tests.

Don't think it worth the time to try other tests right now, there is going to be a lot of other improvements in the upcoming 5.1.7 release (such as retrying with backoff) that might be the root source of some reported battery drain.

PR #2036, opened 2024-03-27
[Improvement] limit refresh User and GET IAMs to foreground

jkasten2:
# Description
## One Line Summary
Limit refresh User and GET IAMs to foreground to save resources.

### Motivation
There are a number of background events that can start an app in the background (example push received), however were wasting resources (both on the device and OneSignal's backend) if the app is never foregrounded.

### Scope
Only affect when we fetch IAMs and when refresh the User cache.

# Testing

## Manual testing
Tested on an Android 14 emulator, ensure we don't fetch IAMs and call GET User until the app is foregrounded.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [X] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2036)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -323,12 +323,6 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                 }
             } else {
                 Logging.debug("initWithContext: using cached user ${identityModelStore!!.model.onesignalId}")
-                operationRepo!!.enqueue(
-                    RefreshUserOperation(
-                        configModel!!.appId,
-                        identityModelStore!!.model.onesignalId,
-                    ),
-                )
             }
 
             startupService!!.start()
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
@@ -4,6 +4,7 @@ import com.onesignal.common.modules.IModule
 import com.onesignal.common.services.ServiceBuilder
 import com.onesignal.core.internal.operations.IOperationExecutor
 import com.onesignal.core.internal.startup.IBootstrapService
+import com.onesignal.core.internal.startup.IStartableService
 import com.onesignal.user.internal.UserManager
 import com.onesignal.user.internal.backend.IIdentityBackendService
 import com.onesignal.user.internal.backend.ISubscriptionBackendService
@@ -24,6 +25,7 @@ import com.onesignal.user.internal.operations.impl.listeners.IdentityModelStoreL
 import com.onesignal.user.internal.operations.impl.listeners.PropertiesModelStoreListener
 import com.onesignal.user.internal.operations.impl.listeners.SubscriptionModelStoreListener
 import com.onesignal.user.internal.properties.PropertiesModelStore
+import com.onesignal.user.internal.service.UserRefreshService
 import com.onesignal.user.internal.subscriptions.ISubscriptionManager
 import com.onesignal.user.internal.subscriptions.SubscriptionModelStore
 import com.onesignal.user.internal.subscriptions.impl.SubscriptionManager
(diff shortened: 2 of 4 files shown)


jinliu9508 (approved):
Tested with Android emulator Pixel 3a API 34 that when the app is swiped up (in background,) the app does not call to refresh user.

nan-li:
I'm running into wonky behavior testing this PR. 

1. View IAMs during the session
2. Swipe kill app away and it logs `OSBackgroundSync scheduleSyncServiceAsJob:atTime: 30000`
3. About 50 seconds later, the SDK initializes (noticed it does not call fetch IAM or get user)
4. It tries to send session time for the recently ended session
```
D  [Thread-15] SessionService.backgroundRun()
D  [Thread-15] SessionService: Session ended. activeDuration: 0
D  [Thread-15] OperationRepo.enqueue(operation: {"name":"track-session-end","appId":"77e32082-ea27-42e3-a898-c72e141824ef","onesignalId":"c65e624d-ea2b-437c-a663-3e3da9d2d3b0","sessionTime":0}, flush: false)
```

nan-li:
I'm running into wonky behavior testing this PR when session time is sent after the app is swiped away.

1. View IAMs during the session.
2. Swipe kill app away and it logs `OSBackgroundSync scheduleSyncServiceAsJob:atTime: 30000`.
3. About 50 seconds later, the SDK initializes (noticed it does not call fetch IAM or get user).
4. It calculates session time wrong as `0` which is the value sent to POST outcomes and PATCH user.
```
D  [Thread-15] SessionService.backgroundRun()
D  [Thread-15] SessionService: Session ended. activeDuration: 0
D  [Thread-15] OperationRepo.enqueue(operation: {"name":"track-session-end","appId":"<id>","onesignalId":"<id>","sessionTime":0}, flush: false)

D  [DefaultDispatcher-worker-7] HttpClient: POST outcomes/measure - FAILED STATUS: 400
W  [DefaultDispatcher-worker-7] HttpClient: POST RECEIVED JSON: {"errors":["There was a problem in the JSON you submitted","\"session_time\" must be positive when submitting os__session_duration"],"success":false}
W  [Thread-16] OutcomeEventsController.sendAndCreateOutcomeEvent: Sending outcome with name: os__session_duration failed with status code: 400 and response: {"errors":["There was a problem in the JSON you submitted","\"session_time\" must be positive when submitting os__session_duration"],"success":false}
Outcome event was cached and will be reattempted on app cold start


D  [OpRepo] UpdateUserOperationExecutor(operation: [{"name":"track-session-end","appId":"<id>","onesignalId":"<id>","sessionTime":0,"id":"<id>"}])
D  [DefaultDispatcher-worker-7] HttpClient: PATCH apps/<id>/users/by/onesignal_id/<id> - {"refresh_device_metadata":false,"deltas":{"session_time":0}}
D  [DefaultDispatcher-worker-7] HttpClient: PATCH apps/<id>users/by/onesignal_id/<id> - STATUS: 202 JSON: {"properties":{}}
```

jkasten2:
@nan-li Hmm, I am not able to reproduce this bug from some reason. I tried it a few times on an Android 14 emulator, both as a clean install and and existing install. It looks like you are using the default app_id we have in our example app for the 2nd test (ending in `-c72e141824ef`) did you use the same one in that first test? Can you also consistently reproduce the issue?

nan-li:
I had tested two fresh installs on Android 13 emulator that reproduced the session time of 0 when I posted the comment above. Then another fresh install without the changes in this PR. All with same default app_id.

But now, trying it again, I am still running new installs and only reproduced it in 1 out of 5 new app installs. Might be random and not related to the changes in this PR.

jkasten2:
From looking at the session code I don't see how this PR could be related to the issue, so going to merge this PR in.

Good find on this bug though, I think it might be related to the time not always getting written to disk before the app is killed. 
[SessionService.onUnfocused](https://github.com/OneSignal/OneSignal-Android-SDK/blob/79a2fbdb18fe6f6998f23617b5bcae0332020689/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionService.kt#L94) writes this time value, but that is buffered through [PreferencesService](https://github.com/OneSignal/OneSignal-Android-SDK/blob/b44b4898ff0751913e3ebfa4e5608c02f4516793/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/impl/PreferencesService.kt#L4) so it may not be written to disk in time. 
Another odd thing about this is we are sending outcomes for IAMs, but the backend doesn't support that.

PR #2037, opened 2024-03-28
[Fix] optIn() not prompting if called before push subscription is created on backend

jkasten2:
# Description
## One Line Summary
Fix `optIn()` not prompting if called before push subscription is created on backend .

### Motivation
It's common for developers to call `OneSignal.User.pushSubscription.optIn()` and it should prompt even if the device is offline or has connection issues.

### Scope
Effects internal push subscription property change events.

# Testing
## Unit testing
Created a new unit test to cover this case.
* Ensured it failed before my change, then passes after my change.

## Manual testing
Tested on a real Samsung S21 with Android 14 (security patch level Feb 1, 2024). Clean installed example app while the device was in airplane mode, ensuring calling `optIn()` now prompts. Also tested adding an email and sms when the device was offline and then when I turn back on network connect it successfully created all the subscriptions as expected.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2037)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
@@ -157,7 +157,10 @@ internal class SubscriptionManager(
         args: ModelChangedArgs,
         tag: String,
     ) {
-        val subscription = subscriptions.collection.firstOrNull { it.id == args.model.id }
+        val subscription =
+            subscriptions.collection.firstOrNull {
+                args.model == (it as Subscription).model
+            }
 
         if (subscription == null) {
             // this shouldn't happen, but create a new subscription if a model was updated and we
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/IDManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/IDManager.kt
@@ -7,7 +7,7 @@ import java.util.UUID
  * and detect whether a provided ID was generated locally.
  */
 object IDManager {
-    private const val LOCAL_PREFIX = "local-"
+    internal const val LOCAL_PREFIX = "local-"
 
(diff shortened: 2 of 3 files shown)


jinliu9508 (approved):
LGTM!

It's weird that the push subscription id was compared to the model id the whole time. I guess the subscription id is updated to its model's id at some point so we didn't see this happening all the time.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
```diff
@@ -157,7 +157,10 @@ internal class SubscriptionManager(
         args: ModelChangedArgs,
         tag: String,
     ) {
-        val subscription = subscriptions.collection.firstOrNull { it.id == args.model.id }
```
nan-li, line 160: I see, to use `id` as a getter, it would have to compare `it.model.id == args.model.id`

nan-li (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
jkasten2: Ya, since the [Susbcription.id](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.1.6/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/Subscription.kt#L14) returns `""` if it starts with `local-` it was not finding it

Issue #2043, opened 2024-04-04
[Bug]: Using OneSignal.login() does not set externalId in Subscription Records


sebastinto:
### What happened?

Calling `OneSignal.login()` does not seem to be updating Subscription Records i.e. the External ID column values are unset. `OneSignal.User.externalId` is correctly set on device.

I tried on a fresh install, multiple device, calling `OneSignal.logout()` before `OneSignal.login()` to no avail. I'm wondering if there's anything I may be missing.

Tags are set correctly.

### Steps to reproduce?

```Markdown
1. Initialize SDK
2. Call OneSignal.login() with a value (e.g. "1234")
3. Observe Subscription Records External ID does not update
```

### What did you expect to happen?

External ID updates in Subscription Records.

### OneSignal Android SDK version

Release 5.1.7

### Android version

13

### Specific Android models

```Markdown
* Pixel 8 Pro
* Samsung S20 5G
```

### Relevant log output

[os_log.txt](https://github.com/OneSignal/OneSignal-Android-SDK/files/14875464/os_log.txt)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sebastinto:
Thanks @jkasten2 !

> 1. Where are you calling `OneSignal.initWithContext` and `OneSignal.login` in your code?

`OneSignal.initWithContext` is called in the `onCreate()` function of the main `Application` class. `OneSignal.login` is called after a user successfully authenticates via our API.

> 2. "Observe Subscription Records External ID does not update" where are you observing this? On the OneSignal dashboard?

Correct!

> 3. Does the issue happen if you simply call `OneSignal.initWithContext` then  `OneSignal.login`?
>    * Or does the issue only happen if you call logout()?

Removing `logout()` fixes the issue.

I have a few follow-up questions if you don't mind:

1. Is this expected behavior? Why would calling `logout()` prevent subsequent `login()` calls?
2. It seems like calling `login()` with a different value is enough to override external id. Is this the recommended implementation? 
3. If it is, what would be the use case for calling `logout()`? Explicitly de-identifying a user? How would you re-identify a user after calling `logout()`?

jkasten2:
> Removing logout() fixes the issue.

Good to hear! However that is bug as `logout()` shouldn't cause an issue like this. Could you explain step-by-step what the order of timing of the login() and logout() calls that cause the issue? Include any delays or if they are called back-to-back.

[...]

Issue #1947, opened 2023-12-19
[Bug]: In v5, the Subscription Status does not change to 'Subscribed'.


kishibata:
### What happened?

<img width="1135" alt="スクリーンショット 2023-12-19 11 19 35" src="https://github.com/OneSignal/OneSignal-Android-SDK/assets/7996578/3f8154ae-c2e6-46a6-98a4-22dddbdc2157">

In v5, even if permission is granted on the OS notification permission dialog, the Subscription Status does not change to 'Subscribed.'

### Steps to reproduce?

```Markdown
1. Turn off notification settings on the device.
2. Launch the app.
3. In the OS permission dialog for notifications, grant permission.
```

### What did you expect to happen?

At the moment when permission is granted on the OS notification permission dialog, the Subscription Status changes to 'Subscribed.'

### OneSignal Android SDK version

5.0.5

### Android version

13

### Specific Android models

```Markdown
Pixel 6a : Android14
Pixel ４ : Android13
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

emawby:
@kishibata Thank you for reaching out I have not been able to reproduce this. If possible could you share verbose logs when permission is granted so that we can try to debug this behavior?

kishibata:
@emawby 
Thank you for your confirmation. I will share the logs at the following link.

https://docs.google.com/document/d/11Z1-UFyvUOGAmuZhiXKFPDdsFSc7TJKahRRue3-C9Q8/edit?usp=sharing

kishibata:
@emawby 
any update on this issue?

emawby:
Hello @kishibata thank you for the reminder. We believe we know the root cause of the issue and are working on a fix, but unfortunately it has not been resolved yet. To confirm our investigations have you noticed this happening if you don't call "Login" for the device? You may need to logout/reinstall to remove the cached user.

kishibata:
@emawby 
Hello, thank you for your message. I'm not quite clear on the meaning of 'Login for the device.' Could you please provide more details or clarify its significance in this context? Thank you.

kishibata:
@emawby 
Hello, and thank you for reaching out. I appreciate your update. I'm not entirely clear on the meaning of 'Login for the device'. Could you kindly provide more information or clarification about its significance in this context? Your assistance in understanding this matter is greatly appreciated. Thank you.

gabriserra:
[...]

PR #2046, opened 2024-04-08
[Fix] external_id is skipped and updates stop if something updates the User (such as addTag) shortly before login is called

jkasten2:
# Description
## One Line Summary
Fixes login silently failing (and operations after gets stuck) when login is called within 5 seconds of a user property update (such as addTag).

## Details
This fixes a bug where calling addTag() then login() would cause the login to be dropped.
The bug happen on all operations didn't have a create or modify key, this is required so operations are not grouped together when they should not be. Since we fixed batching in 5.1.7 to always wait 5 seconds this issue will be most common on 5.1.7, however all versions back to 5.0.0 are effected.

### Motivation
It's critical that login works consistently.

### Scope
Affects operation repo grouping and error handling of executors.

### How this issue happened
This is because`SetTagOperation.modifyComparisonKey` was always `""` which matched `LoginUserOperation.modifyComparisonKey` as it was also `""` and both these operations were returned by `OperationRepo.getGroupableOperations`. Both operations were given to `UpdateUserOperationExecutor` which it would skip the `LoginUserOperation` silently.

### Recovering from stalled state
Since the bug was dropping the login operation, it needs to be replaced to get those already in the bad state out. Commit bc917772972ae5e652567d1fcb904d9516e40aff does this.

If the commit above was not done then even if the developer called OneSignal.login() again with the same `externalId` it would not correct the stalled User. Only calling logout() or login() with different `externalId` would have un-stalled the `OperationRepo`. And then only after logging back to that stalled User would it have recover all the unsent operations they may exist.

## Related PRs
This mistake originated from PR #1794, commit f1d204e3ebaff51aa8f4a7239bc681e918d75078, which was shipped in the 5.0.0 (wasn't in 5.0.0-beta4 yet). I preserved the intent of PR #1794, and fixed the mistake in commit 7fe2010dd1326b1a1e47619c0f3ee17b919aa221.

## Related Issues
* Issue #2043 - Will fix
* Issue #1947 - Possibly related, if login was called (or possibly some other rare case?)

# Testing
## Unit testing
Corrected some existing unit tests.

## Manual testing
Android 14 Emulator.
* Fresh install - addTag, login
   - The bug that initially uncovered this issue.
   - Also ensured accepting notification permission updated subscription 
* Fresh install - addTag, login, addTag
    - Ensures the keys for comparison key values are being handled correctly
* Fresh install - Add and remove an alias back-to-back
    - Confirms assumption that these are split up due to grouping
* After restarting the app - Logout and login again
* Fresh install - Device offline, accept notifications, device online, observer subscription is enabled on the dashboard.
* Fresh install - Delayed starting OperationRepo (artificially with delay) - observer subscription is enabled on the dashboard.
* Tested recovery code
   - Fresh installed app with 5.1.7
   - Reproduced stalled bug
   - Upgraded to this PR
   - Ensured it recovered and it sets the externalId
       * NOTE: There is a backend bug with POST /users, where subscription updates are not being applied so the subscription may stay disabled.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2046)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationRepo.kt
@@ -1,5 +1,7 @@
 package com.onesignal.core.internal.operations
 
+import kotlin.reflect.KClass
+
 /**
  * The operation queue provides a mechanism to queue one or more [Operation] with the promise
  * it will be executed in a background thread at some point in the future.  Operations are
@@ -31,4 +33,13 @@ interface IOperationRepo {
         operation: Operation,
         flush: Boolean = false,
     ): Boolean
+
+    /**
+     * Check if the queue contains a specific operation type
+     */
+    fun <T : Operation> containsInstanceOf(type: KClass<T>): Boolean
 }
+
+// Extension function so the syntax containsInstanceOf<Operation>() can be used over
+// containsInstanceOf(Operation::class)
+inline fun <reified T : Operation> IOperationRepo.containsInstanceOf(): Boolean = containsInstanceOf(T::class)
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -15,6 +15,7 @@ import com.onesignal.debug.internal.logging.Logging
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.withTimeoutOrNull
 import java.util.UUID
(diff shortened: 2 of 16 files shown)


nan-li (approved):
I think the commit [add operation checks to all executors](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2046/commits/80dbf09971158f11d92c762f33a04c388caf453b) points out some things we may want to address in a future PR, but this PR to address the issues it is fixing, I approve after testing on my end.

PR #2051, opened 2024-04-12
Add AndroidManifest option to override In-App Messages gray overlay and dropshadow

jennantilla:
# Description
## One Line Summary
Add AndroidManifest option to override In-App Messages gray overlay and dropshadow.

## Details
Customers would like the ability to control some In-App Message display "defaults":

- a dark gray background in center modal and full screen IAMs
- a dropshadow on all IAMs

 Customers would like the option to disable this overlay.

With this PR, developers can now add `com.onesignal.inAppMessageHideGrayOverlay` and `com.onesignal.inAppMessageHideDropShadow` to AndroidManifest.xml. Setting these values to true will not show a shadow or gray overlay. Omitting these values or setting them to false will display the default In-App design with drop shadow and gray overlay.

```
<meta-data android:name="com.onesignal.inAppMessageHideGrayOverlay" android:value="true"/> 
<meta-data android:name="com.onesignal.inAppMessageHideDropShadow" android:value="true"/>
```

**Examples without the dropshadow**
 

**Example with the dropshadow**
 

**Example with the gray overlay**
 

**Example without the gray overlay**

### Scope
New and existing In-App Message visual behavior.

### Other
In the future, we can further consider the use case of having a drop shadow _and_ transparent background:

Currently, the drop shadow is for the entire IAM, including the transparent parts. The question would be whether this is expected, or a shadow around each visible element only.

Future work could include aligning Android and iOS appearance based on whatever is deemed "expected".

# Testing
## Manual testing
Tested on all IAM types with a Pixel 7 Pro emulator running Android 14 without AndroidManifest meta-data, with meta-data set to `true`, and with meta-data set to `false`.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2051)

### DIFF
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/InAppMessageView.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/InAppMessageView.kt
@@ -49,6 +49,7 @@ internal class InAppMessageView(
     private var webView: WebView?,
     private val messageContent: InAppMessageContent,
     private val disableDragDismiss: Boolean,
+    private val hideGrayOverlay: Boolean,
 ) {
     private var popupWindow: PopupWindow? = null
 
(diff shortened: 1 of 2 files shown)


OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/display/impl/InAppMessageView.kt
```diff
@@ -626,11 +641,17 @@ internal class InAppMessageView(
                 }
             }
 
+        var overlayColor = Color.parseColor("#BB000000")
+
```
nan-li, line 645: Since this block of code repeats twice, we can make it into a method and call that method when we need to get the color.

`Color.parseColor("#00000000")` below is also the transparent color. We can consolidate into one variable and prefer `Color.TRANSPARENT` for clarity.

nan-li (changes requested):

nan-li (approved):

jkasten2 (approved):

PR #2052, opened 2024-04-12
Fix for WorkManager not Initialized Crash 

nan-li:
# Description
## One Line Summary
This is a fix for the non-reproducible WorkManager is not Initialized Crash we did in `v4` already, with one small change in a try-catch.

### Motivation
Based on the the fix we made for `v4` at #1721 

The change I added was an additional try-catch after [this comment](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1672#issuecomment-1720744291) from an SDK user after we made the `v4` fix. They ran into the "WorkManager is already initialized" exception.

At the time that PR was made, the fixes were speculative and we didn't completely understand this crash. By now, and seeing other reports from SDK authors at our submitted [Google Issue Tracker](https://issuetracker.google.com/issues/258176803), I believe it is a bug in the library. 

### Scope
* Before using WorkManager, check for its existence. Else, in rare cases that caused crashing, initialize it ourselves.
* Additionally, as another layer to handling this issue, when initializing WorkManager ourselves, surround it in a try-catch as it is may be the slightest bit possible for another thread to initialize it in the interim between checking and initializing.
* Provides a method to check if WorkManager is initialized in this process.
  - This is effectively the `WorkManager.isInitialized()` public method introduced in `androidx.work:work-*:2.8.0-alpha02`.
  - Please see https://android-review.googlesource.com/c/platform/frameworks/support/+/1941186 for the library's implementation

### ⚠️ Concerns ⚠️
* There was a known issue with Hilt dependency manager that we were not able to figure out in the `v4` of the SDK

# Testing
## Unit testing
No unit testing

## Manual testing
Limited ability to test manually due to unable to reproduce
- Tested normal app flow in an emulator
- Receiving notifications in background and swiped away state
- Opening app, etc.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2052)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import androidx.work.Configuration
 import androidx.work.WorkManager
 import androidx.work.impl.WorkManagerImpl
+import com.onesignal.debug.internal.logging.Logging
 
 object OSWorkManagerHelper {
     /**
@@ -34,11 +35,15 @@ object OSWorkManagerHelper {
     fun getInstance(context: Context): WorkManager {
         if (!isInitialized()) {
             try {
-                // Initialization can fail if another thread initializes in the small time gap
-                // https://android.googlesource.com/platform/frameworks/support/+/60ae0eec2a32396c22ad92502cde952c80d514a0/work/workmanager/src/main/java/androidx/work/impl/WorkManagerImpl.java#177
                 WorkManager.initialize(context, Configuration.Builder().build())
             } catch (e: IllegalStateException) {
-                // Admittedly starting to get hacky
+                /*
+                This catch is meant for the exception -
+                https://android.googlesource.com/platform/frameworks/support/+/60ae0eec2a32396c22ad92502cde952c80d514a0/work/workmanager/src/main/java/androidx/work/impl/WorkManagerImpl.java#177
+                1. We lost the race with another call to  WorkManager.initialize outside of OneSignal.
+                2. It is possible for some other unexpected error is thrown from WorkManager.
+                 */
+                Logging.error("OSWorkManagerHelper initializing WorkManager failed: ", e)
             }
         }
         return WorkManager.getInstance(context)
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
@@ -1,0 +1,46 @@
+package com.onesignal.notifications.internal.common
+
+import android.annotation.SuppressLint
+import android.content.Context
+import androidx.work.Configuration
+import androidx.work.WorkManager
+import androidx.work.impl.WorkManagerImpl
+
+object OSWorkManagerHelper {
+    /**
+     * Helper method to provide a way to check if WorkManager is initialized in this process.
+     * The purpose of this helper is to work around this bug - https://issuetracker.google.com/issues/258176803
+     *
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
```diff
@@ -0,0 +1,46 @@
...
+    @JvmStatic
+    @Synchronized
+    fun getInstance(context: Context): WorkManager {
+        if (!isInitialized()) {
+            try {
+                // Initialization can fail if another thread initializes in the small time gap
+                // https://android.googlesource.com/platform/frameworks/support/+/60ae0eec2a32396c22ad92502cde952c80d514a0/work/workmanager/src/main/java/androidx/work/impl/WorkManagerImpl.java#177
+                WorkManager.initialize(context, Configuration.Builder().build())
+            } catch (e: IllegalStateException) {
+                // Admittedly starting to get hacky
```
jkasten2, line 41: I would recommend we print something here, it could be helpful to know what `e` is here if we ever hit it.

Also, I believe we would probably only get here in two cases:
1. We lost the race with another call to ` WorkManager.initialize` outside of OneSignal.
2. If some other expected error is thrown from WorkManager.

I would replace your "Admittedly starting to get hacky" comment with something like that above, assuming I got all the details right.

jkasten2 (changes requested):

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
nan-li: Changed! The added try-catch was for this [comment](https://github.com/OneSignal/OneSignal-Android-SDK/issues/1672#issuecomment-1720744291) after the player model fix. 

I added the log. The original issue of WorkManager not initialized is already so rare, it would be even rarer to then reproduce failing on its initialization, but it's possible.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
```diff
@@ -0,0 +1,51 @@
...
+            try {
+                WorkManager.initialize(context, Configuration.Builder().build())
+            } catch (e: IllegalStateException) {
+                /*
+                This catch is meant for the exception -
+                https://android.googlesource.com/platform/frameworks/support/+/60ae0eec2a32396c22ad92502cde952c80d514a0/work/workmanager/src/main/java/androidx/work/impl/WorkManagerImpl.java#177
+                1. We lost the race with another call to  WorkManager.initialize outside of OneSignal.
+                2. It is possible for some other unexpected error is thrown from WorkManager.
+                 */
+                Logging.error("OSWorkManagerHelper initializing WorkManager failed: $e")
```
jkasten2, line 46: nit, there is a 2nd param to `Logging.error` we can put the exception under, like this:
`Logging.error("OSWorkManagerHelper initializing WorkManager failed", e)`

I am not sure if your stringify version displays the fullstack trace or just the exception.

jkasten2 (approved):

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/common/OSWorkManagerHelper.kt
nan-li: Good callout, the stringify version did not.

jkasten2 (approved):

PR #2055, opened 2024-04-14
[Improvement] Never set OperationRepo.enqueue's flush to true

jkasten2:
# Description
## One Line Summary
Never set `OperationRepo.enqueue`'s `flush` to `true` to improve network batching.

### Motivation
The 5.x.x branch of the OneSignal SDK is not as efficient on it's network calls as 4.x.x was. This is one small change to help improve some of this.

### Scope
Only effects delaying networks calls generated when calling `OneSignal.login` by up to seconds 5.

# Testing

## Manual testing
Tested calling `OneSignal.login` and ensured it was delayed by 5 seconds as expected. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2055)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -110,6 +110,7 @@ internal class OperationRepo(
         return waiter.waitForWake()
     }
 
+    // WARNING: Never set to true, until budget rules are added, even for internal use!
     private fun internalEnqueue(
         queueItem: OperationQueueItem,
         flush: Boolean,
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -380,7 +380,6 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                         externalId,
                         if (currentIdentityExternalId == null) currentIdentityOneSignalId else null,
                     ),
-                    true,
                 )
 
             if (!result) {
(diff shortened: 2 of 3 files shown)


jinliu9508:
Verified in my test that the login call is waited for more than 5 seconds in the queue to be processed.
image

However, I have two questions:
  1. Considering that both login() and logout() call LoginUserOperation, and several places call RefreshUserOperation, have we properly handled the case when the device or network is interrupted after the operation being enqueued but never executed?
  2. Now I am thinking about the old behavior regarding how adding Login/Refresh to the queue can affect other operations in the queue. Will setting the flush to true forcibly execute operations that are supposed to wait for the next interval, thereby causing some operations to be in a tight loop? If so, this PR may possibly solve that issue.

jkasten2:
> 1. Considering that both login() and logout() call LoginUserOperation, and several places call RefreshUserOperation, have we properly handled the case when the device or network is interrupted after the operation being enqueued but never executed?

Yes, these operations are stored locally and then replayed if the app is killed before it can complete them. Then when the app is opened again we restore the queue.

> 2. Now I am thinking about the old behavior regarding how adding Login/Refresh to the queue can affect other operations in the queue. Will setting the flush to true forcibly execute operations that are supposed to wait for the next interval, thereby causing some operations to be in a tight loop? If so, this PR may possibly solve that issue.

Removing the use of this flush flag means that the starting delay won't be skipped, so more operations could be grouped together after this PR. On the 404 issue, I believe this should help slow down (by 5 seconds) the tight loop (with the introduction of the bucket concept) but not fix the looping of create User itself.

jinliu9508 (approved):

PR #2068, opened 2024-04-29
[Improvement] make part of operationrepo initialization async

jinliu9508:
# Description
## One Line Summary
Make part of the initialization of OperationRepo asynchronous so that previously saved operations can be added asynchronously, preventing long-loading operations from blocking the main thread.

### Motivation
We have observed numerous ANRs during the initialization phase, with OperationRepo.init being the top cause. This issue does not occur consistently, and we suspect it may be related to the device's state or having a problem accessing device's disk. To address this, we plan to make the initialization process asynchronous in OperationRepo. By moving the loading part to a background thread, we aim to prevent the main thread from being blocked when the initialization process unexpectedly takes a long time.

### Scope
Saved operations from previous session will not be executed until they are loaded successfully. The order may be incorrect depends on the timing of the loading completion This change will try to insert saved operations starting from the beginning of the queue, and any later operation will be added to the end of the queue. 

# Testing
## Manual testing
The manual test I have done to ensure the SDK is loading saved operation correctly:
  1. Turn off wifi
  2. Login user and add a tag
  3. Kill the app
  4. Re-enter the app with wifi on, and observe that the saved operations are successfully loaded and executed shortly after.
image

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2068)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
@@ -25,10 +25,15 @@ import com.onesignal.user.internal.operations.impl.executors.LoginUserOperationE
 import com.onesignal.user.internal.operations.impl.executors.RefreshUserOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.SubscriptionOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.UpdateUserOperationExecutor
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
 import org.json.JSONObject
 
 internal class OperationModelStore(prefs: IPreferencesService) : ModelStore<Operation>("operations", prefs) {
-    init {
+
+    fun loadOperations() {
         load()
     }
 
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
@@ -25,14 +25,9 @@ import com.onesignal.user.internal.operations.impl.executors.LoginUserOperationE
 import com.onesignal.user.internal.operations.impl.executors.RefreshUserOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.SubscriptionOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.UpdateUserOperationExecutor
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.coroutineScope
-import kotlinx.coroutines.launch
 import org.json.JSONObject
 
 internal class OperationModelStore(prefs: IPreferencesService) : ModelStore<Operation>("operations", prefs) {
-
     fun loadOperations() {
         load()
     }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -81,13 +81,6 @@ internal class OperationRepo(
             }
         }
         this.executorsMap = executorsMap
-
-        // load operation in a separate thread to avoid halting the main process
-        coroutineScope.launch {
-            for (operation in _operationModelStore.list().withIndex()) {
-                internalEnqueue(OperationQueueItem(operation.value, bucket = enqueueIntoBucket), flush = false, addToStore = false, operation.index)
-            }
-        }
     }
 
     override fun <T : Operation> containsInstanceOf(type: KClass<T>): Boolean {
@@ -98,7 +91,11 @@ internal class OperationRepo(
 
     override fun start() {
         paused = false
-        coroutineScope.launch { processQueueForever() }
+        coroutineScope.launch {
+            // load saved operations first then start processing the queue to ensure correct operation order
+            loadSavedOperations()
+            processQueueForever()
+        }
     }
 
     override fun enqueue(
@@ -128,10 +125,10 @@ internal class OperationRepo(
         queueItem: OperationQueueItem,
         flush: Boolean,
         addToStore: Boolean,
-        index: Int = -1,
+        index: Int? = null,
     ) {
         synchronized(queue) {
-            if (index >= 0)
+            if (index != null)
                 queue.add(index, queueItem)
             else
                 queue.add(queueItem)
@@ -362,5 +359,17 @@ internal class OperationRepo(
         }
 
         return ops
+    }
+
+    /**
+     * Load saved operations from preference service and add them into the queue
+     * NOTE: sometimes the loading might take longer than expectedly due to device's state
+     */
+    private fun loadSavedOperations() {
+        // load operation in a separate thread to avoid halting the main process
+        _operationModelStore.loadOperations()
+        for (operation in _operationModelStore.list().withIndex()) {
+            internalEnqueue(OperationQueueItem(operation.value, bucket = enqueueIntoBucket), flush = false, addToStore = false, operation.index)
+        }
     }
 }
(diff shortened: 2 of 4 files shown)


jkasten2 (commented):
We need to delay `OperationModelStore.load()` as well, as this is what does the disk read. See this ANR stack trace:
```java
       at com.onesignal.common.modeling.Model.initializeFromJson(Model.kt:98)
       at com.onesignal.core.internal.operations.impl.OperationModelStore.create(OperationModelStore.kt:68)
       at com.onesignal.core.internal.operations.impl.OperationModelStore.create(OperationModelStore.kt:30)
       at com.onesignal.common.modeling.ModelStore.load(ModelStore.kt:162)
       at com.onesignal.core.internal.operations.impl.OperationModelStore.<init>(OperationModelStore.kt:32)
       at java.lang.reflect.Constructor.newInstance0(Native method)
       at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
       at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:89)
       at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
       at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
       at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:82)
       at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
       at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
       at com.onesignal.internal.OneSignalImp.initWithContext(OneSignalImp.kt:510)
       at com.onesignal.OneSignal.initWithContext(OneSignal.kt:135)
```

So the order of operations of `ServiceProvider` creating instances of classes is it goes deep first and works its way back up. So in this case since `OperationRepo` requires an instance of `ConfigModelStore` as part of it's constructor an instance of `ConfigModelStore` is created before `OperationRepo`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -121,12 +128,16 @@ internal class OperationRepo(
         queueItem: OperationQueueItem,
         flush: Boolean,
         addToStore: Boolean,
+        index: Int = -1,
```
jkasten2, line 131: Using `null` is better practice over magic numbers (especially since Kotlin has null safety)

jinliu9508:
> We need to delay `OperationModelStore.load()` as well, as this is what does the disk read. See this ANR stack trace:
> 
> ```java
>        at com.onesignal.common.modeling.Model.initializeFromJson(Model.kt:98)
>        at com.onesignal.core.internal.operations.impl.OperationModelStore.create(OperationModelStore.kt:68)
>        at com.onesignal.core.internal.operations.impl.OperationModelStore.create(OperationModelStore.kt:30)
>        at com.onesignal.common.modeling.ModelStore.load(ModelStore.kt:162)
>        at com.onesignal.core.internal.operations.impl.OperationModelStore.<init>(OperationModelStore.kt:32)
>        at java.lang.reflect.Constructor.newInstance0(Native method)
>        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
>        at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:89)
>        at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
>        at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
>        at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:82)
>        at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
>        at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
>        at com.onesignal.internal.OneSignalImp.initWithContext(OneSignalImp.kt:510)
>        at com.onesignal.OneSignal.initWithContext(OneSignal.kt:135)
> ```
> 
> So the order of operations of `ServiceProvider` creating instances of classes is it goes deep first and works its way back up. So in this case since `OperationRepo` requires an instance of `ConfigModelStore` as part of it's constructor, an instance of `ConfigModelStore` is created before `OperationRepo`.

Since load() is a genetic function from ModelStore, should we delay all model stores or limit the change to OperationModelStore only?

Also, both load() and persist() may be locking the models for longer than needed, especially they include the access to the preference service inside the synchronized block. Do you think we can also introduce a little optimization along with this issue?

jkasten2:
> Since load() is a genetic function from ModelStore, should we delay all model stores or limit the change to OperationModelStore only?

Longer term we probably want change `ModelStore`, so none of the models read from disk in the constructor. Or ensure we never create these instances on the main thread. In the short term to get a quick fix out scoping it to only `OperationModelStore` will probably what we should do for now.

> Also, both load() and persist() may be locking the models for longer than needed, especially they include the access to the preference service inside the synchronized block. Do you think we can also introduce a little optimization along with this issue?

Ya we could make those changes in this PR as well.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
```diff
@@ -25,10 +25,15 @@ import com.onesignal.user.internal.operations.impl.executors.LoginUserOperationE
 import com.onesignal.user.internal.operations.impl.executors.RefreshUserOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.SubscriptionOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.UpdateUserOperationExecutor
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
```
jkasten2, line 31: These 4 imports are not being used. I believe the klint points this out too.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -69,6 +69,10 @@ internal class OperationRepo(
     private val executeBucket get() =
         if (enqueueIntoBucket == 0) 0 else enqueueIntoBucket - 1
 
+    /**
+     * Perform initialization in background to prevent possible operation point of failures from
+     * blocking the creation of OperationRepo.
+     */
```
jkasten2, line 75: It would make more sense if this sort of comment was located around `loadSavedOperations()` in the code.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -352,4 +360,16 @@ internal class OperationRepo(
 
         return ops
     }
+
+    /**
+     * Load saved operations from preference service and add them into the queue
+     * NOTE: sometimes the loading might take longer than expectedly due to device's state
+     */
+    private fun loadSavedOperations() {
+        // load operation in a separate thread to avoid halting the main process
```
jkasten2, line 369: Would call out due to disk I/O here, or we probably can just remove this comment, since it will be duplicate with method as a whole comment above this one.

jkasten2 (changes requested):
All the logic changes look good to me.

Just needs some tests and some fixes to the linting and comments.

jkasten2 (changes requested):
Fix ups look good, but ktlint is still failing on CI. You can run it locally with the following command:
`./gradlew ktlintCheck --console=plain`

Also using the ktlint plugin in Android Studio helps make sure you follow lint rules.

jkasten2 (approved):

jkasten2:
@jinliu9508 I believe this PR will break [RecoverFromDroppedLoginBug.kt](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.1.10/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBug.kt). As when it calls [`OperationRepo.containsInstanceOf()`](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.1.10/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBug.kt#L64) it assumes it will already have loaded all the save operations from disk. Can you address this in a follow up PR?

Issue #2014, opened 2024-03-04
[Bug]: android.view.WindowManager$BadTokenException - Unable to add window


albertoderodrigo:
### What happened?

I see this error on firebase from your SDK

```
Fatal Exception: android.view.WindowManager$BadTokenException
Unable to add window -- token android.os.BinderProxy@8c472c5 is not valid; is your activity running?
```

This could help https://stackoverflow.com/questions/7811993/error-binderproxy45d459c0-is-not-valid-is-your-activity-running

Full log:

[io.socialgamesonline.slotcom_issue_e128821ca7d4cb21571b73e42c6bd165_crash_session_65E5B6C4004300016D57E2D4CA33E9FD_DNE_0_v2_stacktrace.txt](https://github.com/OneSignal/OneSignal-Android-SDK/files/14481743/io.socialgamesonline.slotcom_issue_e128821ca7d4cb21571b73e42c6bd165_crash_session_65E5B6C4004300016D57E2D4CA33E9FD_DNE_0_v2_stacktrace.txt)

### Steps to reproduce?

```Markdown
We are unable to reproduce it manually
```

### What did you expect to happen?

No errors

### OneSignal Android SDK version

5.1.6

### Android version

13, 12

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jennantilla:
@albertoderodrigo thank you for reporting; we'll look into this.

yosefBazaart:
same here, only see it in Google's Play store vitals.
this is our top crash so please fix it. 

`Exception android.view.WindowManager$BadTokenException:
  at android.view.ViewRootImpl.setView (ViewRootImpl.java:1743)
  at android.view.WindowManagerGlobal.addView (WindowManagerGlobal.java:524)
  at android.view.WindowManagerImpl.addView (WindowManagerImpl.java:150)
  at android.app.Dialog.show (Dialog.java:512)
  at android.app.AlertDialog$Builder.show (AlertDialog.java:1153)
  at com.onesignal.core.internal.permissions.AlertDialogPrepromptForAndroidSettings.show (AlertDialogPrepromptForAndroidSettings.kt)
  at com.onesignal.notifications.internal.permissions.impl.NotificationPermissionController.showFallbackAlertDialog (NotificationPermissionController.kt)
  at com.onesignal.notifications.internal.permissions.impl.NotificationPermissionController.onReject (NotificationPermissionController.kt)
[...]

PR #2073, opened 2024-05-03
BadTokenException  and WindowLeaked bug fix

sooraj-r-pai:
This code has fix for
#2014 [Bug]: android.view.WindowManager$BadTokenException - Unable to add window
and
android.view.WindowLeaked:

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2073)


jkasten2 (changes requested):
@sooraj-r-pai Thanks for submitting this PR. This PR however does have a side effect of the dialog never showing in these conditions. Ideally we wait until the next Activity is available before show this dialog. Something similar to the code we have in `RequestPermissionService.startPrompt()` could be done to do this.

The the above is a large undertaking, so logging as an error would be acceptable, so it is known the dialog was skipped. Use the OneSingal's internal `Logging.error` method if we go down this path.

PR #2075, opened 2024-05-03
Follow Up: add operationRepoLoadedListener and implement it to RecoverFromDroppe…

jinliu9508:
# Description
## One Line Summary
This is a followup PR to address a broken change made by this PR https://github.com/OneSignal/OneSignal-Android-SDK/pull/2068

### Motivation
The PR mentioned above has made the initialization of OperationRepo asynchronous that we can no longer assume all saved operations are loaded at start. This PR adds an event listener to RecoverFromDroppedLoginBug so that it listens to the loading completion event and starts the recovery after the loading has completed.

### Scope
The recover does no longer begin at start. Instead, it will begin when it receives the event notifier that the loading of OperationRepo has completed.

# Testing
## Unit testing
I have added two test units to ensure both classes are working properly.

## Manual testing
My manual testing step:
  1. enter the example app and turn wifi off
  2. perform a Login
  3. kill and re-enter the app
  4. notice that RecoverFromDroppedLoginBug inserts a LoginUser operation into the operationRepo queue, after the loading is completed

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2075)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
@@ -9,9 +9,6 @@ import com.onesignal.core.internal.time.impl.Time
 import com.onesignal.debug.LogLevel
 import com.onesignal.debug.internal.logging.Logging
 import com.onesignal.mocks.MockHelper
-import com.onesignal.user.internal.identity.IdentityModel
-import com.onesignal.user.internal.identity.IdentityModelStore
-import com.onesignal.user.internal.migrations.RecoverFromDroppedLoginBug
 import com.onesignal.user.internal.operations.ExecutorMocks.Companion.getNewRecordState
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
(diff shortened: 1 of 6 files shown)


OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
```diff
@@ -590,6 +593,20 @@ class OperationRepoTests : FunSpec({
...
+        val mockIdentityModelStore = mockk<IdentityModelStore>()
+        val recovery = RecoverFromDroppedLoginBug(mocks.operationRepo, mockIdentityModelStore, mocks.configModelStore)
+
+        mocks.operationRepo.start()
+        recovery.start()
+
+        verify {
+            mocks.operationRepo.subscribe(recovery)
+        }
+    }
```
jkasten2, line 609: This is testing `RecoverFromDroppedLoginBug` instead of OperationRepo, as the only thing we are verifying is that `operationRepo.subscribe` was called. This is a fine test, but such a test would belong in `RecoverFromDroppedLoginBugTest.kt`.

If we want to test the new `OperationRepo` feature, we should test that it actually fired the event instead.

jinliu9508: Good point! I have fixed that and add one test unit in each class to ensure the event listener works correctly.

jkasten2 (approved):

PR #2076, opened 2024-05-07
Follow Up to PR #2068: ensure internalEnqueue can't add the same operation

jkasten2:
# Description
## One Line Summary
Fixes a bug with the unreleased PR #2068, where `OperationRepo.queue` would have duplicated operations on the first clean start.

## Details
This fixes a bug where loadSavedOperations would duplicate operations in OperationRepo.queue. This is because it is possible for operations to be added before it is started and these are also persisted to disk.

### Motivation
The duplicate operations would cause the OperationRepo to stop processing requests, which causes all User and Subscription creates and updates to stop working.

### Scope
Only affects the operation repo enqueue logic.

# Testing
## Unit testing
Added a new Unit test for loadSavedOperations, ensuring it doesn't duplicate requests. Also OperationRepoTest's mockOperationModelStore now simulates a store, with an in memory implementation.

## Manual testing
Test on Android 14 with a clean install, ensure the logcat error is no long there. Also that we are now able to get a OneSignal id in orignal scenario.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2076)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -48,7 +48,7 @@ internal class OperationRepo(
     )
 
     private val executorsMap: Map<String, IOperationExecutor>
-    private val queue = mutableListOf<OperationQueueItem>()
+    internal val queue = mutableListOf<OperationQueueItem>()
     private val waiter = WaiterWithValue<LoopWaiterMessage>()
     private var paused = false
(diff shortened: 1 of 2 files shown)


jinliu9508 (approved):
Good solution!

I have tested that the operation queue no longer contains duplicate operation and no more error has occurred.
image

PR #2083, opened 2024-05-09
[Tests] Fix flaky tests 2024/05/09

jkasten2:
# Description
## One Line Summary
Fix some flaky test related to the OperationRepo and HttpClient.

### Motivation
We want tests to be stable and reliable.

### Scope
Only corrects tests.

# Testing
## Unit testing
Fixed three tests, recommend reviewing commit by commit.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2083)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/http/HttpClientTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/http/HttpClientTests.kt
@@ -239,7 +239,7 @@ class HttpClientTests : FunSpec({
 
         mocks.factory.mockResponse = mockSuccessfulResponse
         val response2 =
-            withTimeoutOrNull(999) {
+            withTimeoutOrNull(975) {
                 mocks.httpClient.post("URL", JSONObject())
             }
 
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBugTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBugTests.kt
@@ -1,6 +1,8 @@
 package com.onesignal.user.internal.migrations
 
+import com.onesignal.common.threading.Waiter
 import com.onesignal.core.internal.config.ConfigModelStore
+import com.onesignal.core.internal.operations.IOperationRepoLoadedListener
 import com.onesignal.core.internal.operations.impl.OperationModelStore
 import com.onesignal.core.internal.operations.impl.OperationRepo
 import com.onesignal.core.internal.time.impl.Time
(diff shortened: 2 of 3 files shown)


OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/http/HttpClientTests.kt
```diff
@@ -239,7 +239,7 @@ class HttpClientTests : FunSpec({
 
         mocks.factory.mockResponse = mockSuccessfulResponse
         val response2 =
-            withTimeoutOrNull(999) {
+            withTimeoutOrNull(975) {
```
jinliu9508, line 242: What is significant in changing the number from 975 to 999?

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBugTests.kt
```diff
@@ -31,16 +33,24 @@ class RecoverFromDroppedLoginBugTests : FunSpec({
...
         recovery.start()
+        val waiter = Waiter()
+        operationRepo.addOperationLoadedListener(
+            object : IOperationRepoLoadedListener {
+                override fun onOperationRepoLoaded() {
+                    waiter.wake()
+                }
+            },
+        )
+        operationRepo.start()
```
jinliu9508, line 48: Should we change the order so that the operationRepo starts before the recovery starts, since that is most of the case in real environment?

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/http/HttpClientTests.kt
jkasten2: Test sometimes failed on "response2 shouldBe null" line below, as `999` was too close to `1000`, which means the background thread could sometimes take priority of the main thread and the main thread could take over 1secound to execute. I ran 200 iterations of the test with 975 and passed.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBugTests.kt
jkasten2: True that it would be more real world, but it also means it could be flaky. This realization was why I created PR #2084 to change the production code.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/http/HttpClientTests.kt
jinliu9508: Interesting!

jinliu9508 (approved):

PR #2059, opened 2024-04-17
[Fix]Handle incorrect 404 responses; add a delay after creates and retries on 404 of new ids

jkasten2:
# Description
## One Line Summary
Handle incorrect 404 responses from the OneSignal's backend by; 1. Add a delay after creates; 2. Retry on 404 for a short window after create.

### Backend Issue
OneSignal sometimes returns a 404 on GET an PATCH requests if you are accessing something immediately after it was created. Normally the OneSignal backend can accept fetch/updates right way, but it all depends on it's server load. There isn't a guaranteed amount of time a client can wait, so SDKs has to work around this problem.

### SDK's work around strategy
This PR introduces two ways to work around this 404 problem:
1. Add a minimum delay after creating records (User or Subscription) before allowing any operations to fetch or update that specific record, based on the onesignalId or subscriptionId.
   * We picked 5 second as this "cool down". This is a best guess, we are balancing SDK responsiveness with the chance of running into this issue.
   * This has a trade off side-effect, improving batching, which may save some network calls, but again, at the cost of some SDK responsiveness.
2. As the 404/410 may still happen when the backend is under unusual load (a 5 second delay may not always be enough) we also retry on 404/410 within a longer but still short window
   * We picked 60 seconds, again this is a best guess, we are balancing the same things here.

### Motivation
404/410 responses are being handled by recreating the missing record (User or Subscription), however this isn't always the right way to handle these, give the backend caveat noted above. Recreating these records can create side-effects and extra load we want to avoid.

### Scope
Add a delay processing some operations in the `OperationRepo`, and affects 404/410 handling in executors.

### Future Considerations
* Calls that do not go through the OperationRepo could have this same logic applied.
    - Such as GET /iams, or maybe this could be move to the OperationRepo.
* The two new parameters (and existing operation repo ones) could be read from remote params, so they can be tweaked later.

# Testing
## Unit testing
* [Failing test to prove the issue](https://github.com/OneSignal/OneSignal-Android-SDK/actions/runs/8728583277/job/23948694395?pr=2059#step:6:783)
* [Test passing after ](https://github.com/OneSignal/OneSignal-Android-SDK/actions/runs/8728911515/job/23949744945#step:6:781)
* [New 404 tests added](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2059/commits/bc9f938dfe46ef93b0000843e75a900310504516)

## Manual testing
Tested on an Android 14 emulator.
Tested with the following code:
```Java
[...]

PR #2087, opened 2024-05-14
[Fix] grouping skipping opRepoPostCreateDelay, causing operations being applied out of order when multiple login operations are pending. (fixes issue since 5.1.10)

jkasten2:
# Description
## One Line Summary
Fix operations being applied out of order when multiple login operations are pending.

## Details
`OperationRepo.getGroupableOperations()` was not respecting `opRepoPostCreateDelay` like `OperationRepo.getNextOps()`, this would lead to some being applied out of order. The only known case is when there is a login in the queue for an Anonymous User crate and also User A and another for User B. The anonymous User create happens first with a push subscription which is correct. Then we have to wait for opRepoPostCreateDelay time after that create before we can attempt to identify it has User A, this is also ok. However we skip to User B and try to create it and transfer the subscription, this isn't correct as we also should wait `opRepoPostCreateDelay` for the new subscription id.

We addressed `getGroupableOperations()` in this PR however it was not enough, the OperationRepo and it's executors still ran into edge cases where ids would get mixed up due to intermixing of operations on different users. So in this PR we consistently stall the whole OperationRepo processing to avoid the issues, until we can make a flow up PR to improve this later.

While this PR fixes the issue noted above there should be another mechanism in place to prevent such an issue. This should be done in a follow up PR.

### Motivation
No matter how quickly or the offline state of the device operations should be applied in the correct order.

### Scope
Corrects `OperationRepo` ordering and makes now consistently delays the OperationRepo after creating records.

### Related
This is a follow up to PR #2059, where that logic created the issue noted above.

# Testing
## Unit testing
* [Failing test proving the issue](https://github.com/OneSignal/OneSignal-Android-SDK/actions/runs/9071886618/job/24926446649?pr=2087#step:6:819).

## Manual testing
Tested on Android 6 with the following scenario:
1. Turned on Airplane mode
2. Opened the app
3. `OneSignal.login("A")`
4. `OneSignal.User.addEmail("a@a.com")`
5. `OneSignal.User.addTag("a", "a")`
3. `OneSignal.login("B")`
4. `OneSignal.User.addEmail("b@b.com")`
5. `OneSignal.User.addTag("b", "b")`
6. Closed app
7. Turned off Airplane mode
8. Opened app again
9. Observed correct REST API calls and checked records on OneSignal dashboard.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [X] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2087)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -358,8 +358,7 @@ internal class OperationRepo(
      * THIS SHOULD BE CALLED WHILE THE QUEUE IS SYNCHRONIZED!!
      */
     private fun getGroupableOperations(startingOp: OperationQueueItem): List<OperationQueueItem> {
-        val ops = mutableListOf<OperationQueueItem>()
-        ops.add(startingOp)
+        val ops = mutableListOf(startingOp)
 
         if (startingOp.operation.groupComparisonType == GroupComparisonType.NONE) {
             return ops
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -372,27 +372,25 @@ internal class OperationRepo(
                 startingOp.operation.modifyComparisonKey
             }
 
-        if (queue.isNotEmpty()) {
-            for (item in queue.toList()) {
-                val itemKey =
-                    if (startingOp.operation.groupComparisonType == GroupComparisonType.CREATE) {
-                        item.operation.createComparisonKey
-                    } else {
-                        item.operation.modifyComparisonKey
-                    }
-
-                if (itemKey == "" && startingKey == "") {
-                    throw Exception("Both comparison keys can not be blank!")
+        for (item in queue.toList()) {
+            val itemKey =
+                if (startingOp.operation.groupComparisonType == GroupComparisonType.CREATE) {
+                    item.operation.createComparisonKey
+                } else {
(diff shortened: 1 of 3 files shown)


nan-li:
I'm testing this flow on a new install with network off:
```java
// Init OneSignal
// Anon user creation
OneSignal.login("<EXISTING_EUID_USER_A");
OneSignal.getUser().addTag("a", "a");

OneSignal.login("B");
OneSignal.getUser().addEmail("b@b.com");
```
Then turn network on.

There's some unexpected ordering and requests made, but eventually everything ends up in the correct states, it seems

1. CreateUser "B" sends immediately after the anonymous user is created (email is in payload, no push sub in payload)
2. IdentifyUser "A" sends 5 seconds after the anonymous user is created, returns 409
3. CreateUser "A" sends with push sub ID in payload (push sub has no token)
4. Call go GetUser "B" happens 5 second after its create
5. Call to GetUser "A" happens 5 seconds after its create
6. Push token received, enqueues a CreateSub with the token, because no push sub is in the store?
7. Call to TransferPushSub to "B" (by push sub ID, note that this sub has no token)
8. Call to CreateSubscription with token in payload for "B". A new push subscription ID is returned.
9. Any updates to push sub do send to the new push sub ID
10. Old push sub ID is still the one in the `OneSignal-Subscription-Id` header

nan-li:
Looking at this scenario again 👆, I think the issue is that for User B, there are 3 operations enqueued:
- login-user
- transfer-subscription
- create-subscription

However, since `transfer-subscription` is having the 5 seconds block, the other 2 operations go together without waiting, and that creates side effects down the line.

nan-li (approved):

PR #2093, opened 2024-05-21
[Fix] Continuation called with OneSignal.getNotifications().requestPermission does not fire when permission is already granted

jinliu9508:
# Description
## One Line Summary
Fix a bug causing OneSignal.getNotifications().requestPermission with continuation not firing when permission is granted.

### Motivation
This will fix some wrappers that utilize a Java bridge (like Unity and Flutter) and await forever and never resolve.

# Testing
## Manual testing
Scenarios that I manually tested:
  a. Notification permission is off:
      1. prompt for permission and allow
      2. request again and observe that no prompt shows up and the continuation is fired with permissionResult = true
  b.  Notification permission is off:
      1. prompt for permission and dismiss both dialogs
      2. observe that the continuation is fired with permissionResult = false
  c. Notification permission is on:
      1. request permission and observe that no prompt shows up and the continuation is fired with permissionResult = true

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2093)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
@@ -43,6 +43,7 @@ import com.onesignal.notifications.R
 import com.onesignal.notifications.internal.common.NotificationHelper
 import com.onesignal.notifications.internal.permissions.INotificationPermissionChangedHandler
 import com.onesignal.notifications.internal.permissions.INotificationPermissionController
+import kotlinx.coroutines.yield
 
 internal class NotificationPermissionController(
     private val _application: IApplicationService,
@@ -84,6 +85,10 @@ internal class NotificationPermissionController(
      * to notify of the status.
      */
     override suspend fun prompt(fallbackToSettings: Boolean): Boolean {
+        // Calling yield() to force a suspension point because Kotlin Continuation won't work
+        // properly from java caller if a suspend function does not actually suspend
+        yield()
+
         if (notificationsEnabled()) {
             return true
         }


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -84,11 +84,7 @@ internal class NotificationPermissionController(
      * to notify of the status.
      */
     override suspend fun prompt(fallbackToSettings: Boolean): Boolean {
-        if (notificationsEnabled()) {
-            return true
```
jkasten2, line 88: What the core thing that fixed this issue? Why doesn't returning a raw boolean value to work for Java? We also still have a `return false` below in this function, doesn't that also have this issue too?

jkasten2 (changes requested):

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
jinliu9508: The suspend functions have an additional parameter of the Continuation type, and this object holds the coroutine context and reference to the function caller. This allows us to direct the flow back to the caller once the suspension has completed. This is the continuation mechanism behind the scenes, which works properly when the suspend function actually suspends.

In my experiment, however, if the suspend function returns without any suspension, the continuation from its caller does not seem to resume properly. In the case of our prompt() function, it may return true without suspension if notificationsEnabled() is true, making the function not "suspend," and thus causing the continuation mechanism not to set up as expected.

An interesting finding: adding delay(1) right before return true makes the continuation work properly. So I think the way the continuation works is that it executes the suspend function until it hits the first suspension point, then sets up a callback to call resumeWith(result) after the suspension point has completed. But if no suspension point is reached at all, resumeWith(result) will never be called, causing the onFinished block registered by Continue.with to never run.

This might be a characteristic of Kotlin Continuation, and there is no easy way around it without delving deep into the Kotlin library. So here are my solutions. The basic idea is to force suspension even if unnecessary, just to make the suspend function truly suspend:

  1. Remain the same change, less efficient. This makes prompt() no longer return true directly but suspends to launch PermissionActivity to call onAccept() that wakes the waiter to resume the suspension. Slightly more resources will be used, and the user may notice a flash on the screen due to switching Android activities. We still need to deal with the case when returning false because the continuation will not work with that as well. 
  2. Easy but ugly: add delay(1) right before the end of any non-suspending logic in a suspend function that can potentially be called with a continuation. This is just to set up a proper continuation mechanism. We will also need to do the same to the line 'return false' in this function.
  3. Modify requestPermission to take a second parameter of a custom callback. This will require updating any wrapper SDK to match the new signature. Not a very promising solution.

Let me know what you think!

jkasten2: I did some searching on this as well and couldn't find any other better solutions.
* Found that there is a `CompletableFuture` type that might be a better API, however this is basically an option B the option 3 you listed above.

I calling `delay` is probably our best option as it doesn't introduce any new APIs where the developer has to read docs and could pick the wrong one.

Feedback on the the `delay` commit 6df7a9f939f781d8c4eb600e1f6383e6362826bf, I would just always call `delay()` at the top of the function. This means we don't have to add a `delay()` call for each raw return (we have a true and a false right now) and so if code is refactored it isn't skipped. Also we should add a comment to the `delay()` call that states it is so the callback works for Java.

Lastly instead of `delay(1)` can we use `yield()` instead? I think this would work but we would need to test to confirm. `yield()` should have a shorter delay and is a bit less hacky

jinliu9508: yield() works perfectly in my test and I have updated the code to replace delay with yield.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -84,6 +85,10 @@ internal class NotificationPermissionController(
      * to notify of the status.
      */
     override suspend fun prompt(fallbackToSettings: Boolean): Boolean {
+        // Calling yield() to force a suspension point because Kotlin Continuation won't work
```
jkasten2, line 88: nit, could we point out this is to specifically address calls from Java?
* You're are correct that the issue with Kotlin's Continuation but bring up the Java caller make is very clear at a glance.

jkasten2 (approved):

nan-li:
Can clean up the commit history?

PR #2102, opened 2024-05-30
Bump minimum FCM version

jinliu9508:
Build failed by using FCM version prior to 21.0.0: e:~/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt: (127, 45): Unresolved reference: token

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2102)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/build.gradle
+++ b/OneSignalSDK/onesignal/notifications/build.gradle
@@ -72,12 +72,11 @@ dependencies {
 
     compileOnly('com.amazon.device:amazon-appstore-sdk:[3.0.1, 3.0.99]')
 
-    // firebase-messaging:18.0.0 is the last version before going to AndroidX
-    // firebase-messaging:19.0.0 is the first version using AndroidX
+    // firebase-messaging:21.0.0 introduces FirebaseMessaging.getToken API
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/notifications/build.gradle
```diff
@@ -77,7 +77,7 @@ dependencies {
     // firebase-messaging:23.0.0 incoporates fix for SecurityException: Not allowed to bind to service
```
nan-li, line 77: We have some comments above that are no longer relevant /  we can add a comment about why moving to version 21...
```
    // firebase-messaging:18.0.0 is the last version before going to AndroidX
    // firebase-messaging:19.0.0 is the first version using AndroidX
```

nan-li (changes requested):

OneSignalSDK/onesignal/notifications/build.gradle
jinliu9508: Good point! I have pushed the change for the comment.

nan-li (approved):

PR #2109, opened 2024-06-05
consentGiven true forces wake when delaying for retry

emawby:
# Description
## One Line Summary
Retriable errors now use waiter instead of delay and it can only be interrupted by force calls to waiter.wake or waiting the 

## Details
We were previously using delay() to wait to retry failed server requests. This included privacy consent not given. This PR adds a way to interrupt that delay when privacy consent goes from false to true.

We cannot use delay() for this so it uses the waiter instead. This has the side effect of making all retry-able errors use the waiter.

### Motivation
We want a way to immediately process deltas when privacy consent goes from false to true.

### Scope
Retry-able operations and privacy consent given

# Testing
## Unit testing

## Manual testing
Tested not giving privacy consent to ensure we were properly delaying retries. Then tested providing privacy consent and ensured that the operations were sent immediately.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2109)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -1,5 +1,6 @@
 package com.onesignal.core.internal.operations.impl
 
+import android.os.Looper
 import com.onesignal.common.threading.WaiterWithValue
 import com.onesignal.core.internal.config.ConfigModelStore
 import com.onesignal.core.internal.operations.ExecutionResult
@@ -183,6 +184,10 @@ internal class OperationRepo(
         }
     }
 
+    override fun forceProcessDeltas() {
+        waiter.wake(LoopWaiterMessage(true, 0))
+    }
+
     /**
      *  Waits until a new operation is enqueued, then wait an additional
      *  amount of time afterwards, so operations can be grouped/batched.
@@ -334,7 +339,14 @@ internal class OperationRepo(
         val delayFor = max(delayForOnRetries, retryAfterSecondsNonNull * 1_000)
         if (delayFor < 1) return
         Logging.error("Operations being delay for: $delayFor ms")
-        delay(delayFor)
+        var wakeMessage = LoopWaiterMessage(false, 0)
+        while (!wakeMessage.force) {
+            val waitedTheFullTime =
+                    withTimeoutOrNull(delayFor) {
+                        wakeMessage = waiter.waitForWake()
+                    } == null
+            if (waitedTheFullTime) break
+        }
     }
 
     internal fun getNextOps(bucketFilter: Int): List<OperationQueueItem>? {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -1,6 +1,5 @@
 package com.onesignal.core.internal.operations.impl
 
-import android.os.Looper
 import com.onesignal.common.threading.WaiterWithValue
 import com.onesignal.core.internal.config.ConfigModelStore
 import com.onesignal.core.internal.operations.ExecutionResult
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -183,6 +184,10 @@ internal class OperationRepo(
         }
     }
 
+    override fun forceProcessDeltas() {
```
jkasten2, line 187: I think this should be called `forceExecuteOperations`, since we have the method name ` executeOperations` in this code base.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -334,7 +339,14 @@ internal class OperationRepo(
         val delayFor = max(delayForOnRetries, retryAfterSecondsNonNull * 1_000)
         if (delayFor < 1) return
         Logging.error("Operations being delay for: $delayFor ms")
-        delay(delayFor)
+        var wakeMessage = LoopWaiterMessage(false, 0)
+        while (!wakeMessage.force) {
+            val waitedTheFullTime =
+                    withTimeoutOrNull(delayFor) {
+                        wakeMessage = waiter.waitForWake()
```
jkasten2, line 346: Thinking about this again it would be better to create a different `waiter` for this. (calling it something like `retryWaiter` maybe?). After seeing the code again I have concerns how things will interact on the same `waiter`, as I don't know thinking `wake()` will wake this one and other existing. Sorry for the orignal suggestion of using the same `waiter` could work.

So here I think we can simply this logic, remove `waitedTheFullTime` as we want to break out of this wait right away if consent becomes true. Then we can call `wake()` on orignal waiter, without force or a time so it counts it as a "normal" operation, so we can batch other things like tags together.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -67,8 +67,12 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
     override var consentGiven: Boolean
         get() = configModel?.consentGiven ?: (_consentGiven == true)
         set(value) {
+            val oldValue = _consentGiven
             _consentGiven = value
             configModel?.consentGiven = value
+            if (oldValue != value && value) {
```
jkasten2, line 73: 👍 on not only checking if the value changes, but also if it currently true.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -334,7 +340,10 @@ internal class OperationRepo(
         val delayFor = max(delayForOnRetries, retryAfterSecondsNonNull * 1_000)
         if (delayFor < 1) return
         Logging.error("Operations being delay for: $delayFor ms")
-        delay(delayFor)
+        withTimeoutOrNull(delayFor) {
+            retryWaiter.waitForWake()
+        }
+        waiter.wake(LoopWaiterMessage(false, 0))
```
jkasten2, line 346: I believe this is a better fit in `forceExecuteOperations()`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -183,6 +185,10 @@ internal class OperationRepo(
         }
     }
 
+    override fun forceExecuteOperations() {
+        retryWaiter.wake(LoopWaiterMessage(true, 0))
```
jkasten2, line 189: nit, we can drop the 2nd param of `LoopWaiterMessage`, it defaults to `0`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -1,5 +1,6 @@
 package com.onesignal.core.internal.operations.impl
 
+import android.os.Looper
```
jkasten2, line 3: It doesn't looks like the new import is used. (it is probably failing ktlint).

jkasten2 (approved):

jkasten2 (approved):

PR #2111, opened 2024-06-06
Fix: Update onNotificationOpened to process each notification ID

jennantilla:
# Description
## One Line Summary
Update onNotificationOpened to process each notification ID

### Motivation
After customer reports of dropping CTR following upgrade from v4 to v5, it was discovered that clicking an unexpanded group notification results in only registering the click result for the final notification in the group. This PR restores the SDK behavior back to "default" of registering clicks on a per-notification basis when the entire group is clicked.

# Testing

## Manual testing
Tested on a Pixel 7 (Android 14) emulator that clicking an uncollapsed group notification results in the click being tracked for each notification.

# Affected code checklist
   - [X] Notifications
      - [ ] Display
      - [X] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2111)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
@@ -95,7 +95,6 @@ internal class NotificationLifecycleService(
     override suspend fun notificationOpened(
         activity: Activity,
         data: JSONArray,
-        notificationId: String,
     ) {
         intLifecycleHandler.suspendingFire { it.onNotificationOpened(activity, data) }
 
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
@@ -97,7 +97,7 @@ internal class NotificationLifecycleService(
         data: JSONArray,
         notificationId: String,
     ) {
-        intLifecycleHandler.suspendingFire { it.onNotificationOpened(activity, data, notificationId) }
+        intLifecycleHandler.suspendingFire { it.onNotificationOpened(activity, data) }
 
         // queue up the opened notification in case the handler hasn't been set yet. Once set,
         // we will immediately fire the handler.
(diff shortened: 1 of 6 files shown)


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
```diff
@@ -97,7 +97,7 @@ internal class NotificationLifecycleService(
         data: JSONArray,
         notificationId: String,
```
jkasten2, line 98: This `notificationId` is now unused so let's remove it. (Also maybe we can remove it in more places up the stack too)

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/listeners/NotificationListener.kt
```diff
@@ -67,26 +68,30 @@ internal class NotificationListener(
...
         val subscriptionId: String = _subscriptionManager.subscriptions.push.id
         val deviceType = _deviceService.deviceType
 
         for (i in 0 until data.length()) {
+            val notificationId = NotificationFormatHelper.getOSNotificationIdFromJson(data[i] as JSONObject?)
             if (postedOpenedNotifIds.contains(notificationId)) {
                 continue
             }
-            postedOpenedNotifIds.add(notificationId)
+            if (notificationId != null) {
```
jkasten2, line 82: Simplification suggestion, we do this `notificationId != null` check twice in this loop, recommend doing an early continue to simplify the code.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/listeners/NotificationListener.kt
```diff
@@ -98,10 +103,16 @@ internal class NotificationListener(
             NotificationHelper.getCampaignNameFromNotification(openResult.notification),
         )
 
+        // Handle the first element in the data array as the latest notification
```
jkasten2, line 106: Simplification suggestion, we could move this outcome logic to a private function, early returns to simply this logic as well.

jkasten2 (changes requested):
Logic itself looks correct, but there are some thing I would like to see cleaned up before merging so the code is more readable / doesn't misdirect.

jkasten2 (approved):

PR #2112, opened 2024-06-06
Poll for notification permission changes

emawby:
# Description
## One Line Summary
This PR adds polling for notification permission changes so that we can detect changes that happen from outside of the SDK in the current session.

## Details
Spawn a new thread where we check notification permission every second. If the permission has changed we fire the permission observer.
This means that we need to track the permission value to know if it has changed. 

This is done using an `_enabled` boolean that gets set when using OneSignal's prompting or when the polling detects a change. 

When the app loses focus we change the polling interval to 1 day to have the thread sleep until the app is brought into focus. 

### Motivation
Currently if an app prompts for permissions without using OneSignal or if the user changes permissions without closing the app (notification center long press on notification), we don't detect those permission changes until the next session begins.

### Scope
notification permission prompting

# Testing
## Unit testing
Added a unit testing file for the PermissionController. Currently has 3 tests around the new polling behavior.

## Manual testing
Tested by prompting directly through Android native APIs to test that the SDK detects the change.
Tested that backgrounding that properly sleeps the thread to ensure we aren't polling while the app is not in focus.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2112)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
@@ -193,9 +193,18 @@ class ConfigModel : Model() {
      * The minimum number of milliseconds required to pass to allow the fetching of IAM to occur.
      */
     var fetchIAMMinInterval: Long
-        get() = getLongProperty(::fetchIAMMinInterval.name) { 30000 }
+        get() = getLongProperty(::fetchIAMMinInterval.name) { 30_000 }
         set(value) {
             setLongProperty(::fetchIAMMinInterval.name, value)
+        }
+
+    /**
+     * The number of milliseconds between fetching the current notification permission value
+     */
+    var fetchNotificationPermissionInterval: Long
+        get() = getLongProperty(::fetchNotificationPermissionInterval.name) { 1_000 }
+        set(value) {
+            setLongProperty(::fetchNotificationPermissionInterval.name, value)
         }
 
     /**
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
@@ -199,12 +199,21 @@ class ConfigModel : Model() {
         }
 
     /**
-     * The number of milliseconds between fetching the current notification permission value
+     * The number of milliseconds between fetching the current notification permission value when the app is in focus
      */
-    var fetchNotificationPermissionInterval: Long
-        get() = getLongProperty(::fetchNotificationPermissionInterval.name) { 1_000 }
+    var foregroundFetchNotificationPermissionInterval: Long
+        get() = getLongProperty(::foregroundFetchNotificationPermissionInterval.name) { 1_000 }
         set(value) {
-            setLongProperty(::fetchNotificationPermissionInterval.name, value)
+            setLongProperty(::foregroundFetchNotificationPermissionInterval.name, value)
+        }
+
+    /**
+     * The number of milliseconds between fetching the current notification permission value when the app is out of focus
+     */
+    var backgroundFetchNotificationPermissionInterval: Long
+        get() = getLongProperty(::backgroundFetchNotificationPermissionInterval.name) { 1_800_000 }
+        set(value) {
+            setLongProperty(::backgroundFetchNotificationPermissionInterval.name, value)
         }
 
     /**
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
@@ -209,9 +209,10 @@ class ConfigModel : Model() {
 
     /**
      * The number of milliseconds between fetching the current notification permission value when the app is out of focus
+     * We want this value to be very large to effectively stop polling in the background
      */
     var backgroundFetchNotificationPermissionInterval: Long
-        get() = getLongProperty(::backgroundFetchNotificationPermissionInterval.name) { 1_800_000 }
+        get() = getLongProperty(::backgroundFetchNotificationPermissionInterval.name) { 86_400_000 }
         set(value) {
             setLongProperty(::backgroundFetchNotificationPermissionInterval.name, value)
         }
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
@@ -77,7 +77,7 @@ internal class NotificationPermissionController(
     init {
         this.enabled = notificationsEnabled()
         _requestPermission.registerAsCallback(PERMISSION_TYPE, this)
-        pollingWaitInterval = _configModelStore.model.foregroundFetchNotificationPermissionInterval
+        pollingWaitInterval = _configModelStore.model.backgroundFetchNotificationPermissionInterval
         registerPollingLifecycleListener()
         coroutineScope.launch {
             pollForPermission()
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
@@ -31,6 +31,7 @@ import android.os.Build
 import androidx.annotation.ChecksSdkIntAtLeast
 import com.onesignal.common.AndroidUtils
 import com.onesignal.common.events.EventProducer
+import com.onesignal.common.threading.Waiter
 import com.onesignal.common.threading.WaiterWithValue
 import com.onesignal.core.internal.application.ApplicationLifecycleHandlerBase
 import com.onesignal.core.internal.application.IApplicationService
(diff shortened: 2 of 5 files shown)


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -64,7 +70,22 @@ internal class NotificationPermissionController(
             )!!
 
     init {
+        _enabled = notificationsEnabled()
         _requestPermission.registerAsCallback(PERMISSION_TYPE, this)
+        coroutineScope.launch {// Should this be Android version restricted?
```
jkasten2, line 75: Seems like the AndroidX library' `areNotificationsEnabled() takes care of all Android versions:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core/src/main/java/androidx/core/app/NotificationManagerCompat.java;l=364-389?q=notificationmanagercompat

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -119,6 +140,7 @@ internal class NotificationPermissionController(
         get() = events.hasSubscribers
 
     override fun onAccept() {
+        _enabled = true
         waiter.wake(true)
         events.fire { it.onNotificationPermissionChanged(true) }
```
jkasten2, line 145: This might be a good time to put these 3 lines into its own function.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -64,14 +70,35 @@ internal class NotificationPermissionController(
             )!!
 
     init {
+        this.enabled = notificationsEnabled()
```
jkasten2, line 73: Not sure if `_application.appContext` is set at this point, `notificationsEnabled()` will always return `false` if it is not.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -64,14 +70,35 @@ internal class NotificationPermissionController(
...
+    }
+
+    private suspend fun pollForPermission() {
+        while (true) {
+            val enabled = this.notificationsEnabled()
+            if (this.enabled != enabled) { // If the permission has changed without prompting through OneSignal
+                this.enabled = enabled
+                events.fire { it.onNotificationPermissionChanged(enabled) }
+            }
+            delay(1_000) // should be a configurable value for unit tests
```
jkasten2, line 87: Note, might want to use a waiter here with a timeout to account for waking when the app is foregrounded, and changing the wait time to something very large when the app is backgrounded (to effectively pause the polling).

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
emawby: It has always been set and returns the correct value in my testing, but that is not to say that it couldn't happen. If it isn't then we will just fire the handler unnecessarily which isn't the worst thing. We could also save the value in sharedPreferences to reduce the likely hood of this occurring.

jkasten2: I wasn't actually sure off hand so wanted to make sure we know if this was the case / tested. Since it is working I think it's fine to set it like this.
* However I can see this being fragile, but caching it has it own problem, as it could be out of date.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
emawby: I updated this to use a waiter. The in focus interval is 1 second and the out of focus is currently 30 minutes. I am not sure what a reasonable value is for the out of focus polling. Do we even want to continue polling while in the background?

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -43,17 +44,26 @@ import com.onesignal.notifications.R
...
 internal class NotificationPermissionController(
     private val _application: IApplicationService,
     private val _requestPermission: IRequestPermissionService,
     private val _applicationService: IApplicationService,
     private val _preferenceService: IPreferencesService,
+    private val _configModelStore: ConfigModelStore,
 ) : IRequestPermissionService.PermissionCallback,
     INotificationPermissionController {
     private val waiter = WaiterWithValue<Boolean>()
+    private val pollingWaiter = WaiterWithValue<Boolean>()
```
jkasten2, line 62: We don't send a value so we can simply this a bit use `Waiter` instead.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
```diff
@@ -193,11 +193,29 @@ class ConfigModel : Model() {
...
+        get() = getLongProperty(::foregroundFetchNotificationPermissionInterval.name) { 1_000 }
+        set(value) {
+            setLongProperty(::foregroundFetchNotificationPermissionInterval.name, value)
+        }
+
+    /**
+     * The number of milliseconds between fetching the current notification permission value when the app is out of focus
+     */
+    var backgroundFetchNotificationPermissionInterval: Long
+        get() = getLongProperty(::backgroundFetchNotificationPermissionInterval.name) { 1_800_000 }
```
jkasten2, line 214: Let's use a high value of 1 day if we are going down this timmer route.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
```diff
@@ -193,11 +193,29 @@ class ConfigModel : Model() {
...
+     * The number of milliseconds between fetching the current notification permission value when the app is in focus
+     */
+    var foregroundFetchNotificationPermissionInterval: Long
+        get() = getLongProperty(::foregroundFetchNotificationPermissionInterval.name) { 1_000 }
+        set(value) {
+            setLongProperty(::foregroundFetchNotificationPermissionInterval.name, value)
+        }
+
+    /**
+     * The number of milliseconds between fetching the current notification permission value when the app is out of focus
```
jkasten2, line 211: Add comment noting that we don't really want to poll in the background, but using a very large value to essentially stop polling in the background.

jkasten2 (changes requested):
Comments are for if we keep this change the polling rate for the background to a value so high it essentially stops polling.

Still for stopping the polling loop and starting a new one when foregrounding if you want to go that route though, as the intent in the code would be more clear.

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/model/MainActivityViewModel.java
```diff
@@ -811,6 +811,9 @@ private void setupSubscriptionSwitch() {
 
     private void setupPromptPushButton() {
         promptPushButton.setOnClickListener(v -> {
+//            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
jkasten2, line 814: Remove commented out code.

jkasten2 (commented):
The implementation itself looks good to me now.

This PR is still noted as WiP, so I'll wait to approve until it's considered complete.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -64,14 +75,57 @@ internal class NotificationPermissionController(
             )!!
 
     init {
+        this.enabled = notificationsEnabled()
         _requestPermission.registerAsCallback(PERMISSION_TYPE, this)
+        pollingWaitInterval = _configModelStore.model.foregroundFetchNotificationPermissionInterval
```
jkasten2, line 80: We should start with the background time here to be safe. I believe the onFcous event fires right away if the app is already in focus so it will be updated right away if the app happens to really be in the foreground at this point.

jkasten2 (changes requested):
After looking through the test I have a concern about the default `pollingWaitInterval` value, left more details on the line of code.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
emawby: Good catch! Yes it does fire immediately and wakes so it starts polling correctly

jkasten2 (changes requested):
New "NotificationPermissionController permission polling fires permission changed event" test is failing on CI from the last commit. I believe the test needs to be updated to account for the new polling rate default.

jkasten2 (approved):
Test passes locally now for me, so it's most likely good now, but let's wait until CI finishes before merging to be sure.

PR #2113, opened 2024-06-06
[Feat] Application service focus update, cold start creates new session and drives user refresh

nan-li:
# Description
## One Line Summary
(1) All cold starts create new sessions, (2) refresh user on all new sessions, (3) Application Service fills in gaps where it did not fire `onFocus` for awaiting subscribers.

## Details
❓ Is my interpretation of this commit[ ApplicationService: Clarify a condition by naming it](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2113/commits/58df743e9eb9b2d474c6ebbbd24d47a292337fb0) accurate?

❓ Now that "refresh-user" and "track-session-start" always enqueue at the same time, the `UserRefreshService` could be removed and its task could move to the `SessionListener`.

**Application Service will fire `onFocus` even when it is too late**
* Typically, the app foregrounding will trigger the ApplicationService to fire `onFocus` to its subscribers.
* However, it is possible for OneSignal to initialize too late, so the ApplicationService itself is started too late to capture the application's early lifecycle events.
* In this case, the app is already foregrounded, so fire a subscriber's `onFocus` callback immediately upon it subscribing, as many services have logic to run in the **first** `onFocus` call of a cold start, and pass a parameter so the subscriber knows if the callback is fired from subscribing or from natural application lifecycle callbacks. Some subscribers such as permission controllers do not want the callback when it is fired from just subscribing.
* This may fix some other bugs commonly seen on wrappers such as permission not being detected and updated between cold starts, or background runnables not being cancelled.

**Session service creates new session on cold start**
* Also fix a bug where a cold start may not detect the start of a session correctly. On wrappers, it is common for OneSignal to initialize too late for the `ApplicationService` to `start()` and capture the Android `onActivityStarted` or `onActivityResumed` lifecycle callbacks, which was the primary way of session detection.
* **Player model reference**: In player model, we addressed this by triggering the `FocusTimeController` in the `init(context)` method when the ActivityLifecycleHandler is added AND the application is already in the foreground: [link](https://github.com/OneSignal/OneSignal-Android-SDK/blob/8a36a3238e7cbe07ddf26494a8916bbbdcce9cc4/OneSignalSDK/onesignal/src/main/java/com/onesignal/OneSignal.java#L976).

**Session service updates tracking active duration with cold starts**
* Additionally, the background task to send the session time **will be triggered by the existence of an active duration of 1 second or more and under 1 day, instead of whether the session is valid**. This is in line with player model logic.

**Refresh user on new sessions (not just cold start)**
* New sessions can happen when the app is backgrounded over 30 seconds. Let's refresh the user then, not just when the app is cold started.
* The user will still be refreshed only when the app is in the foreground.

### Motivation

- Align behavior with expectation
- Update last active always when cold started
- Fix some other bugs commonly found on wrappers due to OneSignal initializing later after activity already started

### Scope
**New session creation and when users are refreshed**
* Users are now refreshed on every new session triggered by backgrounding the app, **whereas they were previously only refreshed on a cold start**
* All cold starts will create new sessions, instead of also relying on 30 seconds backgrounded time

**Application Service will always fire `onFocus` when services are started up**
* Previously the SDK had to be initialized in the Applications `onCreate` call to capture and fire the early `onFocus` callback
* Now if the SDK is initialized later, the Application Service will still fire that callback when handlers subscribe to it

# Testing
## Unit testing
**Update Session Service and Application Service Tests**
* Extract helper methods them into a helper class for reuse
* `SessionServiceTests` - Add test `"session created in start when application is in the foreground"`
* `ApplicationServiceTests` - Add test `"focus will occur on subscribe when activity is already started"`

## Manual testing
Tested these primary scenarios
1. SDK initialized in Application `onCreate`
2. SDK initialized after activity is already running
3. Background app for under 30 seconds and over 30 seconds
4. Swipe away app and re-open in under 30 seconds and over 30 seconds

Checked for:
- [x]  Last active is updated (`track-session-start` operation)
- [x]  Refresh user (`refresh-user` operation )
- [x]  Session time accumulated correctly still
- [x]  Fetch IAMs correct

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2113)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/application/impl/ApplicationService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/application/impl/ApplicationService.kt
@@ -30,6 +30,7 @@ class ApplicationService() : IApplicationService, ActivityLifecycleCallbacks, On
     private val activityLifecycleNotifier = EventProducer<IActivityLifecycleHandler>()
     private val applicationLifecycleNotifier = EventProducer<IApplicationLifecycleHandler>()
     private val systemConditionNotifier = EventProducer<ISystemConditionHandler>()
+    private var shouldFireOnFocusOnSubscribing = false
 
     override val isInForeground: Boolean
         get() = entryState.isAppOpen || entryState.isNotificationClick
@@ -72,6 +73,7 @@ class ApplicationService() : IApplicationService, ActivityLifecycleCallbacks, On
 
     /**
      * Call to "start" this service, expected to be called during initialization of the SDK.
+     * Detects if this service should fire subscribers' onFocus() callbacks immediately on subscribing.
      *
      * @param context The context the SDK has been initialized under.
      */
@@ -110,6 +112,8 @@ class ApplicationService() : IApplicationService, ActivityLifecycleCallbacks, On
                 activityReferences = 1
                 nextResumeIsFirstActivity = false
             }
+            // Once listeners subscribe, fire their callbacks
+            shouldFireOnFocusOnSubscribing = true
         } else {
             nextResumeIsFirstActivity = true
             entryState = AppEntryAction.APP_CLOSE
@@ -120,6 +124,9 @@ class ApplicationService() : IApplicationService, ActivityLifecycleCallbacks, On
 
     override fun addApplicationLifecycleHandler(handler: IApplicationLifecycleHandler) {
         applicationLifecycleNotifier.subscribe(handler)
+        if (shouldFireOnFocusOnSubscribing) {
+            handler.onFocus(true)
+        }
     }
 
     override fun removeApplicationLifecycleHandler(handler: IApplicationLifecycleHandler) {
@@ -387,7 +394,7 @@ class ApplicationService() : IApplicationService, ActivityLifecycleCallbacks, On
                 entryState = AppEntryAction.APP_OPEN
             }
 
-            applicationLifecycleNotifier.fire { it.onFocus() }
+            applicationLifecycleNotifier.fire { it.onFocus(false) }
         } else {
             Logging.debug("ApplicationService.handleFocus: application never lost focus")
         }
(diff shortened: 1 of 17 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
```diff
@@ -47,6 +47,12 @@ internal class SessionListener(
 
     override fun onSessionEnded(duration: Long) {
         val durationInSeconds = duration / 1000
+
+        // Time is invalid if below 1 second or over a day
+        if (durationInSeconds < 1L || durationInSeconds > SECONDS_IN_A_DAY) {
+            return
```
jkasten2, line 53: I don't like that this can silently hide time accounting issues in the future. Thinking we should log the error.

Also think we should still send this time to OneSignal's servers, that way we know server side if we are still having issues.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
nan-li: I brought it over from the [player model](https://github.com/OneSignal/OneSignal-Android-SDK/blob/8a36a3238e7cbe07ddf26494a8916bbbdcce9cc4/OneSignalSDK/onesignal/src/main/java/com/onesignal/FocusTimeController.java#L111). 

We can still send it and log an error.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/application/impl/ApplicationService.kt
```diff
@@ -117,6 +124,9 @@ class ApplicationService() : IApplicationService, ActivityLifecycleCallbacks, On
 
     override fun addApplicationLifecycleHandler(handler: IApplicationLifecycleHandler) {
         applicationLifecycleNotifier.subscribe(handler)
+        if (shouldFireOnFocusOnSubscribing) {
```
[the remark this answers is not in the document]
nan-li: Agree, that is better. 

I tied this state to when the ApplicationService started, but regardless of when the ApplicationService started, another service can still subscribe too late.

jkasten2 (approved):

PR #2125, opened 2024-06-13
Fix/fcm push token bug

rgomezp:
# Description
**One line summary:** Remove redundant `retrievePushToken` call and refine local model replacement logic.

### Remove Unnecessary `retrievePushToken` Call
The `retrievePushToken` call in the `start` function has been removed. Since the config model is replaced every time, this call is redundant.

### Replace Local Models with Non-Push Subscriptions
Local models are now only replaced with non-push subscriptions. The backend should sync its push token state with the client state, as the client directly obtains the token during subscription.

# Testing
## Manual testing
Tested locally.
- Change project number
- Send push - failed
- Open app
- Send push - success

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2125)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
@@ -117,9 +117,22 @@ internal class RefreshUserOperationExecutor(
                 subscriptionModel.carrier = subscription.carrier ?: ""
                 subscriptionModel.appVersion = subscription.appVersion ?: ""
 
-                // We only add a push subscription if it is this device's push subscription.
-                if (subscriptionModel.type != SubscriptionType.PUSH || subscriptionModel.id == _configModelStore.model.pushSubscriptionId) {
+                // We only add a non-push subscriptions. For push, the device is the source of truth
+                // so we don't want to cache these subscriptions from the backend.
+                if (subscriptionModel.type != SubscriptionType.PUSH) {
                     subscriptionModels.add(subscriptionModel)
+                }
+            }
+            // Retrieve the push subscription ID from the backend configuration model store
+            val pushSubscriptionIdFromConfig = _configModelStore.model.pushSubscriptionId
+
+            if (pushSubscriptionIdFromConfig != null) {
+                // Retrieve the push subscription model from the model store
+                val cachedPushSubscriptionModel = _subscriptionsModelStore.get(pushSubscriptionIdFromConfig)
+
+                // If non-null, the cached push subscription matches the ID coming from backend config
+                if (cachedPushSubscriptionModel != null) {
+                    subscriptionModels.add(cachedPushSubscriptionModel)
                 }
             }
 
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -117,11 +117,19 @@ internal class RefreshUserOperationExecutor(
                 subscriptionModel.carrier = subscription.carrier ?: ""
                 subscriptionModel.appVersion = subscription.appVersion ?: ""
 
-                // We only add a push subscription if it is this device's push subscription.
-                if (subscriptionModel.type != SubscriptionType.PUSH || subscriptionModel.id == _configModelStore.model.pushSubscriptionId) {
+                // We only add a non-push subscriptions
```
jkasten2, line 120: Would be good to add a detail to this comment we don't want to cache push subscriptions from the backend, as we will add it below is the device is the source of truth.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -117,11 +117,19 @@ internal class RefreshUserOperationExecutor(
...
+                if (subscriptionModel.type != SubscriptionType.PUSH) {
                     subscriptionModels.add(subscriptionModel)
                 }
             }
+            var currentPushSubscriptionId = _configModelStore.model.pushSubscriptionId
```
jkasten2, line 125: It's confusing to me what the difference between `current` and `my`  `PushSubscription`.

jkasten2 (changes requested):
Left some comments, also looks like the failing test is due to this change and has to be updated.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -117,11 +117,24 @@ internal class RefreshUserOperationExecutor(
...
-                // We only add a push subscription if it is this device's push subscription.
-                if (subscriptionModel.type != SubscriptionType.PUSH || subscriptionModel.id == _configModelStore.model.pushSubscriptionId) {
+                // We only add a non-push subscriptions. For push, the device is the source of truth
+                // so we don't want to cache these subscriptions from the backend.
+                if (subscriptionModel.type != SubscriptionType.PUSH) {
                     subscriptionModels.add(subscriptionModel)
                 }
             }
+            // Retrieve the push subscription ID from the backend configuration model store
+            var pushSubscriptionIdFromConfig = _configModelStore.model.pushSubscriptionId
```
jkasten2, line 127: nit, can be `val` as it is not reassigned later.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -117,11 +117,24 @@ internal class RefreshUserOperationExecutor(
...
+
+            if (pushSubscriptionIdFromConfig != null) {
+                // Retrieve the push subscription model from the model store
+                var cachedPushSubscriptionModel = _subscriptionsModelStore.get(pushSubscriptionIdFromConfig)
```
jkasten2, line 131: nit, can be `val` as it is not reassigned later.

jkasten2 (approved):

PR #2127, opened 2024-06-17
Move start call of unused services off the main thread

jinliu9508:
# Description
## One Line Summary
Move start call of unused services off the main thread to make the initWithContext more efficient.

### Motivation
Our SDK initialization is currently taking longer to finish compare to some other SDKs. One of the reason is we construct service components like Location Manager or IAM Manager during the initialization phrase when they are not really needed until later. This PR aims to move the initialization of these services to background thread so they don't slow down the overall SDK initialization.

### Scope
The retrieval of services that were changed to initialize in background is not longer from a saved instance in OneSignalImp.kt. Instead, each of these components has a getter that retrieve its instance from the service provider. 

# Testing
## Manual testing
* I measure how long the initialization takes by tracking the start and finish time when 'initWithContext' is called. In my testing in a pixel 3a API 34 emulator, it took an average of 600 ms to complete initialization before the change and around 300 ms after the change, which is around ~50% faster. 
* Send notification, add tags, and outcome work.
* Calling 'initWithContext' in a background thread works. Other OneSignal calls can only work if they are called after 'initWithContext' and in the same thread.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2127)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
@@ -1,18 +1,23 @@
 package com.onesignal.core.internal.startup
 
+import com.onesignal.common.services.ServiceProvider
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.newSingleThreadContext
+
 internal class StartupService(
-    private val _bootstrapServices: List<IBootstrapService>,
-    private val _startableServices: List<IStartableService>,
+    private val services: ServiceProvider,
 ) {
+    private val coroutineScope = CoroutineScope(newSingleThreadContext(name = "StartupService"))
+
     fun bootstrap() {
-        // now that we have the params initialized, start everything else up
-        for (bootstrapService in _bootstrapServices)
-            bootstrapService.bootstrap()
+        services.getAllServices<IBootstrapService>().forEach { it.bootstrap() }
     }
 
-    fun start() {
-        // now that we have the params initialized, start everything else up
-        for (startableService in _startableServices)
-            startableService.start()
+    // schedule to start all startable services in a separate thread
+    fun scheduleStart() {
+        coroutineScope.launch {
+            services.getAllServices<IStartableService>().forEach { it.start() }
+        }
     }
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -86,7 +86,7 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
     override val debug: IDebugManager = DebugManager()
     override val session: ISessionManager get() =
         if (isInitialized) {
-            _session!!
+            services.getService()
         } else {
             throw Exception(
                 "Must call 'initWithContext' before use",
@@ -94,7 +94,7 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         }
     override val notifications: INotificationsManager get() =
         if (isInitialized) {
-            _notifications!!
+            services.getService()
         } else {
             throw Exception(
                 "Must call 'initWithContext' before use",
@@ -102,7 +102,7 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         }
     override val location: ILocationManager get() =
         if (isInitialized) {
-            _location!!
+            services.getService()
         } else {
             throw Exception(
                 "Must call 'initWithContext' before use",
@@ -110,26 +110,32 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         }
     override val inAppMessages: IInAppMessagesManager get() =
         if (isInitialized) {
-            iam!!
+            services.getService()
         } else {
             throw Exception(
                 "Must call 'initWithContext' before use",
             )
         }
-    override val user: IUserManager get() = if (isInitialized) _user!! else throw Exception("Must call 'initWithContext' before use")
+    override val user: IUserManager get() =
+        if (isInitialized) {
+            services.getService()
+        } else {
+            throw Exception(
+                "Must call 'initWithContext' before use",
+            )
+        }
 
     // Services required by this class
-    private var _user: IUserManager? = null
-    private var _session: ISessionManager? = null
-    private var iam: IInAppMessagesManager? = null
-    private var _location: ILocationManager? = null
-    private var _notifications: INotificationsManager? = null
-    private var operationRepo: IOperationRepo? = null
-    private var identityModelStore: IdentityModelStore? = null
-    private var propertiesModelStore: PropertiesModelStore? = null
-    private var subscriptionModelStore: SubscriptionModelStore? = null
-    private var startupService: StartupService? = null
-    private var preferencesService: IPreferencesService? = null
+    private val operationRepo: IOperationRepo
+        get() = services.getService()
+    private val identityModelStore: IdentityModelStore
+        get() = services.getService()
+    private val propertiesModelStore: PropertiesModelStore
+        get() = services.getService()
+    private val subscriptionModelStore: SubscriptionModelStore
+        get() = services.getService()
+    private val preferencesService: IPreferencesService
+        get() = services.getService()
 
     // Other State
     private val services: ServiceProvider
@@ -234,21 +240,10 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                 configModel!!.disableGMSMissingPrompt = _disableGMSMissingPrompt!!
             }
 
-            // "Inject" the services required by this main class
-            _location = services.getService()
-            _user = services.getService()
-            _session = services.getService()
-            iam = services.getService()
-            _notifications = services.getService()
-            operationRepo = services.getService()
-            propertiesModelStore = services.getService()
-            identityModelStore = services.getService()
-            subscriptionModelStore = services.getService()
-            preferencesService = services.getService()
+            val startupService = StartupService(services)
 
-            // Instantiate and call the IStartableServices
-            startupService = services.getService()
-            startupService!!.bootstrap()
+            // bootstrap services
+            startupService.bootstrap()
 
             if (forceCreateUser || !identityModelStore!!.model.hasProperty(IdentityConstants.ONESIGNAL_ID)) {
                 val legacyPlayerId =
@@ -298,8 +293,12 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
 
                         pushSubscriptionModel.sdk = OneSignalUtils.SDK_VERSION
                         pushSubscriptionModel.deviceOS = Build.VERSION.RELEASE
-                        pushSubscriptionModel.carrier = DeviceUtils.getCarrierName(services.getService<IApplicationService>().appContext) ?: ""
-                        pushSubscriptionModel.appVersion = AndroidUtils.getAppVersion(services.getService<IApplicationService>().appContext) ?: ""
+                        pushSubscriptionModel.carrier = DeviceUtils.getCarrierName(
+                            services.getService<IApplicationService>().appContext,
+                        ) ?: ""
+                        pushSubscriptionModel.appVersion = AndroidUtils.getAppVersion(
+                            services.getService<IApplicationService>().appContext,
+                        ) ?: ""
 
                         configModel!!.pushSubscriptionId = legacyPlayerId
                         subscriptionModelStore!!.add(
@@ -328,10 +327,9 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
                 Logging.debug("initWithContext: using cached user ${identityModelStore!!.model.onesignalId}")
             }
 
-            startupService!!.start()
-
+            // schedule service starts out of main thread
+            startupService.scheduleStart()
             isInitialized = true
-
             return true
         }
     }
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/startup/StartupServiceTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/startup/StartupServiceTests.kt
@@ -1,16 +1,30 @@
 package com.onesignal.core.internal.startup
 
+import com.onesignal.common.services.ServiceBuilder
+import com.onesignal.common.services.ServiceProvider
 import com.onesignal.debug.LogLevel
 import com.onesignal.debug.internal.logging.Logging
 import io.kotest.assertions.throwables.shouldThrowUnit
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
+import io.mockk.coVerifyOrder
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.spyk
 import io.mockk.verify
 
 class StartupServiceTests : FunSpec({
+    fun setupServiceProvider(
+        bootstrapServices: List<IBootstrapService>,
+        startableServices: List<IStartableService>,
+    ): ServiceProvider {
+        val serviceBuilder = ServiceBuilder()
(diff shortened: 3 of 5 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -234,21 +246,10 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-            subscriptionModelStore = services.getService()
-            preferencesService = services.getService()
-
-            // Instantiate and call the IStartableServices
-            startupService = services.getService()
-            startupService!!.bootstrap()
+            // bootstrap services
+            for (bootstrapService in services.getAllServices<IBootstrapService>()) {
+                bootstrapService.bootstrap()
+            }
```
jkasten2, line 252: Why was `startupService!!.bootstrap()` removed and replaced with this for loop?
* `startupService` abstracted this away so I think it was clearer before.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
```diff
@@ -80,6 +85,26 @@ class ServiceProvider(
         }
     }
 
+    fun scheduleStartServices() {
```
jkasten2, line 88: This method is stretching what I think this ServiceProvider should do. We already have a `StartupService`start()` method that I think this is a much better fit for. Can we not just change the code in that function instead?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
```diff
@@ -80,6 +85,26 @@ class ServiceProvider(
...
+        synchronized(serviceMap) {
+            for (serviceReg in serviceMap!![IStartableService::class.java]!!) {
+                // we can directly start services that were previously resolved
+                if (serviceReg.isResolved()) {
```
jkasten2, line 92: Why not just use getService() to simply this?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
```diff
@@ -80,6 +85,26 @@ class ServiceProvider(
...
+                    val service = serviceReg.resolve(this) as IStartableService
+                    service.start()
```
jkasten2, line 94: We should always run `start()` off the main thread if that is the optimization we are trying to do for all of them. Is there any reason why you wouldn't?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -120,16 +121,27 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
 
     // Services required by this class
     private var _user: IUserManager? = null
+        get() = services.getService()
```
jkasten2, line 124: Do you think these should be `lazy`?, so getService() only have to be called once.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Calling startupService within initWithContext causes the serviceProvider to resolve all bootstrap and startable services right away. In my benchmark calling getService for startupService contributes around 80% of execution time in initWithContext. So the idea of this PR is to prevent startupService from resolving all services immediately and instead resolve the unused services in a background thread. 

Will making startupService a different object other than service makes more sense? So we can manage bootstrap and start there without resolving the services.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: We can certainly do that. I don't think we will gain much by doing this though as calling resolve() will just return the obj if it was already instantiated.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jkasten2: ah right, my mistake. We can probably leave it here then, since it is only 3 lines.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jkasten2: ya maybe we don't add lazy, as `services.getService()` is already doing what `lazy` would be doing any way.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -120,16 +121,27 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
     private var identityModelStore: IdentityModelStore? = null
+        get() = services.getService()
     private var propertiesModelStore: PropertiesModelStore? = null
+        get() = services.getService()
     private var subscriptionModelStore: SubscriptionModelStore? = null
+        get() = services.getService()
     private var startupService: StartupService? = null
+        get() = services.getService()
     private var preferencesService: IPreferencesService? = null
```
jkasten2, line 143: One nice side affect of this change is we can make all these non-null now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
jinliu9508: Good point, I just pushed a commit that stop checking isResolved and schedule start off the main thread for all services. It is much cleaner now. I have tested that this approach works the same.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
```diff
@@ -80,6 +85,19 @@ class ServiceProvider(
         }
     }
 
+    // schedule to start all startable services separate threads
+    fun scheduleStartServices() {
+        val provider = this
+        coroutineScope.launch {
+            synchronized(serviceMap) {
+                for (serviceReg in serviceMap!![IStartableService::class.java]!!) {
+                    val service = serviceReg.resolve(provider) as IStartableService
```
jkasten2, line 94: This looks like a bug, shouldn't we be passing `serviceReg` here instead of `provider`?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -234,21 +240,10 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-            _notifications = services.getService()
-            operationRepo = services.getService()
-            propertiesModelStore = services.getService()
-            identityModelStore = services.getService()
-            subscriptionModelStore = services.getService()
-            preferencesService = services.getService()
-
-            // Instantiate and call the IStartableServices
-            startupService = services.getService()
-            startupService!!.bootstrap()
```
jkasten2, line 251: I believe this was the only spot we were calling `StartupService.bootstrap()`, so we should delete the function as well.

jkasten2 (changes requested):
Left another 2 comments.


Also this existing comment is still unanswered and not addressed:
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2127/files#r1644805601

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
jinliu9508: Yes the newest commit simplifies the code and now directly calls getService.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
jinliu9508: serviceReg.resolve(IServiceProvider) requires a service provider in order resolve serviceReg. The latest commit removes the manual resolve and call getService instead.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: We can probably delete the entire StartupService class since it is no longer used. Should we also remove the StartupServiceTests as well?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
jkasten2: This implementation can replace what we have in `startupService.start()`. We can then also remove `IStartableService` from it's constructor, to avoid the main thread instantiation you are talking about.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
```diff
@@ -85,13 +85,11 @@ class ServiceProvider(
         }
     }
 
-    // schedule to start all startable services separate threads
+    // schedule to start all startable services in a separate thread
     fun scheduleStartServices() {
-        val provider = this
         coroutineScope.launch {
             synchronized(serviceMap) {
```
jkasten2, line 91: Since we are now using `getAllServices()`, I believe we can remove this `synchronized`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jkasten2: > We can probably delete the entire StartupService class since it is no longer used. 

We might want to keep it for what I noted in my other comment:
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2127/files#r1651606179


> Should we also remove the StartupServiceTests as well?

For bootstrap() yes, unless you think we should move this code back there so we can keep the tests. They tests are very basic, so kind of touse up.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
jinliu9508: That's right, synchronized is no longer needed here. I just pushed a change for this.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: I reworked StartupService so it handles calling bootstrap() and scheduleStart(). Also I added a test unit to ensure that scheduleStart does not block the main thread.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/services/ServiceProvider.kt
jinliu9508: StartupService has been reworked so it no longer accepts lists of `IStartableService` and `IBootstrapService`. Instead, it will take a service provider and implement logic for 'bootstrap' and 'scheduleStart'.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
```diff
@@ -1,18 +1,27 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.newSingleThreadContext
+
 internal class StartupService(
-    private val _bootstrapServices: List<IBootstrapService>,
-    private val _startableServices: List<IStartableService>,
+    private val services: ServiceProvider,
 ) {
+    private var coroutineScope = CoroutineScope(newSingleThreadContext(name = "ServiceProvider"))
```
jkasten2, line 11: Should be `val`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
```diff
@@ -1,18 +1,27 @@
...
-    fun start() {
-        // now that we have the params initialized, start everything else up
-        for (startableService in _startableServices)
-            startableService.start()
+    // schedule to start all startable services in a separate thread
+    fun scheduleStart() {
+        coroutineScope.launch {
+            for (service in services.getAllServices<IStartableService>()) {
+                service.start()
+            }
```
jkasten2, line 24: nit, I noticed there are a few updated and new for loops in this PR. Kotlin has `forEach` feature which is a bit shorter. Example for this line:
```kotlin
services.getAllServices<IStartableService>().forEach { it.start() }
````

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt (hunk 10 above)
Comment (jkasten2) [reply OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt:11]: name = should be "StartupService"

**Review by jkasten2 — CHANGES_REQUESTED** [review]

File: OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/startup/StartupServiceTests.kt [hunk 12]
```diff
@@ -68,35 +82,34 @@ class StartupServiceTests : FunSpec({
...
 
-        val startupService = StartupService(listOf(), listOf(mockStartupService1, mockStartupService2))
+        val startupService = StartupService(setupServiceProvider(listOf(), listOf(mockStartupService1, mockStartupService2)))
 
         // When
-        startupService.start()
+        startupService.scheduleStart()
 
         // Then
         verify(exactly = 1) { mockStartupService1.start() }
```
jkasten2, line 91: This will probably fail most of the time since the work in `scheduleStart()` probably won't finish in time.

jkasten2 (changes requested):
Tests are failing on CI, made a comment about one test that will not pass often.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
jinliu9508: Changed

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
jinliu9508: Changed

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/startup/StartupServiceTests.kt
jinliu9508: Hmm for some reason it works on my local environment. I added a short delay before verifying the mocks and it looks like it's working now.

jkasten2 (approved):
Changes look good now, but probably want to rebase to clean up a few a few of the fixup commits.

jkasten2 (approved):

Issue #1969, opened 2024-01-22
[Bug]: SDK Crashes on release/app bundle build when not all onesignal modules are included: Service interface i6.a could not be instantiated


CarlBarks:
### What happened?

The application crashes on startup with the following error. It happens only on release/app bundle builds with proguard enabled and works as expected on debug builds.

```
FATAL EXCEPTION: main
                                                                                                                Process: xxxx, PID: 13210
                                                                                                                java.lang.RuntimeException: Unable to create application xxxx: java.lang.Exception: Service interface i6.a could not be instantiated
                                                                                                                	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6764)
                                                                                                                	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
                                                                                                                	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2133)
                                                                                                                	at android.os.Handler.dispatchMessage(Handler.java:106)
                                                                                                                	at android.os.Looper.loopOnce(Looper.java:201)
                                                                                                                	at android.os.Looper.loop(Looper.java:288)
                                                                                                                	at android.app.ActivityThread.main(ActivityThread.java:7872)
                                                                                                                	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                                	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
                                                                                                                	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:936)
                                                                                                                Caused by: java.lang.Exception: Service interface i6.a could not be instantiated
                                                                                                                	at l5.d.getService(Unknown Source:58)
                                                                                                                	at com.onesignal.internal.a.initWithContext(Unknown Source:257)
                                                                                                                	at j5.b.e(Unknown Source:16)
                                                                                                                	at xxx.onCreate(Unknown Source:80)
                                                                                                                	at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1277)
                                                                                                                	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6759)
                                                                                                                	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0) 
                                                                                                                	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2133) 
                                                                                                                	at android.os.Handler.dispatchMessage(Handler.java:106) 
[...]

PR #2136, opened 2024-06-26
[Fix] "could not be instantiated" exception when; some modules are omitted AND android.enableR8.fullMode=true

jkasten2:
# Description
## One Line Summary
Fix "could not be instantiated" exception when; some modules are omitted AND `android.enableR8.fullMode` is enabled.

## Details
This also fixes the stack traces reported when attempting to use a module that wasn't included 2b2e76ae403a3910073d087aad9b8a48ca36490f

### Motivation
This results in crashes when these two conditions are meet. SDK should include all Proguard rules so app developers don't have to manually add them to avoid issues like this.

### Scope
Only effects keeping a bit more code when minification is enabled.

# Testing
## Manual testing
Tested with the included example app with the following modifications:
* Added `android.enableR8.fullMode=true` to `Examples/OneSignalDemo/gradle.properties`
* Replaced `com.onesignal:OneSignal:5.1.15` with `com.onesignal:core:5.1.15` in `Examples/OneSignalDemo/app/buidle.gradle`
* Modified OneSignalDemo's code no longer call things that require modules;

Reproduced the crash reported in issue #1969 and ensured it no longer happens after the change. Tested on an Android 6.0 emulator.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing
         - CI doesn't currently build or run the demo app.

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2136)

### DIFF
--- a/OneSignalSDK/onesignal/core/consumer-rules.pro
+++ b/OneSignalSDK/onesignal/core/consumer-rules.pro
@@ -14,4 +14,6 @@
 
 -keepclassmembers class com.onesignal.common.** { *; }
 
+-keepclassmembers @com.onesignal.core.internal.minification.KeepStub class * { <init>(...); }
+
 -keep class ** implements com.onesignal.common.modules.IModule { *; }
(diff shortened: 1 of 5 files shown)


jinliu9508 (approved):
LGTM!!!

PR #2137, opened 2024-06-27
[FIX] IAM with dynamic trigger showing forever

jinliu9508:
# Description
## One Line Summary
Fix an issue with IAM with dynamic triggers like "SESSION_TIME" showing up repetitively after being dismissed.

### Motivation
Fix the incorrect app behavior when IAM with both custom trigger and session_time trigger continues to show up even it was dismissed.

### Scope
- This issue happens strictly when both a time based trigger like Session_Time and an in-app trigger of "IS NOT EQUAL" exist with AND. The first time both triggers evaluated to true, IAM shows up, gets dismissed, and will reevaluate the triggers again. In the reevaluation, the system first determines if there is only dynamic trigger. If there is, it forces evaluate to false to suppress the IAM from showing up again. But in this case, another in-app trigger exists and will always evaluate to true due to the IS NOT EQUAL characteristics we currently have, thus a repeating IAM occurs.
- This PR alters the logic on when to redisplay message so that we no longer make redisplay messages available to avoid duplicate redisplay of a same message. A message can be redisplayed if at least one trigger has changed or a new trigger is added when a dynamic trigger exists.

## Manual testing
Before the fix:
IAM setting to reproduce the issue:
image
This IAM would show up 5 seconds after app start up and continue to show up after being dismissed.

After the fix:
1. IAM with whatever IS NOT 5 AND Session_Time > 5 with Every time trigger conditions are satisfied checked:
    - Result: This IAM will show up 5 seconds after app start up and will not show up again once dismissed. Whenever I add a new trigger, the IAM shows up again.
2. IAM with whatever = true AND Session_Time > 5 with Every time trigger conditions are satisfied checked:
    - Result: This IAM will not show up initially, but will show up if I add a trigger whatever = true EVERY TIME after 5 seconds of startup. If I add a trigger with an unrelated key, the IAM will not show.
3. IAM with a single trigger whatever2 = 2
    - Result: This IAM will not show up unless a trigger whatever2 = 2 is added. Any value for the same key that is not equals to 2 will not show the IAM. 
4. IAM with a single dynamic trigger Session_Time > 5
    - Result: This IAM will only show once 5 seconds after startup.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2137)

### DIFF
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
@@ -451,12 +451,16 @@ internal class InAppMessagesManager(
      *
      * Make all messages with redisplay available if:
      * - Already displayed
-     * - At least one Trigger has changed
+     * - At least one existing Trigger has changed OR a new trigger is added
      */
-    private fun makeRedisplayMessagesAvailableWithTriggers(newTriggersKeys: Collection<String>) {
+    private fun makeRedisplayMessagesAvailableWithTriggers(
+        newTriggersKeys: Collection<String>,
+        isNewTriggerAdded: Boolean,
+    ) {
         for (message in messages) {
-            if (!message.isTriggerChanged && redisplayedInAppMessages.contains(message) &&
-                _triggerController.isTriggerOnMessage(message, newTriggersKeys)
+            if (!message.isTriggerChanged &&
+                redisplayedInAppMessages.contains(message) &&
+                (_triggerController.isTriggerOnMessage(message, newTriggersKeys) || isNewTriggerAdded)
             ) {
                 Logging.debug("InAppMessagesManager.makeRedisplayMessagesAvailableWithTriggers: Trigger changed for message: $message")
                 message.isTriggerChanged = true
@@ -643,7 +647,6 @@ internal class InAppMessagesManager(
         Logging.debug("InAppMessagesManager.onTriggerCompleted: called with triggerId: $triggerId")
         val triggerIds: MutableSet<String> = HashSet()
         triggerIds.add(triggerId)
-        makeRedisplayMessagesAvailableWithTriggers(triggerIds)
     }
 
     /**
@@ -653,8 +656,10 @@ internal class InAppMessagesManager(
      *
      * @see OSInAppMessageController.setDataForRedisplay
      */
-    override fun onTriggerConditionChanged() {
+    override fun onTriggerConditionChanged(triggerId: String) {
         Logging.debug("InAppMessagesManager.onTriggerConditionChanged()")
+
+        makeRedisplayMessagesAvailableWithTriggers(listOf(triggerId), false)
 
         suspendifyOnThread {
             // This method is called when a time-based trigger timer fires, meaning the message can
@@ -666,7 +671,7 @@ internal class InAppMessagesManager(
     override fun onTriggerChanged(newTriggerKey: String) {
         Logging.debug("InAppMessagesManager.onTriggerChanged(newTriggerKey: $newTriggerKey)")
 
-        makeRedisplayMessagesAvailableWithTriggers(listOf(newTriggerKey))
+        makeRedisplayMessagesAvailableWithTriggers(listOf(newTriggerKey), true)
 
         suspendifyOnThread {
             // This method is called when a time-based trigger timer fires, meaning the message can
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
@@ -458,13 +458,10 @@ internal class InAppMessagesManager(
         isNewTriggerAdded: Boolean,
     ) {
         for (message in messages) {
-            if (!message.isTriggerChanged &&
-                redisplayedInAppMessages.contains(message) &&
-                (
-                    _triggerController.isTriggerOnMessage(message, newTriggersKeys) ||
-                        isNewTriggerAdded && _triggerController.messageHasOnlyDynamicTriggers(message)
-                )
-            ) {
+            val isMessageDisplayed = redisplayedInAppMessages.contains(message)
+            val isTriggerOnMessage = _triggerController.isTriggerOnMessage(message, newTriggersKeys)
+            val isOnlyDynamicTriggers = _triggerController.messageHasOnlyDynamicTriggers(message)
+            if (!message.isTriggerChanged && isMessageDisplayed && (isTriggerOnMessage || isNewTriggerAdded && isOnlyDynamicTriggers)) {
                 Logging.debug("InAppMessagesManager.makeRedisplayMessagesAvailableWithTriggers: Trigger changed for message: $message")
                 message.isTriggerChanged = true
             }
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
@@ -451,7 +451,7 @@ internal class InAppMessagesManager(
      *
      * Make all messages with redisplay available if:
      * - Already displayed
-     * - At least one existing Trigger has changed OR a new trigger is added
+     * - At least one existing Trigger has changed OR a new trigger is added when there is only dynamic trigger
      */
     private fun makeRedisplayMessagesAvailableWithTriggers(
         newTriggersKeys: Collection<String>,
@@ -460,7 +460,10 @@ internal class InAppMessagesManager(
         for (message in messages) {
             if (!message.isTriggerChanged &&
                 redisplayedInAppMessages.contains(message) &&
-                (_triggerController.isTriggerOnMessage(message, newTriggersKeys) || isNewTriggerAdded)
+                (
+                    _triggerController.isTriggerOnMessage(message, newTriggersKeys) ||
+                        isNewTriggerAdded && _triggerController.messageHasOnlyDynamicTriggers(message)
+                )
             ) {
                 Logging.debug("InAppMessagesManager.makeRedisplayMessagesAvailableWithTriggers: Trigger changed for message: $message")
                 message.isTriggerChanged = true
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -450,13 +450,14 @@ internal class InAppMessagesManager(
...
-     * - Already displayed
+     * - Already displayed in a different session
      * - At least one Trigger has changed
      */
     private fun makeRedisplayMessagesAvailableWithTriggers(newTriggersKeys: Collection<String>) {
         for (message in messages) {
             if (!message.isTriggerChanged && redisplayedInAppMessages.contains(message) &&
-                _triggerController.isTriggerOnMessage(message, newTriggersKeys)
+                _triggerController.isTriggerOnMessage(message, newTriggersKeys) &&
+                !message.isDisplayedInSession
```
jkasten2, line 460: This breaks the "Every time trigger conditions are satisfied" logic so we shouldn't do this.

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
jinliu9508: Ok, I just pushed a different change that only target dynamic trigger firing logic.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/triggers/impl/DynamicTriggerController.kt
```diff
@@ -44,14 +44,31 @@ internal class DynamicTriggerController(
...
+                            _time.currentTimeMillis - lastTimeAppDismissed
+                        }
+                    if (evaluateTimeIntervalWithOperator(
+                            lastTimeDismissedInterval.toDouble(),
+                            currentTimeInterval.toDouble(),
+                            trigger.operatorType,
+                        )
+                    ) {
+                        return false
+                    }
```
jkasten2, line 65: Won't this break the "Every time trigger conditions are satisfied" if the IAM had a custom trigger with AND?

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/triggers/impl/DynamicTriggerController.kt
```diff
@@ -44,14 +44,31 @@ internal class DynamicTriggerController(
             }
 
             var currentTimeInterval: Long = 0
+            val lastTimeAppDismissed = _state.lastTimeInAppDismissed
```
jkasten2, line 47: `lastTimeAppDismissed` is missing `In` in the name, should be `lastTimeInAppDismissed`

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/triggers/impl/DynamicTriggerController.kt
jinliu9508: Yes it does. I pushed a new approach with memorization.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/triggers/impl/DynamicTriggerController.kt
jinliu9508: Outdated due to rework

jkasten2 (changes requested):
This latest solution still looks like it would break the "Every time trigger conditions are satisfied" behavior. We want to keep the behavior where the developer should be able to reshow the IAM by calling `addTrigger` again. Can you test this scenario and add what you tested to the PR notes?

jinliu9508:
> This latest solution still looks like it would also break the "Every time trigger conditions are satisfied" behavior. We want to keep the behavior where the developer should be able to reshow the IAM by calling `addTrigger` again with the same key. Can you test this scenario and add what you tested to the PR notes?

The latest change adds a set that stores evaluated combinations of trigger to prevent a certain combination of trigger from showing multiple time. The set will get cleared every time addTrigger or removeTrigger is called. And it appears in my test that the IAM will show up once per startup and every time I add a new trigger.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/triggers/impl/TriggerController.kt
```diff
@@ -17,6 +17,7 @@ internal class TriggerController(
     private var _dynamicTriggerController: DynamicTriggerController,
 ) : ITriggerController, IModelStoreChangeHandler<TriggerModel> {
     val triggers: ConcurrentHashMap<String?, Any?> = ConcurrentHashMap()
+    val evaluatedCombinations = mutableSetOf<String>()
```
jkasten2, line 20: This should be at the Message level, not global for all IAMs. Otherwise if new IAMs are fetched when the app is running then they won't evaluate to true when they should.

jkasten2 (changes requested):
This new logic has a lot of overlap with the pre-existing `isTriggerChanged` and `isDisplayedInSession` logic. I think we should remove these (if they cover everything) before merging in these changes so we don't have redundant logic that would much harder to reason about later.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/triggers/impl/TriggerController.kt
jkasten2: I think this is probably fine actually, since trigger.id is a UUID.

jinliu9508:
> This new logic has a lot of overlap with the pre-existing `isTriggerChanged` and `isDisplayedInSession` logic. I think we should remove these (if they cover everything) before merging in these changes so we don't have redundant logic that would much harder to reason about later.

Yes, I also realized that this particular approach uses similar logic as 'isTriggerChanged' so I made a different approach that only alter the 'makeRedisplayMessagesAvailableWithTriggers' logic so that onTriggerCompleted no longer redisplay messages. Instead, it was done by onTriggerConditionChanged before trigger evaluation. I have updated the test part in the description to include some common IAM setup that is potentially affected by the change.

jkasten2 (changes requested):
>2. IAM with whatever = true AND Session_Time > 5 with Every time trigger conditions are satisfied checked:

There is still a bug with that IAM in the follow scenario.
1. Call `addTriger("whatever", true)`
2. IAM displays as expected
3. Call `addTriger("somethingElse", true)`
4. BUG: IAM displays again when it should not
   - It is expected that it only re-display if `addTriger("whatever", true)` was called again.

jinliu9508:
> > 2. IAM with whatever = true AND Session_Time > 5 with Every time trigger conditions are satisfied checked:
> 
> There is still a bug with that IAM in the follow scenario.
> 
> 1. Call `addTriger("whatever", true)`
> 2. IAM displays as expected
> 3. Call `addTriger("somethingElse", true)`
> 4. BUG: IAM displays again when it should not
>    
>    * It is expected that it only re-display if `addTriger("whatever", true)` was called again.

Nice found! I have fixed this bug and removed some unneeded code for the latest commit.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -451,12 +451,19 @@ internal class InAppMessagesManager(
...
+        isNewTriggerAdded: Boolean,
+    ) {
         for (message in messages) {
-            if (!message.isTriggerChanged && redisplayedInAppMessages.contains(message) &&
-                _triggerController.isTriggerOnMessage(message, newTriggersKeys)
+            if (!message.isTriggerChanged &&
+                redisplayedInAppMessages.contains(message) &&
+                (
+                    _triggerController.isTriggerOnMessage(message, newTriggersKeys) ||
+                        isNewTriggerAdded && _triggerController.messageHasOnlyDynamicTriggers(message)
```
jkasten2, line 465: This `if` statement is now well over the line of what is considered readable. Let's split this up so it is more readable before we merge this.

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
jinliu9508: Agreed, the if statement is now broken into multiple descriptive boolean values to enhance readability.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -451,13 +451,17 @@ internal class InAppMessagesManager(
      *
      * Make all messages with redisplay available if:
      * - Already displayed
-     * - At least one Trigger has changed
+     * - At least one existing Trigger has changed OR a new trigger is added when there is only dynamic trigger
```
jkasten2, line 454: 👍 on updating the comments!

jkasten2 (approved):
No action needed in this PR, but noted here for the future. I think there is still more states the can be cleaned up to simplify this manager's logic. Some that come to mind are:
1. I don't think `messageHasOnlyDynamicTriggers` needs to be a concept
2. `redisplayedInAppMessages` overlaps with `message.isDisplayedInSession`

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -451,13 +451,17 @@ internal class InAppMessagesManager(
...
-    private fun makeRedisplayMessagesAvailableWithTriggers(newTriggersKeys: Collection<String>) {
+    private fun makeRedisplayMessagesAvailableWithTriggers(
+        newTriggersKeys: Collection<String>,
+        isNewTriggerAdded: Boolean,
+    ) {
         for (message in messages) {
-            if (!message.isTriggerChanged && redisplayedInAppMessages.contains(message) &&
-                _triggerController.isTriggerOnMessage(message, newTriggersKeys)
-            ) {
+            val isMessageDisplayed = redisplayedInAppMessages.contains(message)
```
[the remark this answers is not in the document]
jinliu9508: Ok I just fixed this.

PR #2138, opened 2024-06-28
Allow `preventDefault` to be fired up to two times

nan-li:
# Description
## One Line Summary
Allow `preventDefault` to be fired up to two times with the `discard` parameter of `false` and then `true`, for the Received Event and the Display Event.

## Details
* The issue was that the event is passed to the developer's callback and the `discard` parameter is read when the callback returns. Then the logic continues on with that state.
* So, a second call to `preventDefault(true)` had no effect.
* Modify the notification display waiter to accept this second prevent default call.
* This is necessary for wrappers which may need to call this twice, once to capture the event internally and once from the developer's intent.
* `preventDefault(false)` can be called multiple times and has no effects due to no state change.

### Motivation
To allow https://github.com/OneSignal/OneSignal-Android-SDK/pull/2094 to work for wrappers.

### Scope
Calling preventDefault twice where the first does not discard and the second discards.

# Testing
## Unit testing

- Cleaned up NotificationGenerationProcessorTests helper methods
- Added 2 tests

## Manual testing

1. Test with `NotificationLifecycleListener` for the will display event.
2. Call `preventDefault()`
3. Delay 3 seconds and call `preventDefault(true)`
5. Delay 3 seconds and call `display()`
6. See that the notification will **not** display.
7. Do the same manual test with `NotificationServiceExtension` for the received event.

Also test there is no breaking of current behavior when preventDefault(discard: false) is called followed by display, will display the notification.

```java
@Override
public void onWillDisplay(@NonNull INotificationWillDisplayEvent event) {
    Log.v(Tag.LOG_TAG, "APP onWillDisplay fired with event: " + event);

    IDisplayableNotification notification = event.getNotification();
    JSONObject data = notification.getAdditionalData();

    event.preventDefault();

    Runnable preventDefaultAgain = () -> {
        try {
            Thread.sleep(3000);
        } catch (InterruptedException ignored) { }
        Log.v(Tag.LOG_TAG, "Dev App INotificationLifecycleListener calling prevent default: discard = true" );
        event.preventDefault(true);
    };

    Runnable callDisplay = () -> {
        try {
            Thread.sleep(6000);
        } catch (InterruptedException ignored) { }
        Log.v(Tag.LOG_TAG, "Dev App INotificationLifecycleListener calling DISPLAY");
        notification.display();
    };

    Thread t1 = new Thread(preventDefaultAgain);
    t1.start();
    Thread t2 = new Thread(callDisplay);
    t2.start();
}
```

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2138)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/INotificationReceivedEvent.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/INotificationReceivedEvent.kt
@@ -53,6 +53,9 @@ interface INotificationReceivedEvent {
 
     /**
      * Call this to prevent OneSignal from displaying the notification automatically.
+     * This method can be called up to two times with false and then true, if processing time is needed.
+     * Typically this is only possible within a short
+     * time-frame (~30 seconds) after the notification is received on the device.
      * @param discard an [preventDefault] set to true to dismiss the notification with no
      * possibility of displaying it in the future.
      */
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/INotificationWillDisplayEvent.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/INotificationWillDisplayEvent.kt
@@ -41,6 +41,9 @@ interface INotificationWillDisplayEvent {
 
     /**
      * Call this to prevent OneSignal from displaying the notification automatically.
+     * This method can be called up to two times with false and then true, if processing time is needed.
+     * Typically this is only possible within a short
+     * time-frame (~30 seconds) after the notification is received on the device.
      * @param discard an [preventDefault] set to true to dismiss the notification with no
      * possibility of displaying it in the future.
      */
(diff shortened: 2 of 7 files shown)


jennantilla (approved):

jkasten2 (approved):
Looks good to me to account for the wrapper use-case. Would be a nice to have if `preventDefault(false)` could be called any number of times, but that is pretty minor and shouldn't hold up getting this PR out there.

nan-li:
`preventDefault(false)` can be called multiple times, and basically no-op since there is no changes in state.
The waiter will wake once it is called with `discard = true`

jkasten2 (approved):

PR #2151, opened 2024-07-16
[Optimization] Init with context in background

jinliu9508:
# Description
## One Line Summary
Move start call of unused services off the main thread to make the initWithContext more efficient.

### Motivation
Our SDK initialization is currently taking longer to finish compare to some other SDKs. One of the reason is we construct service components like Location Manager or IAM Manager during the initialization phrase when they are not really needed until later. This PR aims to move the initialization of these services to background thread so they don't slow down the overall SDK initialization.

### Scope
The retrieval of services that were changed to initialize in background is not longer from a saved instance in OneSignalImp.kt. Instead, each of these components has a getter that retrieve its instance from the service provider. 

# Testing
## Manual testing
* Bench mark tests: I measure how long the initialization takes by tracking the start and finish time when 'initWithContext' is called. In my testing in a pixel 3a API 34 emulator, it took an average of 600 ms to complete initialization before the change and around 300 ms after the change, which is around ~50% faster. 
* Normal SDK behavior test: Send notification, add tags, and outcome work.
* Calling 'initWithContext' in a background thread: no error and the demo app works as normal. Other OneSignal calls can only work if they are called after 'initWithContext' and in the same thread.
* Calling 'requestPermission' in a background thread: a test case is added in MainApplication to ensure the app won't crash by calling requestPermission in a thread right after initWithContext.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2151)

### DIFF
--- a/Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplication.java
+++ b/Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplication.java
@@ -1,12 +1,13 @@
 package com.onesignal.sdktest.application;
 
+import android.annotation.SuppressLint;
 import android.os.StrictMode;
 import android.util.Log;
-
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.multidex.MultiDexApplication;
 
+import com.onesignal.Continue;
 import com.onesignal.OneSignal;
 import com.onesignal.inAppMessages.IInAppMessageClickListener;
 import com.onesignal.inAppMessages.IInAppMessageClickEvent;
@@ -28,8 +29,10 @@ import com.onesignal.sdktest.util.SharedPreferenceUtil;
 import com.onesignal.user.state.IUserStateObserver;
 import com.onesignal.user.state.UserChangedState;
 import com.onesignal.user.state.UserState;
-
 import org.json.JSONObject;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 public class MainApplication extends MultiDexApplication {
     private static final int SLEEP_TIME_TO_MIMIC_ASYNC_OPERATION = 2000;
@@ -40,6 +43,7 @@ public class MainApplication extends MultiDexApplication {
             StrictMode.enableDefaults();
     }
 
+    @SuppressLint("NewApi")
     @Override
     public void onCreate() {
         super.onCreate();
@@ -54,7 +58,17 @@ public class MainApplication extends MultiDexApplication {
         }
 
         OneSignalNotificationSender.setAppId(appId);
+
         OneSignal.initWithContext(this, appId);
+
+        // Ensure calling requestPermission in a thread right after initWithContext does not crash
+        // This will reproduce result similar to Kotlin CouroutineScope.launch{}, which may potentially crash the app
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        @SuppressLint({"NewApi", "LocalSuppress"}) CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
+            OneSignal.getNotifications().requestPermission(true, Continue.none());
+        }, executor);
+        future.join(); // Waits for the task to complete
+        executor.shutdown();
 
         OneSignal.getInAppMessages().addLifecycleListener(new IInAppMessageLifecycleListener() {
             @Override
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
@@ -1,18 +1,18 @@
 package com.onesignal.core.internal.startup
 
+import com.onesignal.common.services.ServiceProvider
+
 internal class StartupService(
-    private val _bootstrapServices: List<IBootstrapService>,
-    private val _startableServices: List<IStartableService>,
+    private val services: ServiceProvider,
 ) {
     fun bootstrap() {
-        // now that we have the params initialized, start everything else up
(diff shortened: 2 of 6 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
```diff
@@ -1,18 +1,23 @@
...
+        services.getAllServices<IBootstrapService>().forEach { it.bootstrap() }
     }
 
-    fun start() {
-        // now that we have the params initialized, start everything else up
-        for (startableService in _startableServices)
-            startableService.start()
+    // schedule to start all startable services in a separate thread
+    fun scheduleStart() {
+        coroutineScope.launch {
```
jkasten2, line 19: Since we only need to run a one off background task can we use a `Thread` instead of a `CoroutineScope`? This has a bit less overhead.

jinliu9508: Sure. By using thread, we can avoid creating coroutineScope. 

But I think Coroutine is a simpler version of Thread as it does not use as much resources going down to the OS threads and only take an object in the JVM heap. In my test, the time it takes for the code inside the block to complete by using coroutineScope.launch vs using thread is very similar. Using thread is a tiny bit slower in average. This may be affected by the current os responsiveness as well. 

Another reason I used coroutine is that this practice is also used in other places like operationRepo so there is some consistency. Do we think it is worth changing it to using thread?

jkasten2 (approved):
I looked over all the classes that implement `IStartableService` and can't see any issues with their `start()` method from being called delayed / called from the background.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
jkasten2: In this specific case we were creating a `newSingleThreadContext` and just running one task ever, the OperationRepo stays alive forever and continually takes work. In this class we simply need to run something in the background and then we no longer need the thread.

We could optimize this in future by using any existing coroutineScope that is shared with other parts of the SDK, but since we don't have that it is more efficient to use a thread.

> In my test, the time it takes for the code inside the block to complete by using coroutineScope.launch vs using thread is very similar.

Did you include `CoroutineScope(newSingleThreadContext(name = "StartupService"))` and `coroutineScope.launch` in your test vs `Thread {}.start()`?

jinliu9508: Ok the newest commit has changed to use 'Thread.start()' instead of using 'coroutineScope'. I placed a mark right before the initialization of 'StartupService', a mark right before 'launch'/'Thread.start', and the final mark at the completion. The diff is insignificant in general.

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplication.java
```diff
@@ -54,8 +58,17 @@ public void onCreate() {
         }
 
         OneSignalNotificationSender.setAppId(appId);
+
         OneSignal.initWithContext(this, appId);
 
+        // ensure calling requestPermission in a thread right after initWithContext does not crash
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        @SuppressLint({"NewApi", "LocalSuppress"}) CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
+            OneSignal.getNotifications().requestPermission(true, Continue.none());
```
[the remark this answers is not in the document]
jinliu9508: If I check out the second commit and use this piece of code in MainApplication.java, I am not able to reproduce the crash anymore. 
```
// ensure calling requestPermission in a thread right after initWithContext does not crash
        new Thread(new Runnable() {
            @Override
            public void run() {
                OneSignal.getNotifications().requestPermission(true, Continue.none());
            }
        }){}.start();
```

Reports from user are suggesting that
```
CoroutineScope(Dispatchers.IO).launch {
  OneSignal.Notifications.requestPermission(true)
}
``` is what causes the crash. 

Although both Java Thread and Kotlin CoroutineScope are used for handling concurrency, they seem to operate differently. The only equivalence in Java I found to reproduce the same crash is by calling java8 CompletableFuture.join() I guess the reason is that Java join() waits for the thread to complete, which is similar to Kotlin Coroutine that can be suspended. 

I added a comment to further explain the reason CompletableFuture and its related annotation is added here.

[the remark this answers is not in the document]
Comment (nan-li) [reply Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplication.java:68]: An alternative is to add a Kotlin object file to the example app that has a function to run 
```
CoroutineScope(Dispatchers.IO).launch { 
   OneSignal.Notifications.requestPermission(true) 
}
```
and then call it from MainApplication.

jinliu9508:
Force push: Consolidating the commits. Code is unchanged since last review.

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplication.java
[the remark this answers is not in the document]
jinliu9508: Good point. I have tried this approach as well, and they behave the same in my test. I chose the Java code here considering "broader" test coverage from other wrappers that run Java code over the bridge.

jkasten2 (approved):

PR #2157, opened 2024-07-27
Recover null onesignal ID crashes for Operations

nan-li:
# Description
## One Line Summary
Don't uncache invalid Operations that are missing onesignalId in order to prevent repetitive crashes when they go on to be processed.

### Motivation
Crash report showing high-volume repetitive crashes happening for a low volume of users.
- Crash report shows `onesignalId` is `null` for an UpdateSubscriptionOperation
- Recovers those users
- Does _not_ prevent the first crash. This is rare and only reported by one app, so we don't want to hide any issues if we start receiving increased reports of this crash. This fix will prevent the repetitive nature of the crash and allow recovery.
- **Note:** this repetitive crash should not typically happen as this Operation should not be instantiated in the first place due to null checks, but this appears to have happened for the reporting app due to more aggressive minification that removes all Kotlin enforced null checks

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.startsWith(java.lang.String)' on a null object reference
       at kotlin.text.StringsKt__StringsJVMKt.startsWith(StringsJVM.kt:421)
       at kotlin.text.StringsKt__StringsJVMKt.startsWith$default(StringsJVM.kt:419)
       at com.onesignal.common.IDManager.isLocalId(IDManager.kt:28)
       at com.onesignal.user.internal.operations.UpdateSubscriptionOperation.getCanStartExecute(UpdateSubscriptionOperation.kt:87)
        ...
```

### Investigation
- Not able to reproduce a scenario where `UpdateSubscriptionOperation` would get created without a `onesignalId`
- This could mean a subscription property was changed before `initWithContext` finishes setting the local onesignalId
- This could also mean the identity model's onesignalId is set to from nonnull to null at the moment this operation is created
- This could also be the identity model in the store is overwritten with an identity model that is missing onesignalId or by a blank Identity Model
- I was not able to recreate any of the above scenarios naturally

### Scope
Basically drop any Operation that is invalid and would not be sendable.

# Testing
## Unit testing
- Add new unit test that uncaches valid and invalid Operations

## Manual testing
**Device: Emulator Pixel 3 API 33**
- Limited ability to test manually due to inability to reproduce
- Attempted multiple paths of reproduction naturally
- Force reproduction by instantiating blank `UpdateSubscriptionOperation` directly

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2157)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationModelStore.kt
@@ -38,8 +38,7 @@ internal class OperationModelStore(prefs: IPreferencesService) : ModelStore<Oper
             return null
         }
 
-        if (!jsonObject.has(Operation::name.name)) {
-            Logging.error("jsonObject must have '${Operation::name.name}' attribute")
+        if (!isValidOperation(jsonObject)) {
             return null
         }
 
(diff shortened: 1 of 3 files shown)


nan-li:
Thoughts on letting the first crash through when the Operation is first processed in the OpRepo? 
Noting again that under normal circumstances, the Operation would not even be instantiated in the first place.

jinliu9508 (commented):
Code and the logic looks good to me and nice job on the testing unit! 

I just want to confirm that the 'shouldRePersist' logic under ModelStore.load() will not persist the invalid operation in cases where the operation model store contains only invalid operation(s). If the bad data get to persist and we let the first crash through, I think the bad data won't get drop out of the persistent store and will cause another crash in next launch.

nan-li:
> Code and the logic looks good to me and nice job on the testing unit!
> 
> I just want to confirm that the 'shouldRePersist' logic under ModelStore.load() will not persist the invalid operation in cases where the operation model store contains only invalid operation(s). If the bad data get to persist and we let the first crash through, I think the bad data won't get drop out of the persistent store and will cause another crash in next launch.

The bad operation won't be loaded and created so when the operations are all persisted again, it wont be there.

jinliu9508 (approved):

PR #2158, opened 2024-07-30
Fix: Don't retry IAM display if 410 is received from backend

jennantilla:
# Description
## One Line Summary
Don't retry IAM display if 410 is received from backend

### Motivation
In instances where an IAM is disabled mid-session, a 410 is received and is caught in a retrying loop. This change will prevent the retry and just not display the IAM, since it just become disabled.

### Scope
Changes to `InAppMessagesManager` only.

Since this change involves changing the `messages` list, synchronization was added to `makeRedisplayMessagesAvailableWithTriggers` and `attemptToShowInAppMessage` to prevent concurrent modification issues. `evaluateInAppMessages` was further updated to collect messages for display inside a synchronized block and process them outside to avoid suspension points within critical sections.

# Testing

## Manual testing
Forced reproduction by setting breakpoint right before IAM is set to display, pausing the IAM from the dashboard, then resuming app flow. A 410 will result.

After updating the code, tested on an Android 14 emulator to see that app resumes normally after IAM is paused in the dashboard. Subsequent 410s are not received and app resumes as expected.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [X] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2158)

### DIFF
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
@@ -106,8 +106,6 @@ internal class InAppMessagesManager(
     private val fetchIAMMutex = Mutex()
     private var lastTimeFetchedIAMs: Long? = null
 
-    private val lock = Any()
-
     override var paused: Boolean
         get() = _state.paused
         set(value) {
@@ -269,7 +267,7 @@ internal class InAppMessagesManager(
         Logging.debug("InAppMessagesManager.evaluateInAppMessages()")
         val messagesToQueue = mutableListOf<InAppMessage>()
 
-        synchronized(lock) {
+        synchronized(messages) {
             for (message in messages) {
                 if (_triggerController.evaluateMessageTriggers(message)) {
                     setDataForRedisplay(message)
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -106,6 +106,8 @@ internal class InAppMessagesManager(
     private val fetchIAMMutex = Mutex()
     private var lastTimeFetchedIAMs: Long? = null
 
+    private val lock = Any()
```
jinliu9508, line 109: It looks like this lock is only for synchronizing 'messages'. In this case we can just call 'synchronized(messages)' instead of introducing a temp lock instance.

jinliu9508 (commented):
Overall, the change works and I am no longer getting repeating 410 errors. Nice work!

jinliu9508 (approved):

PR #2168, opened 2024-08-06
Read-Your-Write Consistency

rgomezp:
# Description

## One Line Summary
Introduce and integrate a ConsistencyManager to manage read-your-write tokens for improved segment membership calculation.

### Motivation
This update introduces the **ConsistencyManager** to manage kafka offsets, which are used as read-your-write tokens. The goal is to improve the accuracy of segment membership calculations by providing an open-ended & highly customizable blocking mechanism for operations that rely on having successfully synchronized client & server state.

For a first use-case, we want to block the fetching of IAMs until we have offsets for a user & subscription state update.

### Scope
- Registers the **ConsistencyManager** in the User and InAppMessages modules.
- Updates the Subscription and User backend services to return offsets needed by the **ConsistencyManager**.
- Modifies Subscription and UpdateUser operation executors to set offsets in the **ConsistencyManager**.
- Updates the IAM manager and backend service to include the offset as a query parameter in the `listInAppMessages` request.

# Testing
## Unit testing
- New coverage for new `ConsistencyManager` class

## Manual testing
Tested on Android emulator
- Set up Android test app in OneSignal
- Note the current app version (e.g. 1.0)
- Create a segment targeting a specific version (e.g. 1.1)
- Create an in-app using the new segment
- Open the app (see you don't get an in-app)
- Close the app
- Change the version to the one in the segment (e.g. 1.1)
- Open the app (see you get the in-app)

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2168)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
@@ -23,17 +23,14 @@ class IamFetchReadyCondition(
     override fun isMet(indexedTokens: Map<String, Map<IConsistencyKeyEnum, String>>): Boolean {
         val tokenMap = indexedTokens[key] ?: return false
         val userUpdateTokenSet = tokenMap[IamFetchRywTokenKey.USER] != null
-        val subscriptionUpdateTokenSet = tokenMap[IamFetchRywTokenKey.SUBSCRIPTION] != null
 
         /**
          * We always update the session count so we know we will have a userUpdateToken. We don't
          * necessarily make a subscriptionUpdate call on every session. The following logic
-         * is written in a way so that if somehow the subscriptionUpdateToken is set *before* the
-         * userUpdateToken, we will wait for the userUpdateToken to also be set. This is because
-         * we know that a userUpdate call was made and both user & subscription properties are
-         * considered during segment calculations.
+         * doesn't consider tokenMap[IamFetchRywTokenKey.SUBSCRIPTION] for this reason. This doesn't
+         * mean it isn't considered if present when doing the token comparison.
          */
-        return (userUpdateTokenSet && subscriptionUpdateTokenSet) || userUpdateTokenSet
+        return userUpdateTokenSet
     }
 
     override fun getNewestToken(indexedTokens: Map<String, Map<IConsistencyKeyEnum, String?>>): String? {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
@@ -1,0 +1,44 @@
+package com.onesignal.common.consistency
+
+import com.onesignal.common.consistency.enums.IamFetchRywTokenKey
+import com.onesignal.common.consistency.models.ICondition
+import com.onesignal.common.consistency.models.IConsistencyKeyEnum
+
+/**
+ * Used for read your write consistency when fetching In-App Messages.
+ *
+ * Params:
+ *  key : String - the index of the RYW token map
+ */
+class IamFetchReadyCondition(
+    private val key: String,
+) : ICondition {
+    companion object {
+        const val ID = "IamFetchReadyCondition"
+    }
+
+    override val id: String
+        get() = ID
+
+    override fun isMet(indexedTokens: Map<String, Map<IConsistencyKeyEnum, String>>): Boolean {
+        val tokenMap = indexedTokens[key] ?: return false
+        val userUpdateTokenSet = tokenMap[IamFetchRywTokenKey.USER] != null
+        val subscriptionUpdateTokenSet = tokenMap[IamFetchRywTokenKey.SUBSCRIPTION] != null
+
+        /**
+         * We always update the session count so we know we will have a userUpdateToken. We don't
+         * necessarily make a subscriptionUpdate call on every session. The following logic
+         * is written in a way so that if somehow the subscriptionUpdateToken is set *before* the
+         * userUpdateToken, we will wait for the userUpdateToken to also be set. This is because
+         * we know that a userUpdate call was made and both user & subscription properties are
+         * considered during segment calculations.
+         */
+        return (userUpdateTokenSet && subscriptionUpdateTokenSet) || userUpdateTokenSet
+    }
+
+    override fun getNewestToken(indexedTokens: Map<String, Map<IConsistencyKeyEnum, String?>>): String? {
+        val tokenMap = indexedTokens[key] ?: return null
+        // maxOrNull compares lexicographically
+        return listOfNotNull(tokenMap[IamFetchRywTokenKey.USER], tokenMap[IamFetchRywTokenKey.SUBSCRIPTION]).maxOrNull()
+    }
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/enums/IamFetchRywTokenKey.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/enums/IamFetchRywTokenKey.kt
@@ -1,0 +1,12 @@
+package com.onesignal.common.consistency.enums
+
+import com.onesignal.common.consistency.models.IConsistencyKeyEnum
+
+/**
+ * Each enum is a key that we use to keep track of read-your-write tokens.
+ * Although the enums are named with "UPDATE", they serve as keys for tokens from both PATCH & POST
+ */
+enum class IamFetchRywTokenKey : IConsistencyKeyEnum {
+    USER,
+    SUBSCRIPTION,
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
@@ -1,0 +1,92 @@
+package com.onesignal.common.consistency.impl
+
+import com.onesignal.common.consistency.models.ICondition
+import com.onesignal.common.consistency.models.IConsistencyKeyEnum
+import com.onesignal.common.consistency.models.IConsistencyManager
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+/**
+ * Manages read-your-write tokens for more accurate segment membership
+ * calculation. Uses customizable conditions that block retrieval of the newest token until met.
+ *
+ * Usage:
+ *  val consistencyManager = ConsistencyManager<MyEnum>()
+ *  val updateConditionDeferred = consistencyManager.registerCondition(MyCustomCondition())
+ *  val rywToken = updateConditionDeferred.await()
+ */
+class ConsistencyManager : IConsistencyManager {
+    private val mutex = Mutex()
+    private val indexedTokens: MutableMap<String, MutableMap<IConsistencyKeyEnum, String>> = mutableMapOf()
+    private val conditions: MutableList<Pair<ICondition, CompletableDeferred<String?>>> =
+        mutableListOf()
+
+    /**
+     * Set method to update the token based on the key.
+     *  Params:
+     *      id: String - the index of the token map (e.g. onesignalId)
+     *      key: K - corresponds to the operation for which we have a read-your-write token
+     *      value: String? - the token (read-your-write token)
+     */
+    override suspend fun setRywToken(
+        id: String,
+        key: IConsistencyKeyEnum,
+        value: String,
+    ) {
+        mutex.withLock {
+            val rywTokens = indexedTokens.getOrPut(id) { mutableMapOf() }
+            rywTokens[key] = value
+            checkConditionsAndComplete()
+        }
+    }
+
+    /**
+     * Register a condition with its corresponding deferred action. Returns a deferred condition.
+     */
+    override suspend fun registerCondition(condition: ICondition): CompletableDeferred<String?> {
+        mutex.withLock {
+            val deferred = CompletableDeferred<String?>()
+            val pair = Pair(condition, deferred)
+            conditions.add(pair)
+            checkConditionsAndComplete()
+            return deferred
+        }
+    }
+
+    override suspend fun resolveConditionsWithID(id: String) {
+        val completedConditions = mutableListOf<Pair<ICondition, CompletableDeferred<String?>>>()
+
+        for ((condition, deferred) in conditions) {
+            if (condition.id == id) {
+                if (!deferred.isCompleted) {
+                    deferred.complete(null)
+                }
+            }
+            completedConditions.add(Pair(condition, deferred))
+        }
+
+        // Remove completed conditions from the list
+        conditions.removeAll(completedConditions)
+    }
+
+    /**
+     * IMPORTANT: calling code should be protected by mutex to avoid potential inconsistencies
+     */
+    private fun checkConditionsAndComplete() {
+        val completedConditions = mutableListOf<Pair<ICondition, CompletableDeferred<String?>>>()
+
+        for ((condition, deferred) in conditions) {
+            if (condition.isMet(indexedTokens)) {
+                val newestToken = condition.getNewestToken(indexedTokens)
+                if (!deferred.isCompleted) {
+                    deferred.complete(newestToken)
+                }
+                completedConditions.add(Pair(condition, deferred))
+            }
+        }
+
+        // Remove completed conditions from the list
+        conditions.removeAll(completedConditions)
+    }
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
@@ -1,5 +1,7 @@
 package com.onesignal.user
 
+import com.onesignal.common.consistency.impl.ConsistencyManager
+import com.onesignal.common.consistency.models.IConsistencyManager
 import com.onesignal.common.modules.IModule
 import com.onesignal.common.services.ServiceBuilder
 import com.onesignal.core.internal.operations.IOperationExecutor
@@ -34,6 +36,9 @@ import com.onesignal.user.internal.subscriptions.impl.SubscriptionManager
 
 internal class UserModule : IModule {
     override fun register(builder: ServiceBuilder) {
+        // Consistency
+        builder.register<ConsistencyManager>().provides<IConsistencyManager>()
+
         // Properties
         builder.register<PropertiesModelStore>().provides<PropertiesModelStore>()
         builder.register<PropertiesModelStoreListener>().provides<IBootstrapService>()
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
@@ -6,6 +6,9 @@ import com.onesignal.common.DeviceUtils
 import com.onesignal.common.NetworkUtils
 import com.onesignal.common.OneSignalUtils
 import com.onesignal.common.RootToolsInternalMethods
+import com.onesignal.common.consistency.IamFetchReadyCondition
+import com.onesignal.common.consistency.enums.IamFetchRywTokenKey
+import com.onesignal.common.consistency.models.IConsistencyManager
 import com.onesignal.common.exceptions.BackendException
 import com.onesignal.common.modeling.ModelChangeTags
 import com.onesignal.core.internal.application.IApplicationService
@@ -39,6 +42,7 @@ internal class SubscriptionOperationExecutor(
     private val _configModelStore: ConfigModelStore,
     private val _buildUserService: IRebuildUserService,
     private val _newRecordState: NewRecordsState,
+    private val _consistencyManager: IConsistencyManager,
 ) : IOperationExecutor {
     override val operations: List<String>
         get() = listOf(CREATE_SUBSCRIPTION, UPDATE_SUBSCRIPTION, DELETE_SUBSCRIPTION, TRANSFER_SUBSCRIPTION)
@@ -101,13 +105,22 @@ internal class SubscriptionOperationExecutor(
                     AndroidUtils.getAppVersion(_applicationService.appContext),
                 )
 
-            val backendSubscriptionId =
+            val result =
                 _subscriptionBackend.createSubscription(
                     createOperation.appId,
                     IdentityConstants.ONESIGNAL_ID,
                     createOperation.onesignalId,
                     subscription,
                 ) ?: return ExecutionResponse(ExecutionResult.SUCCESS)
+
+            val backendSubscriptionId = result.first
+            val rywToken = result.second
+
+            if (rywToken != null) {
+                _consistencyManager.setRywToken(createOperation.onesignalId, IamFetchRywTokenKey.SUBSCRIPTION, rywToken)
+            } else {
+                _consistencyManager.resolveConditionsWithID(IamFetchReadyCondition.ID)
+            }
 
             // update the subscription model with the new ID, if it's still active.
             val subscriptionModel = _subscriptionModelStore.get(createOperation.subscriptionId)
@@ -175,7 +188,13 @@ internal class SubscriptionOperationExecutor(
                     AndroidUtils.getAppVersion(_applicationService.appContext),
                 )
 
-            _subscriptionBackend.updateSubscription(lastOperation.appId, lastOperation.subscriptionId, subscription)
+            val rywToken = _subscriptionBackend.updateSubscription(lastOperation.appId, lastOperation.subscriptionId, subscription)
+
+            if (rywToken != null) {
+                _consistencyManager.setRywToken(startingOperation.onesignalId, IamFetchRywTokenKey.SUBSCRIPTION, rywToken)
+            } else {
+                _consistencyManager.resolveConditionsWithID(IamFetchReadyCondition.ID)
+            }
         } catch (ex: BackendException) {
             val responseType = NetworkUtils.getResponseStatusType(ex.statusCode)
 
@@ -216,6 +235,7 @@ internal class SubscriptionOperationExecutor(
         return ExecutionResponse(ExecutionResult.SUCCESS)
     }
 
+    // TODO: whenever the end-user changes users, we need to add the read-your-write token here, currently no code to handle the re-fetch IAMs
     private suspend fun transferSubscription(startingOperation: TransferSubscriptionOperation): ExecutionResponse {
         try {
             _subscriptionBackend.transferSubscription(
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/UpdateUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/UpdateUserOperationExecutor.kt
@@ -1,6 +1,9 @@
 package com.onesignal.user.internal.operations.impl.executors
 
 import com.onesignal.common.NetworkUtils
+import com.onesignal.common.consistency.IamFetchReadyCondition
+import com.onesignal.common.consistency.enums.IamFetchRywTokenKey
+import com.onesignal.common.consistency.models.IConsistencyManager
 import com.onesignal.common.exceptions.BackendException
 import com.onesignal.common.modeling.ModelChangeTags
 import com.onesignal.core.internal.operations.ExecutionResponse
@@ -31,12 +34,13 @@ internal class UpdateUserOperationExecutor(
     private val _propertiesModelStore: PropertiesModelStore,
     private val _buildUserService: IRebuildUserService,
     private val _newRecordState: NewRecordsState,
+    private val _consistencyManager: IConsistencyManager,
 ) : IOperationExecutor {
     override val operations: List<String>
         get() = listOf(SET_TAG, DELETE_TAG, SET_PROPERTY, TRACK_SESSION_START, TRACK_SESSION_END, TRACK_PURCHASE)
 
-    override suspend fun execute(ops: List<Operation>): ExecutionResponse {
-        Logging.log(LogLevel.DEBUG, "UpdateUserOperationExecutor(operation: $ops)")
+    override suspend fun execute(operations: List<Operation>): ExecutionResponse {
+        Logging.log(LogLevel.DEBUG, "UpdateUserOperationExecutor(operation: $operations)")
 
         var appId: String? = null
         var onesignalId: String? = null
@@ -45,7 +49,7 @@ internal class UpdateUserOperationExecutor(
         var deltasObject = PropertiesDeltasObject()
         var refreshDeviceMetadata = false
 
-        for (operation in ops) {
+        for (operation in operations) {
             when (operation) {
                 is SetTagOperation -> {
                     if (appId == null) {
@@ -83,7 +87,8 @@ internal class UpdateUserOperationExecutor(
                     // that exist in this group.
                     val sessionCount = if (deltasObject.sessionCount != null) deltasObject.sessionCount!! + 1 else 1
 
-                    deltasObject = PropertiesDeltasObject(deltasObject.sessionTime, sessionCount, deltasObject.amountSpent, deltasObject.purchases)
+                    deltasObject =
+                        PropertiesDeltasObject(deltasObject.sessionTime, sessionCount, deltasObject.amountSpent, deltasObject.purchases)
                     refreshDeviceMetadata = true
                 }
                 is TrackSessionEndOperation -> {
@@ -94,9 +99,15 @@ internal class UpdateUserOperationExecutor(
 
                     // The session time we pass up is the total session time across all `TrackSessionEndOperation`
                     // operations that exist in this group.
-                    val sessionTime = if (deltasObject.sessionTime != null) deltasObject.sessionTime!! + operation.sessionTime else operation.sessionTime
+                    val sessionTime =
+                        if (deltasObject.sessionTime != null) {
+                            deltasObject.sessionTime!! + operation.sessionTime
+                        } else {
+                            operation.sessionTime
+                        }
 
-                    deltasObject = PropertiesDeltasObject(sessionTime, deltasObject.sessionCount, deltasObject.amountSpent, deltasObject.purchases)
+                    deltasObject =
+                        PropertiesDeltasObject(sessionTime, deltasObject.sessionCount, deltasObject.amountSpent, deltasObject.purchases)
                 }
                 is TrackPurchaseOperation -> {
                     if (appId == null) {
@@ -107,7 +118,12 @@ internal class UpdateUserOperationExecutor(
                     // The amount spent we pass up is the total amount spent across all `TrackPurchaseOperation`
                     // operations that exist in this group, while the purchases is the union of all
                     // `TrackPurchaseOperation` operations that exist in this group.
-                    val amountSpent = if (deltasObject.amountSpent != null) deltasObject.amountSpent!! + operation.amountSpent else operation.amountSpent
+                    val amountSpent =
+                        if (deltasObject.amountSpent != null) {
+                            deltasObject.amountSpent!! + operation.amountSpent
+                        } else {
+                            operation.amountSpent
+                        }
                     val purchasesArray = if (deltasObject.purchases != null) deltasObject.purchases!!.toMutableList() else mutableListOf()
 
                     for (purchase in operation.purchases) {
@@ -122,18 +138,25 @@ internal class UpdateUserOperationExecutor(
 
         if (appId != null && onesignalId != null) {
             try {
-                _userBackend.updateUser(
-                    appId,
-                    IdentityConstants.ONESIGNAL_ID,
-                    onesignalId,
-                    propertiesObject,
-                    refreshDeviceMetadata,
-                    deltasObject,
-                )
+                val rywToken =
+                    _userBackend.updateUser(
+                        appId,
+                        IdentityConstants.ONESIGNAL_ID,
+                        onesignalId,
+                        propertiesObject,
+                        refreshDeviceMetadata,
+                        deltasObject,
+                    )
+
+                if (rywToken != null) {
+                    _consistencyManager.setRywToken(onesignalId, IamFetchRywTokenKey.USER, rywToken)
+                } else {
+                    _consistencyManager.resolveConditionsWithID(IamFetchReadyCondition.ID)
+                }
 
                 if (_identityModelStore.model.onesignalId == onesignalId) {
                     // go through and make sure any properties are in the correct model state
-                    for (operation in ops) {
+                    for (operation in operations) {
                         when (operation) {
                             is SetTagOperation ->
                                 _propertiesModelStore.model.tags.setStringProperty(
--- a/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
+++ b/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
@@ -4,6 +4,8 @@ import android.app.AlertDialog
 import com.onesignal.common.AndroidUtils
 import com.onesignal.common.IDManager
 import com.onesignal.common.JSONUtils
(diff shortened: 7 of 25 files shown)


michael-winkler:
What are you coding here? 
There is no PR description.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
```diff
@@ -0,0 +1,61 @@
...
+        return deferred
+    }
+
+    private fun checkConditionsAndComplete() {
+        for ((condition, deferred) in conditions) {
+            if (condition.isMet(indexedOffsets)) {
+                val newestOffset = condition.getNewestOffset(indexedOffsets)
+                if (!deferred.isCompleted) {
+                    deferred.complete(newestOffset)
+                }
```
rgomezp, line 57: can we remove after completion?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Manages offsets that function as read-your-write tokens for more accurate segment membership
+ * calculation. Uses customizable conditions that block retrieval of the newest offset until met.
+ *
+ * Usage:
+ *  val consistencyManager = ConsistencyManager()
+ *  val updateConditionDeferred = consistencyManager.registerCondition(MyCustomCondition())
+ *  val newestUpdateOffset = updateConditionDeferred.await()
+ */
+class ConsistencyManager : IConsistencyManager {
```
rgomezp, line 19: Add generic type to replace `OffsetKey` to decouple logic

e.g. `class ConsistencyManager<T>`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
```diff
@@ -0,0 +1,61 @@
...
+     */
+    override fun registerCondition(condition: ICondition): CompletableDeferred<Long?> {
+        val deferred = CompletableDeferred<Long?>()
+        val pair = Pair(condition, deferred)
+        conditions.add(pair)
+        checkConditionsAndComplete()
+        return deferred
+    }
+
+    private fun checkConditionsAndComplete() {
```
rgomezp, line 51: add specific tests

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
```diff
@@ -0,0 +1,61 @@
...
+            offsets[key] = value
+            checkConditionsAndComplete()
+        }
+    }
+
+    /**
+     * Register a condition with its corresponding deferred action. Returns a deferred condition.
+     */
+    override fun registerCondition(condition: ICondition): CompletableDeferred<Long?> {
+        val deferred = CompletableDeferred<Long?>()
```
rgomezp, line 44: wrap with a mutex lock

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
```diff
@@ -34,6 +35,9 @@ import com.onesignal.user.internal.subscriptions.impl.SubscriptionManager
 
 internal class UserModule : IModule {
     override fun register(builder: ServiceBuilder) {
+        // Consistency
+        builder.register<IConsistencyManager>().provides<IConsistencyManager>()
```
rgomezp, line 39: register concrete implementation and provides the interface

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/UpdateUserOperationExecutor.kt
```diff
@@ -131,6 +135,11 @@ internal class UpdateUserOperationExecutor(
                     deltasObject,
                 )
 
+                // deltas includes property that may impact IAM fetch
+                if (deltasObject.hasAtLeastOnePropertySet) {
```
jkasten2, line 139: We want to use the new offset for any update, not just fields that support deltas. Example tags are not under delts but we want to wait for those.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -208,7 +214,7 @@ internal class InAppMessagesManager(
         }
 
         suspendifyOnThread {
-            fetchMessages()
+            fetchMessages(null)
```
jkasten2, line 217: We should pass in the offset here. It's a subscription update that we should be waiting on that will probably have a new offset value.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -182,7 +188,7 @@ internal class InAppMessagesManager(
...
```
jkasten2, line 191: This could result in skipping an offset we should be waiting on, we should pass the `offset` value in here.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -191,7 +197,7 @@ internal class InAppMessagesManager(
         tag: String,
     ) {
         suspendifyOnThread {
-            fetchMessages()
+            fetchMessages(null)
```
jkasten2, line 200: This could result in skipping an offset we should be waiting on, we should pass the `offset` value in here.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +201,59 @@ internal class InAppBackendService(
...
+        var attempts = 1
+
+        while (true) {
+            val urlWithOffset = "$baseUrl?offset=$offset"
+            val response = _httpClient.get(urlWithOffset)
+
+            if (response.isSuccess) {
+                val jsonResponse = response.payload?.let { JSONObject(it) }
+                return jsonResponse?.let { hydrateInAppMessages(it) }
+            } else if (response.statusCode == 429) { // 429 Too Many Requests
```
jkasten2, line 215: I believe the spec defined 425 "Too Early". However we should handle both 425 and 429

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,82 @@ internal class InAppBackendService(
...
+            }
+            headerValues.offset = offset
+            headerValues.secondsSinceAppOpen = sessionTime
+            val response = _httpClient.get(baseUrl, headerValues)
+
+            if (response.isSuccess) {
+                val jsonResponse = response.payload?.let { JSONObject(it) }
+                return jsonResponse?.let { hydrateInAppMessages(it) }
+            } else if (response.statusCode == 425) { // 425 Too Early
+                val retryLimit = response.retryLimit ?: 3
```
jkasten2, line 223: A retry value should not be inline. Either make a retry function that does this logic and/or make a safe default constant.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,82 @@ internal class InAppBackendService(
...
+
+                if (attempts > retryLimit) {
+                    break
+                }
+
+                delay(retryAfter * 1000L) // Convert seconds to milliseconds
+            } else if (response.statusCode == 429) { // 429 Too Many Requests
+                val retryAfter = response.retryAfterSeconds ?: delayTime
+
+                if (attempts > (response.retryLimit ?: 3)) {
```
jkasten2, line 234: A retry value should not be inline. Either make a retry function that does this logic and/or make a safe default constant.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,82 @@ internal class InAppBackendService(
...
+            headerValues.offset = offset
+            headerValues.secondsSinceAppOpen = sessionTime
+            val response = _httpClient.get(baseUrl, headerValues)
+
+            if (response.isSuccess) {
+                val jsonResponse = response.payload?.let { JSONObject(it) }
+                return jsonResponse?.let { hydrateInAppMessages(it) }
+            } else if (response.statusCode == 425) { // 425 Too Early
+                val retryLimit = response.retryLimit ?: 3
+                val retryAfter = response.retryAfterSeconds ?: 1
```
jkasten2, line 224: A retry value should not be inline. Either make a retry function that does this logic and/or make a safe default constant.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,82 @@ internal class InAppBackendService(
...
+
+                delay(retryAfter * 1000L) // Convert seconds to milliseconds
+            } else if (response.statusCode == 429) { // 429 Too Many Requests
+                val retryAfter = response.retryAfterSeconds ?: delayTime
+
+                if (attempts > (response.retryLimit ?: 3)) {
+                    break
+                }
+
+                delay(retryAfter * 1000L) // Convert seconds to milliseconds
```
jkasten2, line 238: Use the format `1_000L`

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,82 @@ internal class InAppBackendService(
...
+                return jsonResponse?.let { hydrateInAppMessages(it) }
+            } else if (response.statusCode == 425) { // 425 Too Early
+                val retryLimit = response.retryLimit ?: 3
+                val retryAfter = response.retryAfterSeconds ?: 1
+
+                if (attempts > retryLimit) {
+                    break
+                }
+
+                delay(retryAfter * 1000L) // Convert seconds to milliseconds
```
jkasten2, line 230: Use the format `1_000L`

jkasten2: Drop the comment, this is very much implied.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
jkasten2: Drop the comment, this is very much implied.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/OptionalHeaderValues.kt
```diff
@@ -0,0 +1,8 @@
+package com.onesignal.core.internal.http.impl
+
+data class OptionalHeaderValues(
```
jkasten2, line 3: Could we add a comment to each field of what they do?

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
```diff
@@ -39,6 +41,7 @@ internal class SubscriptionOperationExecutor(
     private val _configModelStore: ConfigModelStore,
     private val _buildUserService: IRebuildUserService,
     private val _newRecordState: NewRecordsState,
+    private val _iamFetchConsistencyManager: IConsistencyManager<IamFetchOffsetKey>,
```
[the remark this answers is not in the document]
rgomezp: We're registering the consistency manager in the [user module](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168/files#diff-c2aa9f3964fe4a5f55d6cccc6769bcd58c5a6dd2a57b829141805b57eeb1cb2fR41), which is in core.

Since the consistency manager base class and related code is already in core, I figured might as well put all consistency related code together (e.g. enums & conditions).

The IAM manager reads the offset while the executors write the offset. Do you have any suggestions regarding how to not have both modules share the dependency?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/impl/ConsistencyManager.kt
```diff
@@ -0,0 +1,72 @@
...
+            conditions.add(pair)
+            checkConditionsAndComplete()
+            return deferred
+        }
+    }
+
+    private fun checkConditionsAndComplete() {
+        val completedConditions = mutableListOf<Pair<ICondition<K>, CompletableDeferred<Long?>>>()
+
+        for ((condition, deferred) in conditions) {
```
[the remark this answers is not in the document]
rgomezp: This is a private method who's calling methods are already protected. I believe that wrapping it here will cause it to never resolve.

rgomezp:
Fixups & cleanup was getting a bit hairy so I re-organized the git tree.

I made a backup of the branch and put all the feedback changes into a [single commit](https://github.com/OneSignal/OneSignal-Android-SDK/commit/c64346b93e7353dbe52d915b7fe36b314730ec9c) (ignore whitespace).

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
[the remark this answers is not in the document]
jkasten2: hmm, ya I can see our reflection implementation not liking the generic.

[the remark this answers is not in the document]
jkasten2: The main thing I is that would like to not add is references to other modules in core. I think `IamFetchOffsetKey` might just be to specific of a name, as it could be used in the future for other things. Maybe rename this to `FetchOffsetKey` or `UserFetchOffsetKey`?

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +201,88 @@ internal class InAppBackendService(
...
+    }
+
+    private fun hydrateInAppMessages(jsonResponse: JSONObject): List<InAppMessage>? =
+        if (jsonResponse.has("in_app_messages")) {
+            val iamMessagesAsJSON = jsonResponse.getJSONArray("in_app_messages")
+            // TODO: Outstanding question on whether we still want to cache this.  Only used when
+            //       hard start of the app, but within 30 seconds of it being killed (i.e. same session startup).
+            // Cache copy for quick cold starts
+            // _prefs.savedIAMs = iamMessagesAsJSON.toString()
+            _hydrator.hydrateIAMMessages(iamMessagesAsJSON)
```
jkasten2, line 284: I don't think we want to cache this any more, as it could lead to the wrong IAMs shown.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +201,88 @@ internal class InAppBackendService(
...
+            } else if (response.statusCode == 429) { // 429 Too Many Requests
+                val retryAfter = response.retryAfterSeconds ?: delayTime
+
+                if (attempts > (response.retryLimit ?: DEFAULT_RETRY_LIMIT)) {
+                    break
+                }
+
+                delay(retryAfter * 1_000L)
+                delayTime *= 2 // Exponential backoff
+            } else if (response.statusCode == 444 || response.statusCode >= 500) {
```
jkasten2, line 244: Should be `404` instead of `444`

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,82 @@ internal class InAppBackendService(
...
+
+    private suspend fun attemptFetchWithRetries(
+        baseUrl: String,
+        offset: Long,
+        sessionTime: Long,
+    ): List<InAppMessage>? {
+        var attempts = 1
+        var delayTime = 1 // Start with a 1-second delay for exponential backoff
+
+        while (true) {
```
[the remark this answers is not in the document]
jkasten2: Why not use `attempts > (response.retryLimit ?: DEFAULT_RETRY_LIMIT)` here so we don't have to have this logic in each of the if conditions?

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/IInAppBackendService.kt
```diff
@@ -13,12 +13,15 @@ internal interface IInAppBackendService {
      *
      * @param appId The ID of the application that the IAM will be retrieved from.
      * @param subscriptionId The specific subscription within the [appId] the IAM will be delivered to.
+     * @param offset OPTIONAL - used for read your write consistency
      *
```
jkasten2, line 17: Missing comment for newly added sessionDurationProvider

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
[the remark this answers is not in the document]
jkasten2: @rgomezp I think I forgot to unresolve this on my 2nd reply if you missed this.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +201,93 @@ internal class InAppBackendService(
...
+            val response = _httpClient.get(baseUrl, values)
+
+            if (response.isSuccess) {
+                val jsonResponse = response.payload?.let { JSONObject(it) }
+                return jsonResponse?.let { hydrateInAppMessages(it) }
+            } else if (response.statusCode == 425) { // 425 Too Early
+                val retryLimit = response.retryLimit ?: DEFAULT_RETRY_LIMIT
+                val retryAfter = response.retryAfterSeconds ?: DEFAULT_RETRY_AFTER_SECONDS
+
+                if (attempts == retryLimit) {
```
jkasten2, line 230: Since `retryLimit` is dynamic we should really check greater than or equal to.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +201,93 @@ internal class InAppBackendService(
...
+                    break
+                }
+
+                delay(retryAfter * 1_000L)
+            } else if (response.statusCode == 429) { // 429 Too Many Requests
+                val retryAfter = response.retryAfterSeconds ?: delayTime
+
+                if (attempts == (response.retryLimit ?: DEFAULT_RETRY_LIMIT)) {
```
jkasten2, line 238: Since `retryLimit` is dynamic we should really check greater than or equal to.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +201,93 @@ internal class InAppBackendService(
...
+                    break
+                }
+
+                delay(retryAfter * 1_000L)
+                delayTime *= 2 // Exponential backoff
+            } else if (response.statusCode >= 500) {
+                if (attempts == DEFAULT_RETRY_LIMIT) {
```
jkasten2, line 245: Since retrying is dynamic we should really check greater than or equal to.

jkasten2 (changes requested):

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/IamFetchReadyCondition.kt
```diff
@@ -0,0 +1,28 @@
...
+ *  id : String - the index of the RYW token map
+ */
+class IamFetchReadyCondition(
+    private val id: String,
+) : ICondition {
+    override fun isMet(indexedTokens: Map<String, Map<IConsistencyKeyEnum, String?>>): Boolean {
+        val tokenMap = indexedTokens[id] ?: return false
+        val userUpdateTokenSet = tokenMap[IamFetchRywTokenKey.USER_UPDATE] != null
+        val subscriptionUpdateTokenSet = tokenMap[IamFetchRywTokenKey.SUBSCRIPTION_UPDATE] != null
+        return (userUpdateTokenSet && subscriptionUpdateTokenSet) || userUpdateTokenSet
```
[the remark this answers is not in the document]
rgomezp: Updated

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/enums/IamFetchRywTokenKey.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.onesignal.common.consistency.models.IConsistencyKeyEnum
+
+/**
+ * Each enum is a key that we use to keep track of read-your-write tokens.
+ * Although the enums are named with "UPDATE", they serve as keys for tokens from both PATCH & POST
+ */
+enum class IamFetchRywTokenKey : IConsistencyKeyEnum {
+    USER_UPDATE,
+    SUBSCRIPTION_UPDATE,
```
Koji23, line 11: do we not need the create endpoints in here?

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/IamFetchReadyCondition.kt
```diff
@@ -0,0 +1,52 @@
...
+        val subscriptionUpdateTokenSet = tokenMap[IamFetchRywTokenKey.SUBSCRIPTION_UPDATE] != null
+
+        /**
+         * If the RYW token was not included in one of the 4 endpoints (PATCH & POST * user,
+         * subscription), we should resolve the condition so as to not block IAM fetch unnecessarily
+         * This can happen if there is a regression where we stop including the tokens in the res
+         * body or we turn off the feature temporarily
+         */
+        val subscriptionOrUserTokenWasNotSet =
+            tokenMap[IamFetchRywTokenKey.SUBSCRIPTION_UPDATE] == ConsistencyManager.RYW_TOKEN_NOT_PROVIDED ||
```
Koji23, line 29: if the value of `RYW_TOKEN_NOT_PROVIDED` is "0" but actual ryw-tokens are always 20 characters long won't this never be true?

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -1,17 +1,23 @@
...
 import com.onesignal.inAppMessages.internal.InAppMessage
 import com.onesignal.inAppMessages.internal.InAppMessageContent
 import com.onesignal.inAppMessages.internal.backend.GetIAMDataResponse
 import com.onesignal.inAppMessages.internal.backend.IInAppBackendService
 import com.onesignal.inAppMessages.internal.hydrators.InAppHydrator
+import kotlinx.coroutines.delay
 import org.json.JSONObject
 
+private const val DEFAULT_RETRY_LIMIT = 3
+private const val DEFAULT_RETRY_AFTER_SECONDS = 1
```
Koji23, line 19: I don't think we should set default retry parameters on the SDK. If turbine doesn't provide these things then that should communicate to the SDK not to retry at all

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +202,84 @@ internal class InAppBackendService(
...
+                    }
+                    val retryAfter = response.retryAfterSeconds ?: DEFAULT_RETRY_AFTER_SECONDS
+                    delay(retryAfter * 1_000L)
+                } else if (response.statusCode == 429) { // 429 Too Many Requests
+                    val retryAfter = response.retryAfterSeconds ?: delayTime
+
+                    delay(retryAfter * 1_000L)
+                    delayTime *= 2 // Exponential backoff
+                } else if (response.statusCode >= 500) {
+                    delay(delayTime * 1_000L)
```
Koji23, line 243: should we even need a delay for a 5xx error? I'd imagine we would not want to retry those. Also maybe add a `&& response.statusCode < 600` condition too

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +202,84 @@ internal class InAppBackendService(
...
+                    if (response.retryLimit != null) {
+                        retryLimit = response.retryLimit!!
+                    }
+                    val retryAfter = response.retryAfterSeconds ?: DEFAULT_RETRY_AFTER_SECONDS
+                    delay(retryAfter * 1_000L)
+                } else if (response.statusCode == 429) { // 429 Too Many Requests
+                    val retryAfter = response.retryAfterSeconds ?: delayTime
+
+                    delay(retryAfter * 1_000L)
+                    delayTime *= 2 // Exponential backoff
```
Koji23, line 241: I'm not sure if we want to implement the exponential backoff in the SDK. I feel like if we need it then we should add that in turbine. I think it would be more predictable if the SDK just adheres to the Retry-After header instead of overriding it with it's own exponential backoff

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/enums/IamFetchRywTokenKey.kt
rgomezp: [see comment](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168/files#diff-337b3906ee241eaf210ae8927ebb0767b98e8dc704ba9e2b549e81dcd494f174R7)

Koji23: ah ok. In that case would it make sense to just remove the `_UPDATE` suffix then?

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/IamFetchReadyCondition.kt
rgomezp: It's a magic number. If the value is null when set, [we set it to this constant](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168/files#diff-ebc9431e7f9803eb553d3c75bbbbc540596a279fe76e419255b4fdcf0ade749aR48). We then use it to [exclude the value from the IAM fetch](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168/files#diff-3f270602e75b4896d8f56cfac018353aa0b7f85e72e5c7ea25423202d7ca9ed9R217).

The alternative is to introduce a new ConsistencyManager method to force-resolve some or all conditions (i.e. blocking IAM fetch)

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
```diff
@@ -34,6 +36,9 @@ import com.onesignal.user.internal.subscriptions.impl.SubscriptionManager
 
 internal class UserModule : IModule {
     override fun register(builder: ServiceBuilder) {
+        // Consistency
```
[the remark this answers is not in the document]
rgomezp: It's just meant to section the dependencies [like below](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168/files/46d528d34cc29cfad13eeda8c6b080f59d96fdc9#diff-c2aa9f3964fe4a5f55d6cccc6769bcd58c5a6dd2a57b829141805b57eeb1cb2fR42)

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/UpdateUserOperationExecutor.kt
```diff
@@ -94,9 +98,15 @@ internal class UpdateUserOperationExecutor(
 
                     // The session time we pass up is the total session time across all `TrackSessionEndOperation`
                     // operations that exist in this group.
-                    val sessionTime = if (deltasObject.sessionTime != null) deltasObject.sessionTime!! + operation.sessionTime else operation.sessionTime
+                    val sessionTime =
+                        if (deltasObject.sessionTime != null) {
+                            deltasObject.sessionTime!! + operation.sessionTime
+                        } else {
+                            operation.sessionTime
```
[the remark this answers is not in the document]
rgomezp: I'm not entirely sure. This looks like a simple lint change though.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
rgomezp: Done

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
rgomezp: Done.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
rgomezp: Done.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/IamFetchReadyCondition.kt
rgomezp: Update: I re-architected it to avoid the magic number. We now call [`resolveConditionsWithId` to unblock the execution](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168/commits/51aaa8b139ad6a9048fb99f9890f44b094f62792)

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
```diff
@@ -216,6 +225,7 @@ internal class SubscriptionOperationExecutor(
         return ExecutionResponse(ExecutionResult.SUCCESS)
     }
 
+    // TODO: whenever the end-user changes users, we need to add the read-your-write token here, currently no code to handle the re-fetch IAMs
```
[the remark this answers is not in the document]
rgomezp: It's a different endpoint. We call
```
val response = _httpClient.patch("apps/$appId/subscriptions/$subscriptionId/owner", requestJSON)
```

Koji23 (commented):
this LGTM from a non-iOS expert view

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        /**
+         * We always update the session count so we know we will have a userUpdateToken. We don't
+         * necessarily make a subscriptionUpdate call on every session. The following logic
+         * is written in a way so that if somehow the subscriptionUpdateToken is set *before* the
+         * userUpdateToken, we will wait for the userUpdateToken to also be set. This is because
+         * we know that a userUpdate call was made and both user & subscription properties are
+         * considered during segment calculations.
+         */
+        return (userUpdateTokenSet && subscriptionUpdateTokenSet) || userUpdateTokenSet
```
jkasten2, line 36: Based on the comment here and logic it seems we don't actually care about `subscriptionUpdateTokenSet` at all. If this line where `return userUpdateTokenSet` it would result in the same returned value.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,75 @@ internal class InAppBackendService(
...
         Logging.error("Encountered a $statusCode error while attempting in-app message $requestType request: $response")
     }
+
+    private suspend fun attemptFetchWithRetries(
+        baseUrl: String,
+        rywToken: String?,
+        sessionDurationProvider: () -> Long,
+    ): List<InAppMessage>? {
+        var attempts = 1
+        var retryLimit: Int? = null // Retry limit will be determined dynamically
```
jkasten2, line 208: We really don't want a default? If so let's change this comment to say the retry limit is remote defined, via the IAM fetch response.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,75 @@ internal class InAppBackendService(
     ) {
         Logging.error("Encountered a $statusCode error while attempting in-app message $requestType request: $response")
     }
+
+    private suspend fun attemptFetchWithRetries(
+        baseUrl: String,
+        rywToken: String?,
+        sessionDurationProvider: () -> Long,
+    ): List<InAppMessage>? {
+        var attempts = 1
```
jkasten2, line 207: Shouldn't this be `0`?, as we haven't made an attempt yet.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,4 +198,75 @@ internal class InAppBackendService(
...
+        var retryLimit: Int? = null // Retry limit will be determined dynamically
+
+        while (retryLimit == null || attempts <= retryLimit + 1) {
```
jkasten2, line 210: We can make this loop condition a bit simpler if we make this a `do { } while` loop instead. That will let us drop the `retryLimit == null ||` part and will also indicate to the reader we will do this logic at least once.

jkasten2 (changes requested):
Left a few more comments in the code.

Also a few more things not in the code:
* Manual testing still says "// TO DO"
* Some other things not checked off in the PR check list.
* Does this code work with that is available on the backend today? If not we will need to either wait until the backend is shipped until we merge this PR into main, or make changes to this PR to support that.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
rgomezp: Correct. Do you see value in leaving it this way though so as to not lose that context?

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
rgomezp: @Koji23 suggested we simply don't retry at all if it's not provided remotely
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2168#discussion_r1773848134

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/consistency/IamFetchReadyCondition.kt
jkasten2: I think a comment about `tokenMap[IamFetchRywTokenKey.SUBSCRIPTION]` is fine to provide what is considered in this `isMet` logic, but it has no impact on the logic in the code there shouldn't be code using it.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
rgomezp: Done.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
rgomezp: Done.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
jkasten2: I see, then `0` should be the default instead of `null`, as the default would be to not retry then.

There is also currently a bug in the code were if the backend doesn't return a value for retries it will result in an infinite loop. Using `0` then instead of `null` will fix this. There are others ways to prevent the infinite loop bug but setting this to `0` instead of `null` also prevent this mistake from happening in the future.

rgomezp: Fixed

jkasten2 (approved):

PR #2182, opened 2024-08-23
Refactor HTTP client to support passing header values + add support for `Retry-Limit`

rgomezp:
# Description
## One Line Summary
Refactor HTTP client to support passing header values + add support for `Retry-Limit`

### Motivation
Make the HTTP client more extensible by encapsulating header values in a single object we pass to the HTTP call. Otherwise we would quickly end up with lots of arguments not needed in all use-cases.

Add support for `Retry-Limit` by adding it to the `HttpResponse` object used higher in the code path. 

### Scope
Changes to the `HttpClient`

# Testing
## Unit testing
Existing unit tests should provide coverage.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2182)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/HttpClient.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/HttpClient.kt
@@ -188,6 +188,7 @@ internal class HttpClient(
                     httpResponse = con.responseCode
 
                     val retryAfter = retryAfterFromResponse(con)
+                    val retryLimit = retryLimitFromResponse(con)
                     val newDelayUntil = _time.currentTimeMillis + (retryAfter ?: 0) * 1_000
                     if (newDelayUntil > delayNewRequestsUntil) delayNewRequestsUntil = newDelayUntil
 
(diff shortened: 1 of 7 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/OptionalHeaderValues.kt
```diff
@@ -0,0 +1,5 @@
+package com.onesignal.core.internal.http.impl
+
+data class OptionalHeaderValues(
+    var cacheKey: String? = null,
```
jkasten2, line 4: It doesn't look like we ever assign this value (or would ever want too). Can we make this immutable by changing this `var` into `val`? You can then use the constructor to set the value in one of two ways:
1. ` OptionalHeaderValues("valueOfCacheKey")`
2. ` OptionalHeaderValues(cacheKey = "valueOfCacheKey")`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/HttpClient.kt
```diff
@@ -110,7 +106,7 @@ internal class HttpClient(
         method: String?,
         jsonBody: JSONObject?,
         timeout: Int,
-        cacheKey: String?,
+        headerValues: OptionalHeaderValues?,
```
jkasten2, line 109: `headerValues`, like noted below with the class name, the wording "value" doesn't add anything.

jkasten2 (changes requested):

jkasten2 (approved):

PR #2185, opened 2024-08-27
Fix ending an already ended session

jkasten2:
# Description
## One Line Summary
Fixes invalid REST API requests, when location is on, to both outcomes/measure and User updates with `session_time` being zero/missing in the background.

## Details
This is only present when location sharing is enabled in the SDK (in addition to the app and end-user allowing it) and the end-user opens the app at any point after receiving a notification. This resulted in 400 errors in the outcomes/measure case and no-op User updates that wasting resources, the later doesn't depend on receiving notifications. The outcomes/measure is also additive as the bad requests are cached and retried on app open.

### Motivation
These invalid and no op request put a lot of extra load on both the device and the OneSignal backend.

### Scope
Session tracking only.

# Testing
## Unit testing
Added a new session test and update an existing one.

## Manual testing
Tested on an Android 14 emulator with our example app.

Reproducing the issue:
1. Allowed location
3. Sent a notification to the device
4. Wait until background location update fired.
   * Modified SDK to a lower time so this takes far less than 10 minutes
5. Observed the invalid request of `os__session_duration` with no session_time sent and the 400 response.

Reproducing repeating bad cached values:
1. Reproduce the root issue noted above.
2. Cold restart the app
3. Observer the missing `session_time ` being relayed.
4. Added migration logic and observed these requests are no longer replied.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [X] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2185)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/background/IBackgroundService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/background/IBackgroundService.kt
@@ -4,9 +4,9 @@ import androidx.annotation.WorkerThread
 
 /**
  * Implement and provide this interface as part of service registration to indicate the service
- * wants to be instantiated and its [backgroundRun] function called when the app is in the background.  The
- * background process is initiated when the application is no longer in focus.  Each background
- * service's [scheduleBackgroundRunIn] will be analyzed to determine when [backgroundRun] should be called.
+ * wants to be instantiated and its [backgroundRun] function called when the app is in the background.
+ * Each background service's [scheduleBackgroundRunIn] will be analyzed to determine when
+ * [backgroundRun] should be called.
  */
 interface IBackgroundService {
     /**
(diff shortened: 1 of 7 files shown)


jinliu9508 (commented):
Code looks good and the changed comments are proper. I have also played with the test unit to verify that the results are correct.

I just have one question that since isValid is false by default and will set to true after the endSession call in onFocus(), is it expected that the first endSession() call after the very first app focus will always return early and do nothing?

jinliu9508:
> Code looks good and the changed comments are proper. I have also played with the test unit to verify that the results are correct.
> 
> I just have one question that since isValid is false by default and will set to true after the endSession call in onFocus(), is it expected that the first endSession() call after the very first app focus will always return early and do nothing?

Please disregard the question here as it is now clear to me that the change is necessary to start a new session if app is cold started.

PR #2200, opened 2024-10-04
Fix setting consentGiven throwing if called before initWithContext

jkasten2:
# Description
## One Line Summary
Prevent throwing if `consentGiven` is set before calling `initWithContext`.

## Details
Since `consentGiven` accesses `operationRepo` which requires `ApplicationService.appContext `to be non-null we stopped using the lazy-loading pattern for `operationRepo`. We put the assignment into `initWithContext` to make this more clear.

### Motivation
While most SDK methods throw if accessed before init the consent methods were an exception from the beginning. We need to restore the orignal behavior so apps do not break unexpectedly.

### Scope
Only affects OneSignalImp access to `operationRepo`, which should only have an effect on  `consentGiven` and `consentRequired`.

# Testing
## Unit testing
Added new tests for `consentRequired` & `consentGiven`.

## Manual testing
1. Called `OneSignal.setConsentGiven(true)` before `OneSignal.initWithContext`, observed crash.
2. Made fix
3. Ensured crash no longer happens.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2200)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -126,8 +126,9 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         }
 
     // Services required by this class
-    private val operationRepo: IOperationRepo
-        get() = services.getService()
+    // WARNING: OperationRepo depends on OperationModelStore which in-turn depends
+    // on ApplicationService.appContext being non-null.
+    private var operationRepo: IOperationRepo? = null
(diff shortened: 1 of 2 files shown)


jinliu9508 (approved):
LGTM!

PR #2208, opened 2024-10-30
Window manager BadTokenException / WindowLeaked

jinliu9508:
# Description
## One Line Summary
Fix the BadTokenException and WindowLeaked exception caused by showing a dialog on a finishing or destroyed activity.

### Motivation
The exception exists since User Model and now becomes the most occurring exception according to the data from Google Play SDK Console. This exception is relatively less impactful to user experience as it normal happens during app closing or backgrounding phrase, but we want to reduce the amount of reported exception as it may potentially affect app's stability rating over the Play Store. 

### Scope
The call to show a permission dialog has no effect if the activity is finishing, backgrounded, or destroyed, and it will not be retried unless user is explicitly asking to do so. There is no behavior change because the current version would just crash and will not retrying showing the dialog upon app restart. 

### OPTIONAL - Other
This PR contains 3 commits: the first commit simulates calling a permission dialog on a finished activity; the second commit shows that the fix is effective in removing the exception in such activity; the third commit cleans up the simulation code and leave the fix code change only. 
Related issue: https://github.com/OneSignal/OneSignal-Android-SDK/issues/2014

# Testing
## Manual testing
By checking out the first commit, I am able to reproduce a WindowLeaked exception by running the sample app. 
image
After the second commit, the exception is no longer reproducible. 
BadTokenException can be caused by showing the dialog in a "finishing" activity but I am unable to simulate the situation other than a "finished" activity. Since this fix checks for both states of the activity, I believe this can fixes both BadTokenException and WindowLeaked exceptions. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2208)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/AlertDialogPrepromptForAndroidSettings.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/AlertDialogPrepromptForAndroidSettings.kt
@@ -30,6 +30,8 @@ package com.onesignal.core.internal.permissions
 import android.app.Activity
 import android.app.AlertDialog
 import com.onesignal.core.R
+import com.onesignal.debug.LogLevel
+import com.onesignal.debug.internal.logging.Logging
 
 /**
  * A singleton helper which will display the fallback-to-settings alert dialog.
@@ -56,18 +58,26 @@ object AlertDialogPrepromptForAndroidSettings {
         // simulate showing a dialog in a finishing activity
         activity.finish()
 
-        AlertDialog.Builder(activity)
-            .setTitle(title)
-            .setMessage(message)
-            .setPositiveButton(R.string.permission_not_available_open_settings_option) { dialog, which ->
-                callback.onAccept()
-            }
-            .setNegativeButton(android.R.string.no) { dialog, which ->
-                callback.onDecline()
-            }
-            .setOnCancelListener {
-                callback.onDecline()
-            }
-            .show()
+        // ensure the activity that will be showing the dialog is available
+        if (activity == null || activity.isDestroyed || activity.isFinishing) {
+            Logging.log(LogLevel.ERROR, "Alert dialog for Android settings was skipped because the activity was unavailable to display it.")
+            return
+        }
+
+        if (activity != null && !activity.isFinishing) {
+            AlertDialog.Builder(activity)
+                .setTitle(title)
+                .setMessage(message)
+                .setPositiveButton(R.string.permission_not_available_open_settings_option) { dialog, which ->
+                    callback.onAccept()
+                }
+                .setNegativeButton(android.R.string.no) { dialog, which ->
+                    callback.onDecline()
+                }
+                .setOnCancelListener {
+                    callback.onDecline()
+                }
+                .show()
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/AlertDialogPrepromptForAndroidSettings.kt
```diff
@@ -53,18 +55,26 @@ object AlertDialogPrepromptForAndroidSettings {
...
-                callback.onDecline()
-            }
-            .setOnCancelListener {
-                callback.onDecline()
-            }
-            .show()
+        // ensure the activity that will be showing the dialog is available
+        if (activity == null || activity.isDestroyed || activity.isFinishing) {
+            Logging.log(LogLevel.ERROR, "Alert dialog for Android settings was skipped because the activity was unavailable to display it.")
+            return
```
jkasten2, line 61: This will prevent the crash but it will create some side effects when this case is hit:
1. Dropped permission requests - Permission prompt won't show when the app is resumed
2. Callback never fires - Customers code may hang for ever thinking the the prompt is still showing.

Here some quick thoughts on soe options to address these:
1. Wait - We can suspend the code here, until an activity is available.
2. Fire onDecline - This is a bit missleading but does solve the problem 2 note above.
* We could add a 3rd callback, something like onSkiped(), but I think that will make things more complex for customer than needed, so really don't like this alt B option here.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/AlertDialogPrepromptForAndroidSettings.kt
```diff
@@ -53,18 +55,26 @@ object AlertDialogPrepromptForAndroidSettings {
...
+        }
+
+        if (activity != null && !activity.isFinishing) {
```
jkasten2, line 64: Redundant if check with above (but missing the `isDestroyed` check), we should remove this.

jkasten2 (changes requested):
While this might be a 1 in million, and may never happen, I think it's possible the activity state to change between when we check and when `show()` is called., but probably only if we are trying to display from a background thread. The thread note is only a theory, but all Android APIs only do work on the main thread from what I have seen..

Do you have any thoughts on maybe doing a try catching on calling `show` instead of checking before calling `show`? We could specifically catch `android.view.WindowLeaked`, and then do our error handling logic there.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/AlertDialogPrepromptForAndroidSettings.kt
jinliu9508: Good point, I have added a code to fire the onDecline callback because it is consistent with the behavior when the dialog is cancelled or dismissed without a response.

jkasten2 (approved):

PR #2212, opened 2024-11-06
[Fix] NPE by getScheduleBackgroundRunIn

jinliu9508:
# Description
## One Line Summary
Fixes a NullPointerException issue thrown by SessionService.getScheduleBackgroundRunIn.

### Motivation
With the optimization introduced to `initWithContext`, part of the initialization is now performed in the background after `initWithContext` is called. This change created a rare race condition where services are accessed shortly after startup, but initialization hasn’t completed. In one scenario, when the app goes into the background before SessionService finishes initializing, a lifecycle callback fires to notify the "unfocus" event, attempting to access `SessionService`, whose component is null at that time. This error can crash the app and disrupt background services. This PR prevents the crash by ensuring that all major components of `SessionService` are available during the `initWithContext` phase.

### Scope
This PR moves part of `SessionService` initialization from `start()` (which was previously running in the background after `initWithContext`) to `bootstrap()`, which is called and completed during the `initWithContext` phase.

## Manual testing
I used a manual approach to reproduce the crash and verify the fix is working for the crash. 

Step to reproduce: 
1. Pull the first commit and run the demo app. During initialization, swipe the app into the background. (Note: This commit includes an arbitrary 10-second delay to the session service start time for testing purposes.)
2. Wait a few seconds and observe the crash in the background.
image

After the fix:
Following the same reproduction steps, the error no longer appears, and the session service functions as expected.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [x] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2212)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/SessionModule.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/SessionModule.kt
@@ -3,6 +3,7 @@ package com.onesignal.session
 import com.onesignal.common.modules.IModule
 import com.onesignal.common.services.ServiceBuilder
 import com.onesignal.core.internal.background.IBackgroundService
+import com.onesignal.core.internal.startup.IBootstrapService
 import com.onesignal.core.internal.startup.IStartableService
 import com.onesignal.session.internal.SessionManager
 import com.onesignal.session.internal.influence.IInfluenceManager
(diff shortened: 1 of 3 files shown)


jkasten2 (approved):

jinliu9508:
Squashing all commits into one before merge

PR #2231, opened 2024-12-10
WIP: Fix ANR caused by operationRepo.enqueue while loading is in progress

jinliu9508:
# Description
## One Line Summary
This PR will fix ANR caused by model store insertion while model store loading is stuck with some prolonged process.

### Motivation
Multiple users have reported that ANRs related to OneSignal have shown up in their analytics. We want to eliminate these ANRs to ensure smooth app start up. 

### Scope
This PR will alter the loading behavior for all models so that model insertion in the main thread becomes possible before loading is completed. The order will be correct because load() will insert cached items in the beginning of the list. However, it is possible for model to process the newly added item before loading is completed. 

### Others
This PR contains 4 commits. 
  * The 1st commit forces a prolonged loading process for OperationModelStore and will cause an ANR during app startup.
  * The 2nd commit adds a test unit that will fail because of the long loading time added by the first commit. 
  * The 3rd commit applies the fix and we should no longer observe the ANR and the test fail.
  * The 4th commit clears up the code that manually forces the loading process to take 5 seconds.

# Testing
## Unit testing
I have included a test unit to simulate a model insertion while loading is blocked by a lengthy process. THE UNIT TEST HAS ONE PENDING TODO.

## Manual testing
The first commit manually force the loading of operationRepo to take at least 5 seconds in order to simulate the issue. The second commit applies the fix and we can observe that ANR can no long occur. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2231)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ModelStore.kt
@@ -169,33 +169,34 @@ abstract class ModelStore<TModel>(
 
         val str = _prefs.getString(PreferenceStores.ONESIGNAL, PreferenceOneSignalKeys.MODEL_STORE_PREFIX + name, "[]")
         val jsonArray = JSONArray(str)
-        synchronized(models) {
-            val shouldRePersist = models.isNotEmpty()
-            for (index in jsonArray.length() - 1 downTo 0) {
-                val newModel = create(jsonArray.getJSONObject(index)) ?: continue
 
-                /*
-                 * NOTE: Migration fix for bug introduced in 5.1.12
-                 * The following check is intended for the operation model store.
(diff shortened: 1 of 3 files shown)


jinliu9508:
I am not sure about this change: Do we really want the behavioral change for all models so that insertion is possible even the loading is not completed?

jkasten2 (changes requested):
I think the change would introduce two issues:
1. A concurrent modification exception could be thrown
2. Operations could be added out of order

PR #2243, opened 2025-01-14
Fix: incorrect activity path for NotificationOpenedActivityHMS

jinliu9508:
# Description
## One Line Summary
Fix the class path of NotificationOpenedActivityHMS so it can be consistent with the path returned by the backend service.

### Motivation
NotificationOpenedActivityHMS is currently broken for all Huawei devices on V5 because the class path is currently under com.onesignal.notifications.activities instead of com.onesignal used in our backend service. This PR moves the class so the path is in consistent with that used by the backend service.

### Scope

# Testing
## Manual testing
  1. Place a breakpoint inside processOpen().
  2. Send a HMS notification and click it.
  3. Observe that the breakpoint is never reached.
  4. Apply the fix and repeat the steps above. Observe that the breakpoint has reached and the intent includes some custom data.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2243)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/AndroidManifest.xml
+++ b/OneSignalSDK/onesignal/notifications/src/main/AndroidManifest.xml
@@ -65,8 +65,9 @@
             </intent-filter>
         </service>
 
+        <!-- CAUTION: OneSignal backend includes the activity name in the payload, modifying the name without sync may result in notification click not firing -->
         <activity
-            android:name="com.onesignal.notifications.activities.NotificationOpenedActivityHMS"
+            android:name="com.onesignal.NotificationOpenedActivityHMS"
             android:noHistory="true"
             android:theme="@android:style/Theme.Translucent.NoTitleBar"
             android:exported="true">
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
@@ -24,12 +24,12 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package com.onesignal.notifications.activities
+package com.onesignal
+// OneSignal backend includes the activity name in the payload, modifying the namespace may result in notification click not firing
 
 import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
-import com.onesignal.OneSignal
 import com.onesignal.common.threading.suspendifyBlocking
 import com.onesignal.notifications.internal.open.INotificationOpenedProcessorHMS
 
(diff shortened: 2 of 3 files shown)


OneSignalSDK/onesignal/notifications/src/main/AndroidManifest.xml
```diff
@@ -66,7 +66,7 @@
         </service>
 
         <activity
-            android:name="com.onesignal.notifications.activities.NotificationOpenedActivityHMS"
+            android:name="com.onesignal.NotificationOpenedActivityHMS"
```
jkasten2, line 69: Let's add a comment here that we can't change the android:name, since the OneSignal backend incudes it in the payload.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
```diff
@@ -24,12 +24,11 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package com.onesignal.notifications.activities
+package com.onesignal
```
jkasten2, line 27: Let's also put a comment here and before the class definition too.

jkasten2 (changes requested):
Left 2 feedback items that we should put comments in the code.

Let's also add a test, the test should use `Class.forName("...")` so it will fail if someone ever tries to use a refactoring tool and doesn't read the comments either.

OneSignalSDK/onesignal/notifications/src/main/AndroidManifest.xml
jinliu9508: Good suggestion. I have added the comment for the change.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
jinliu9508: Good suggestion. I have added the comment for the change. Note that I added it below the first line as first line comment is not allowed in the coding standard.

jinliu9508:
> Left 2 feedback items that we should put comments in the code.
> 
> Let's also add a test, the test should use `Class.forName("...")` so it will fail if someone ever tries to use a refactoring tool and doesn't read the comments either.

Good point. I have added comments and a new test unit that specifically tests for the class path.

jkasten2 (approved):

PR #2244, opened 2025-01-14
V4 to v5 upgrade will migrate app ID for wrapper SDKs

nan-li:
# Description
## One Line Summary
(Resolves an issue that only affects wrappers) When migrating from v4 to v5, check for a cached app ID from v4 if none are available on v5.

### Motivation
* In wrappers, after an app upgrades from player to user model... if the app is not opened yet and a notification is received, there was no app ID detected and `initWithContext` returns early.
* Now, check for the cached app ID from v4 and use that to continue the init process. This mimics what should happen when using the native SDK directly.
* The reason this is an issue on wrappers is because when an app upgrades, the Application `onCreate` is automatically called, without needing to open the app. In native SDK clients, their OneSignal initialization code will run here and populate the app ID, setup the user, etc. However, in wrappers, clients don't modify the native code so the `onCreate` does not set app ID.

### Scope

- Using cached app ID from v4 if the app ID has not been set yet after migration
- Affects wrappers
- I aimed to mimic the same behavior as an upgrade using the native SDK, what would happen before the application is opened.

# Testing
## Manual testing

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2244)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/IPreferencesService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/IPreferencesService.kt
@@ -178,6 +178,11 @@ object PreferenceOneSignalKeys {
     // Legacy
 
     /**
+     * (String) The legacy app ID from SDKs prior to 5.
+     */
+    const val PREFS_LEGACY_APP_ID = "GT_APP_ID"
+
+    /**
      * (String) The legacy player ID from SDKs prior to 5.
      */
     const val PREFS_LEGACY_PLAYER_ID = "GT_PLAYER_ID"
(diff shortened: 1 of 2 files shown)


jkasten2 (approved):

jinliu9508 (approved):
LGTM!

Also verified that Player Model is using the same key to store a legacy app ID.

nan-li:
Looks like a maybe-flaky test added recently is failing the unit tests: https://github.com/OneSignal/OneSignal-Android-SDK/actions/runs/13064783338/job/36455552170?pr=2244
```
com.onesignal.core.internal.operations.OperationRepoTests > ensure loading in the background thread does not block enqueue FAILED
    io.kotest.assertions.AssertionFailedError: expected:<TERMINATED> but was:<BLOCKED>
        at app//com.onesignal.core.internal.operations.OperationRepoTests$1$2.invokeSuspend(OperationRepoTests.kt:135)
        at app//com.onesignal.core.internal.operations.OperationRepoTests$1$2.invoke(OperationRepoTests.kt)
        at app//com.onesignal.core.internal.operations.OperationRepoTests$1$2.invoke(OperationRepoTests.kt)
```

PR #2251, opened 2025-01-31
Fix: Anonymous Login request not cleared if app is forced close within 5 seconds on a new install

jinliu9508:
# Description
## One Line Summary
This PR fixes an issue where subscriptions do not update correctly and results in a 400 error when calling Login. It also handles this bad state when migrating from any previous version.

### Motivation
There is an edge case where the app successfully creates a user for the first time, but the request remains in the queue due to the `PostCreateDelay`. When this happens, all future Login calls return a 400 error, and any subscription updates fail. This PR prevents this scenario by correctly updating the operation queue and handling the bad state during migration.

### Scope
  * OperationRepo: No longer holds up operation processing before the `PostCreateDelay`. Instead, the waiting time for `PostCreateDelay` is now integrated into `delayBeforeNextExecution` at the end of execution.
  * LoginUserOperationExecutor: Now detects if an anonymous Login request is made without other operations. When this happens, it returns a `FAIL_NORETRY` and drops the duplicated Login request.

# Testing
## Manual testing
###Steps to reproduce
This bug can happen strictly after CreateUser is successful for the first time but the app is force-closed before the 5 seconds `PostCreateDelay` has reached. One way to reproduce this issue is:
  1. Fresh install the app.
  2. Right after the subscription is created or receiving a 200 code from the Create User request, quickly force close the app within 5 seconds.
  3. Re-open the app, observe that a 400 error code is received and any updates like subscribing is not updated in the dashboard.

###New Install
  Step 1-2 same as above
  3. Re-open the app, wait for the server response and observe that Login with the external ID is successful. Confirmed that no error is logged and the fix works.

###Update From the Broken Version
  Step 1-2 same as above
  3. upgrade to this version without uninstalling the previous version or clearing the storage, observe that Login with the externalID is successful without error.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2251)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -241,18 +241,6 @@ internal class OperationRepo(
                     queue.forEach { it.operation.translateIds(response.idTranslations) }
                 }
                 response.idTranslations.values.forEach { _newRecordState.add(it) }
-                // Stall processing the queue so the backend's DB has to time
-                // reflect the change before we do any other operations to it.
-                // NOTE: Future: We could run this logic in a
-                // coroutineScope.launch() block so other operations not
-                // effecting this these id's can still be done in parallel,
-                // however other parts of the system don't currently account
-                // for this so this is not safe to do.
-                val waitTime = _configModelStore.model.opRepoPostCreateDelay
-                delay(waitTime)
-                synchronized(queue) {
-                    if (queue.isNotEmpty()) waiter.wake(LoopWaiterMessage(false, waitTime))
-                }
             }
 
             var highestRetries = 0
@@ -316,7 +304,11 @@ internal class OperationRepo(
                 }
             }
 
+            // wait for retry and post create waiters to start next operation
             delayBeforeNextExecution(highestRetries, response.retryAfterSeconds)
+            if (response.idTranslations != null) {
+                delayForPostCreate(_configModelStore.model.opRepoPostCreateDelay)
+            }
         } catch (e: Throwable) {
             Logging.log(LogLevel.ERROR, "Error attempting to execute operation: $ops", e)
 
@@ -342,6 +334,22 @@ internal class OperationRepo(
         Logging.error("Operations being delay for: $delayFor ms")
         withTimeoutOrNull(delayFor) {
             retryWaiter.waitForWake()
+        }
+    }
+
+    /**
+     * Stall processing the queue so the backend's DB has to time
+     * reflect the change before we do any other operations to it.
+     * NOTE: Future: We could run this logic in a
+     * coroutineScope.launch() block so other operations not
+     * effecting this these id's can still be done in parallel,
+     * however other parts of the system don't currently account
+     * for this so this is not safe to do.
+     */
+    suspend fun delayForPostCreate(postCreateDelay: Long) {
+        delay(postCreateDelay)
+        synchronized(queue) {
+            if (queue.isNotEmpty()) waiter.wake(LoopWaiterMessage(false, postCreateDelay))
         }
     }
 
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
@@ -66,6 +66,13 @@ internal class LoginUserOperationExecutor(
         loginUserOp: LoginUserOperation,
         operations: List<Operation>,
     ): ExecutionResponse {
+        // Handle a bad state that can happen in User Model 5.1.27 or earlier versions that old Login
+        // request is not removed after processing if app is force-closed within the PostCreateDelay.
+        // Anonymous Login being processed alone will surely be rejected, so we need to drop the request
+        val containsSubscriptionOperation = operations.any { it is CreateSubscriptionOperation || it is TransferSubscriptionOperation }
+        if (!containsSubscriptionOperation && loginUserOp.externalId == null) {
+            return ExecutionResponse(ExecutionResult.FAIL_NORETRY)
+        }
         if (loginUserOp.existingOnesignalId == null || loginUserOp.externalId == null) {
             // When there is no existing user to attempt to associate with the externalId provided, we go right to
             // createUser.  If there is no externalId provided this is an insert, if there is this will be an
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/operations/LoginUserOperationExecutorTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/operations/LoginUserOperationExecutorTests.kt
@@ -39,6 +39,16 @@ class LoginUserOperationExecutorTests : FunSpec({
     val localSubscriptionId2 = "local-subscriptionId2"
     val remoteSubscriptionId1 = "remote-subscriptionId1"
     val remoteSubscriptionId2 = "remote-subscriptionId2"
+    val createSubscriptionOperation =
+        CreateSubscriptionOperation(
+            appId,
+            localOneSignalId,
+            "subscriptionId1",
+            SubscriptionType.PUSH,
+            true,
+            "pushToken1",
+            SubscriptionStatus.SUBSCRIBED,
+        )
 
     test("login anonymous user successfully creates user") {
         // Given
@@ -58,7 +68,7 @@ class LoginUserOperationExecutorTests : FunSpec({
         val loginUserOperationExecutor =
             LoginUserOperationExecutor(
                 mockIdentityOperationExecutor,
-                MockHelper.applicationService(),
+                AndroidMockHelper.applicationService(),
                 MockHelper.deviceService(),
                 mockUserBackendService,
                 mockIdentityModelStore,
@@ -67,7 +77,11 @@ class LoginUserOperationExecutorTests : FunSpec({
                 MockHelper.configModelStore(),
                 MockHelper.languageContext(),
             )
-        val operations = listOf<Operation>(LoginUserOperation(appId, localOneSignalId, null, null))
+        val operations =
+            listOf<Operation>(
+                LoginUserOperation(appId, localOneSignalId, null, null),
+                createSubscriptionOperation,
+            )
 
         // When
         val response = loginUserOperationExecutor.execute(operations)
@@ -98,7 +112,7 @@ class LoginUserOperationExecutorTests : FunSpec({
         val loginUserOperationExecutor =
             LoginUserOperationExecutor(
                 mockIdentityOperationExecutor,
-                MockHelper.applicationService(),
+                AndroidMockHelper.applicationService(),
                 MockHelper.deviceService(),
                 mockUserBackendService,
                 mockIdentityModelStore,
@@ -107,7 +121,11 @@ class LoginUserOperationExecutorTests : FunSpec({
                 MockHelper.configModelStore(),
                 MockHelper.languageContext(),
             )
-        val operations = listOf<Operation>(LoginUserOperation(appId, localOneSignalId, null, null))
+        val operations =
+            listOf<Operation>(
+                LoginUserOperation(appId, localOneSignalId, null, null),
+                createSubscriptionOperation,
+            )
 
         // When
         val response = loginUserOperationExecutor.execute(operations)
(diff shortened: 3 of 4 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
```diff
@@ -66,6 +66,13 @@ internal class LoginUserOperationExecutor(
         loginUserOp: LoginUserOperation,
         operations: List<Operation>,
     ): ExecutionResponse {
+        // Handle a bad state that can happen in User Model 5.1.27 or earlier versions that ID translations
+        // are lost if app is force-closed within the PostCreateDelay.
+        // Login operation being processed alone will surely be rejected, so we need to add an existing
+        // external ID to this legacy operation and have it picked up by SetAlias.
+        if (operations.isEmpty() && _identityModelStore.model.externalId != null) {
+            loginUserOp.externalId = _identityModelStore.model.externalId
```
jkasten2, line 74: Could there be an issue with properties being set on the wrong user?
Example:
1. Login(A)
2. setTag(A)
3. Login(B)
4. setTag(B)

Could this cause User B to get both tag A and B, when it was intended for User A to get tag A and User B to get Tag B?

jkasten2 (changes requested):
Let's add a test too

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
jinliu9508: I just tested this scenario and this is how the manual test is setup:
  1. Fresh install on main branch, right after the anonymous user is created but before the post create delay, programmatically call Login(A), setTag(A), Login(B), setTag(B) altogether and close the app.
  2. Switch to this branch and update the app. 
  3. Run the app and observe in the dashboard that both users are created successfully and each with the correct tag.

After analyzing the log closely, the code in this if statement will actually never hit for the scenario. OperationRepo will process the operations in this order:
  1. Create subscription
  2. Login anonymously
  3. Refresh user (anonymous)
  4. Login A AND transfer subscription
  5. Set Alias for A but got 409
  6. Refresh user (A)
  7. Set Tag for A
  8. Login B AND transfer subscription
  9. Set Tag for B

Since operations like setting tags does not go with the Login call in one request and the handler for 409 conflict will take care of the IDs, this will not cause the user to be assigned with the wrong tag.

Further more, I tested out that this bug only happen when Login is interrupted with no other operations like set tag within a few seconds after app starts up.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
```diff
@@ -66,6 +66,12 @@ internal class LoginUserOperationExecutor(
         loginUserOp: LoginUserOperation,
         operations: List<Operation>,
     ): ExecutionResponse {
+        // Handle a bad state that can happen in User Model 5.1.27 or earlier versions that old Login
+        // request is not removed after processing if app is force-closed within the PostCreateDelay.
+        // Login operation being processed alone will surely be rejected, so we need to drop the request
+        if (operations.isEmpty()) {
+            return ExecutionResponse(ExecutionResult.FAIL_NORETRY)
+        }
```
jkasten2, line 74: If `loginUserOp.externalId` is set then the operation is still valid, we probably shouldn't drop that.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
jinliu9508: I have updated the PR description, code, and test cases so that Login with external ID is no long be dropped. Only anonymous Login that is not grouped with any other operations will be dropped.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -327,17 +317,21 @@ internal class OperationRepo(
     }
 
     /**
-     * Wait which ever is longer, retryAfterSeconds returned by the server,
+     * Wait which ever is longer, post create delay, retryAfterSeconds returned by the server,
```
nan-li, line 320: The PostCreateDelay should probably be a separate delay from the original intention of this method, which is to generate a delay in failure scenarios, since it is operating on the `retryWaiter`.

nan-li (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
jinliu9508: I think the delays should be interceptable because if both postCreateDelay and retryDelay are set to 5 seconds, the total wait time should just be 5 seconds instead of 10 seconds if wait for them separately.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/operations/LoginUserOperationExecutorTests.kt
```diff
@@ -679,4 +701,42 @@ class LoginUserOperationExecutorTests : FunSpec({
             )
         }
     }
+
+    test("ensure anonymous login with no other operations will fail with FAIL_NORETRY to release the queue from a bad state") {
```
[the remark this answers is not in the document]
jinliu9508: Good point! I have fixed the explanation and created a new test case in OperationRepoTests to test this behavior.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: We shouldn't confound the `waiter` with the `retryWaiter` just because in this particular scenario, the result feels the same. The `retryWaiter` is meant for retryable errors and can be forced when consent changes. 

Both are not going to run: (1) The postCreateDelay is after a create call is successful and the response has ID translations, (2) The retries count and retryAfterSeconds that is original to this method is for failures. Either the operations are successful and contain translation IDs or it is not and is retryable.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
```diff
@@ -66,6 +66,12 @@ internal class LoginUserOperationExecutor(
         loginUserOp: LoginUserOperation,
         operations: List<Operation>,
     ): ExecutionResponse {
+        // Handle a bad state that can happen in User Model 5.1.27 or earlier versions that old Login
+        // request is not removed after processing if app is force-closed within the PostCreateDelay.
+        // Anonymous Login being processed alone will surely be rejected, so we need to drop the request
+        if (operations.isEmpty() && loginUserOp.externalId == null) {
```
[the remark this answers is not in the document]
jinliu9508: I was thinking to prevent dropping the login request with an externalId. However, even login with an externalId will usually require a transfer subscription so theoretically these will have the same result. Do you think we should allow login operation with an externalId that has no other operations to be processed?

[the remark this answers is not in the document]
nan-li: Yeah we should keep login with external ID regardless of subscriptions, since generally the SDK can recover in that state.
I meant for the check itself: `operations.isEmpty() && loginUserOp.externalId == null` vs `operations.doesNotContainSub() && loginUserOp.externalId == null`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
jinliu9508: I see, the waiters are different. I separated the logic in the fixup commit. Both waiters can be waiting together for different duration, but operations processing will be resumed as soon as both waiter are awake. In this case the stall time is the same as the maximum wait time among them.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -345,6 +335,26 @@ internal class OperationRepo(
...
+     * reflect the change before we do any other operations to it.
+     * NOTE: Future: We could run this logic in a
+     * coroutineScope.launch() block so other operations not
+     * effecting this these id's can still be done in parallel,
+     * however other parts of the system don't currently account
+     * for this so this is not safe to do.
+     */
+    suspend fun delayForPostCreate(hasIDTranslation: Boolean) {
+        val postCreateDelay: Long = _configModelStore.model.opRepoPostCreateDelay
+        if (!hasIDTranslation || postCreateDelay == 0L) {
```
jkasten2, line 349: Remove check for `postCreateDelay == 0L to simply this code`.
* It's valid to `delay(0)`, it behaves just a like `yield` which is ok in this scenarion.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -345,6 +335,26 @@ internal class OperationRepo(
...
+     * Stall processing the queue so the backend's DB has to time
+     * reflect the change before we do any other operations to it.
+     * NOTE: Future: We could run this logic in a
+     * coroutineScope.launch() block so other operations not
+     * effecting this these id's can still be done in parallel,
+     * however other parts of the system don't currently account
+     * for this so this is not safe to do.
+     */
+    suspend fun delayForPostCreate(hasIDTranslation: Boolean) {
+        val postCreateDelay: Long = _configModelStore.model.opRepoPostCreateDelay
```
jkasten2, line 348: Can allow type inferencing here, remove ": Long".

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -345,6 +335,26 @@ internal class OperationRepo(
...
+    /**
+     * Stall processing the queue so the backend's DB has to time
+     * reflect the change before we do any other operations to it.
+     * NOTE: Future: We could run this logic in a
+     * coroutineScope.launch() block so other operations not
+     * effecting this these id's can still be done in parallel,
+     * however other parts of the system don't currently account
+     * for this so this is not safe to do.
+     */
+    suspend fun delayForPostCreate(hasIDTranslation: Boolean) {
```
nan-li, line 347: Nit: I think this method should be conditionally called, instead of letting this method check if it should delay using the passed in argument.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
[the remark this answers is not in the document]
jkasten2: I agree, with `operations.doesNotContainSub() && loginUserOp.externalId == null`. This way the validation matches what the backend does.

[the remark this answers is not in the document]
jinliu9508: I made a change that checks if there exists any CreateSubscription or TransferSubscription operation. Will this be sufficient?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
jinliu9508: Good point, I have made that change in the fixup commit.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -327,17 +317,21 @@ internal class OperationRepo(
...
         retryAfterSeconds: Int?,
+        postCreateDelay: Long,
     ) {
         Logging.debug("retryAfterSeconds: $retryAfterSeconds")
         val retryAfterSecondsNonNull = retryAfterSeconds?.toLong() ?: 0L
         val delayForOnRetries = retries * _configModelStore.model.opRepoDefaultFailRetryBackoff
-        val delayFor = max(delayForOnRetries, retryAfterSecondsNonNull * 1_000)
+        val delayFor = max(delayForOnRetries, max(retryAfterSecondsNonNull * 1_000, postCreateDelay))
         if (delayFor < 1) return
         Logging.error("Operations being delay for: $delayFor ms")
```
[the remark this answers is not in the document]
jinliu9508: The log will not be related to the postCreateDelay anymore since it is moved to a separate logic.

jkasten2 (approved):

nan-li (approved):
Can clean up commit history, and reword the PR title?

jinliu9508:
Title is updated and commits are cleaned up.

PR #2259, opened 2025-02-20
Fix: Notification click not foreground the app in the first click if app is closed and no clickListener is added

jinliu9508:
# Description
## One Line Summary
This PR fixes a bug where clicking on a notification does not always bring the app to the foreground under certain conditions.

### Motivation
Some users have reported that clicking a notification for the first time does not foreground the app when it is fully closed. This creates unexpected behavior, and we aim to ensure the app opens correctly when a notification is clicked.

### Scope
This PR will result in two app behavior changes:
* The app will now always foreground when a notification is clicked if the In-App Messaging module is included.
* If the module is not included, the notification click will not foreground the app unless 'addClickListener' has been called.

### OPTIONAL - Root cause explained
  1. When a notification is clicked, 'NotificationOpenedProcessor' determines whether the notification should open.
  2. 'NotificationLifecycleService.canOpenNotification' is accessed, and its result depends on whether:
    * There is a subscriber in 'extOpenedCallback', or
    * 'setInternalNotificationLifecycleCallback' is called and subscribed.
  3. If 'addClickListener' is used, 'extOpenedCallback' has a subscriber, so 'canOpenNotification' returns 'true', allowing the app to open. This explains why adding 'addClickListener()' works as a workaround.
  4. However, if 'addClickListener' is never used, we rely on 'setInternalNotificationLifecycleCallback', which is only called in 'InAppMessagePreviewHandler.start()'. Since this runs asynchronously after 'initWithContext', accessing it before 'start()' always returns 'false'.
  5. This also explains why the second click opens the app—because 'start()' has already been called after the first click.

# Testing
## Unit testing
(Pending on whether the behavior changes under Scope is what we wanted)

## Manual testing
To reproduce the scenario:
  1. Switch to the first commit that removes all click listeners.
  2. Run the app, consent to privacy, and grant push notification permissions.
  3. Send a push notification.
  4. Move the app to the background and force close it.
  5. Click on the notification.
Expected: The notification should open the app.

Observed (before fix): The first click neither opened the app nor cleared the notification.
After the fix, we can observe that the first click always open the app and clear the notification tray.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass


jinliu9508:
Squashing and removing commits around reproducing the issue.

PR #2270, opened 2025-03-11
fix: restore subscription from backend if the subscription ID was not…

jinliu9508:
# Description
## One Line Summary
This PR fixes a bug where the push subscription is no longer updated, and In-App Messages (IAMs) are not displayed after downgrading the Android SDK from 5.2.0-beta to a 5.1.x version.

### Motivation
When downgrading from 5.2.0-beta, an issue may arise where backend requests are executed, but local data is not correctly updated. This leads to the SDK mistakenly creating a duplicate create-push-subscription request instead of fetching the existing push subscription information from the backend. This PR addresses that issue by ensuring proper data synchronization.

### Scope
The changes in this PR are limited to RefreshUserOperation. A edge case will be the push subscription from the backend was created from a different device. But by the time RefreshUserOperation is created, we have an unique OneSignalID so the edge case won't happen.

### Root Cause Explained
In some cases, responses from backend requests such as CreateUser and CreateSubscription in 5.2.0-beta were not properly saved locally. As a result, the SDK could enter a state where no push subscription ID exists locally, even though the backend has all the necessary data.

To resolve this, the PR introduces a condition check that compares the local push subscription state with backend data. If an inconsistency is detected, the SDK will use the push subscription information from the RefreshUser request instead of making an unnecessary create-push-subscription request.

# Testing

## Manual testing
Step to reproduce:
  1. Install 5.2.0-beta, close the app, and reopen it. A 400 error occurs.
  2. Downgrade to 5.1.29. After setting setPaused(false), no IAMs are displayed, and the push subscription is not updated.

Expected Behavior After the Fix:
  1. The push subscription ID is correctly updated.
  2. All IAMs display as expected.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2270)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
@@ -94,6 +94,7 @@ internal class RefreshUserOperationExecutor(
             }
 
             val subscriptionModels = mutableListOf<SubscriptionModel>()
+            var backendPushSubscription: SubscriptionModel? = null
             for (subscription in response.subscriptions) {
                 val subscriptionModel = SubscriptionModel()
                 subscriptionModel.id = subscription.id!!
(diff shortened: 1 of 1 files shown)


jkasten2 (changes requested):
This approach isn't going to work, the user can have multiple push subscriptions, and there isn't a way to know which one is owned by this device.
* It's possible to compare pushTokens, but the device isn't guaranteed to have one either.

If are in the state where the OneSignal id is valid (fetches and updates to the User are fine) but we get 404s on accessing the push subscription then we should create a new push subscription on the user to recover.

jinliu9508:
Closing this PR. The new approach is in this new PR https://github.com/OneSignal/OneSignal-Android-SDK/pull/2271

PR #2271, opened 2025-03-20
Fix subscription/IAM not updated after upgrading from 5.2.0-beta or between 5.1.9 to 5.1.27

jinliu9508:
# Description
## One Line Summary
Added a MigrationRecovery that detects and recovers a bad state that push sub ID is missing in the config store, caused by a bug in 5.2.0-beta or any version between 5.1.9-5.1.27.

### Motivation
5.2.0-beta and all versions between 5.1.9-5.1.27 have a specific logic for push subscription that depends on the result of a remote params fetch. It may cause the push sub ID to be missing from the config model if a postCreateDelay bug also happens that the app is closed within 5 seconds of first start.

### Scope
A new MigrationRecovery that will check the status of the config model store and the subscription model store at the start() level. 

### OPTIONAL - Other
Interface IMigrationRecovery and a base class MigrationRecovery are added for the ease of creating more recovery in the future.

# Testing
## Unit testing

## Manual testing

Step to reproduce:
1. Check out the commit that release 5.2.0-beta that is specifically based on 5.1.25. Beta rebased to 5.1.27 will not have this issue.
2. close the app within 5 seconds, and reopen it. A 400 error occurs.
3. Downgrade to 5.1.29. After setting setPaused(false), no IAMs are displayed, and the push subscription is not updated.

Expected Behavior After the Fix:
After downgrade, the push subscription ID is correctly updated. All IAMs are displayed as expected.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing
     - PR also includes a new interface IMigrationRecovery and a base class MigrationRecovery for the ease of creating new recovery if needed in the future.

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2271)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
@@ -1,0 +1,13 @@
+package com.onesignal.user.internal.migrations
+
+import com.onesignal.debug.internal.logging.Logging
+
+abstract class MigrationRecovery : IMigrationRecovery {
+    override fun start() {
+        // log a message and take the corrective action if it is determined that app is in a bad state
+        if (isInBadState()) {
+            Logging.warn(recoveryMessage())
+            recover()
+        }
+    }
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
@@ -17,6 +17,7 @@ import com.onesignal.user.internal.backend.impl.UserBackendService
 import com.onesignal.user.internal.builduser.IRebuildUserService
 import com.onesignal.user.internal.builduser.impl.RebuildUserService
 import com.onesignal.user.internal.identity.IdentityModelStore
+import com.onesignal.user.internal.migrations.RecoverConfigPushSubscription
 import com.onesignal.user.internal.migrations.RecoverFromDroppedLoginBug
 import com.onesignal.user.internal.operations.impl.executors.IdentityOperationExecutor
 import com.onesignal.user.internal.operations.impl.executors.LoginUserFromSubscriptionOperationExecutor
@@ -74,6 +75,7 @@ internal class UserModule : IModule {
         builder.register<UserRefreshService>().provides<IStartableService>()
 
         builder.register<RecoverFromDroppedLoginBug>().provides<IStartableService>()
+        builder.register<RecoverConfigPushSubscription>().provides<IStartableService>()
 
         // Shared state between Executors
         builder.register<NewRecordsState>().provides<NewRecordsState>()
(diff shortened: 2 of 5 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
```diff
@@ -0,0 +1,25 @@
+package com.onesignal.user.internal.migrations
+
+import com.onesignal.debug.internal.logging.Logging
+
+open class MigrationRecovery : IMigrationRecovery {
```
jkasten2, line 5: Let's make this an abstract class, as it isn't indented to be used directly, only extended.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
```diff
@@ -0,0 +1,25 @@
...
+        return false
+    }
+
+    override fun recover() {
+        // left blank intentionally
+    }
+
+    override fun recoveryMessage(): String {
+        return ""
+    }
```
jkasten2, line 16: If this is made an abstract class then these 3 methods do not need to be defined here.

jkasten2 (approved):
Some improvements could be made to MigrationRecovery but everything else looks good.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
jinliu9508: Good point! This was actually designed but was playing with it as a base class.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
jinliu9508: I have updated the code and tested the change in the latest push.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/MigrationRecovery.kt
```diff
@@ -0,0 +1,19 @@
+package com.onesignal.user.internal.migrations
+
+import com.onesignal.debug.internal.logging.Logging
+
+abstract class MigrationRecovery : IMigrationRecovery {
+    abstract override fun isInBadState(): Boolean
+
+    abstract override fun recover()
+
+    abstract override fun recoveryMessage(): String
```
jkasten2, line 10: These lines don't need to be here at all, these all implied and the complier doesn't require them.

jinliu9508: Ok I have removed them in the latest push.

PR #2281, opened 2025-04-10
Fix: ANR caused by model.add(), model.initializeFromJson(), or modelstore.load()

jinliu9508:
# Description
## One Line Summary
Fix the ANR that can caused by model.add(), model.initializeFromJson(), or modelstore.load()

### Motivation
Fix the ANR that may potentially interrupt the normal usage of apps. 

### Scope
OperationRepo.enqueue() now performs internal enqueue in a specific coroutine and the operation may not be added instantly even the method returns immediately. The purpose of it is to prevent a potential deadlock situation when enqueuing is running into loading. This deadlock currently only happens to OperationModelStore because it is loaded in a background thread to optimize the starting process. When operations like addTag or optIn is called immediately after initWithContext, there is a chance to lock some specific resource that is needed for the loading to complete. 

# Testing
## Unit testing
Was not able to make a targeted unit test because the deadlock situation involves instances not only in modeling but also in the app level. 

## Manual testing
The first commit simulates a deadlock situation by calling addTag shortly after initWithContext and purposely delay model.initializeFromJson for addTag in the main thread to be blocked. Now if we run the test app, it will freeze on the startup screen.

After the fix, we can see that app does not freeze anymore.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2281)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -1,5 +1,6 @@
 package com.onesignal.core.internal.operations.impl
 
+import com.onesignal.common.threading.OSPrimaryCoroutineScope
 import com.onesignal.common.threading.WaiterWithValue
 import com.onesignal.core.internal.config.ConfigModelStore
 import com.onesignal.core.internal.operations.ExecutionResult
@@ -102,6 +103,13 @@ internal class OperationRepo(
         }
     }
 
+    /**
+     *  Enqueuing will be performed in a designate coroutine and may not be added instantly.
+     *  This is to prevent direct enqueuing from the main thread that may cause a deadlock if loading
+     *  is still in process.
+     *
+     *  If the enqueuing needs to be suspended, use enqueueAndWait() instead.
+     */
     override fun enqueue(
         operation: Operation,
         flush: Boolean,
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
```diff
@@ -81,6 +84,11 @@ open class Model(
     fun initializeFromJson(jsonObject: JSONObject) {
         synchronized(data) {
             data.clear()
+            if (this is LoginUserOperation) {
+                // loading operationModelStore will get stuck for 15 seconds while locking data
+                Logging.log(LogLevel.DEBUG, "initializeFromJson LoginUserOperation")
+                Thread.sleep(15000)
+            }
```
jkasten2, line 91: Since this is test code, revert this file before merging.

jkasten2 (changes requested):
Logic changes themselves look good.

The changes done to `Model.kt` need to be reverted before merging though. I can approve once that last thing is done.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
jinliu9508: I just pushed the third commit that reverts all test code. I will squash the test commits before merging.

jkasten2 (approved):

jinliu9508:
Squashing the commits for test codes before merging.

PR #2284, opened 2025-04-18
Fix: IAM now showing after switching to 3-button-navigation mode

jinliu9508:
# Description
## One Line Summary
Fix some IAMs being suppressed under some specific view layouts with the 3 button navigation mode

### Motivation
Fix the bug that causes IAM not shown when user is on the 3 button navigation mode.

### Scope
The PR changes how we determine if keyboard is currently open by using WindowInsets instead of using a fixed value for the offset of current view height and the screen height. This change will only be available for devices running on Android 11 and up. 

# Testing
## Manual testing
I was unable to reproduce the scenario that prevent IAM from showing. Instead, I tested whether isKeyboardUp returns the correct value with the soft keyboard being open and close. 

Devices I tested:
  - Pixel 9 API 35: used imeInsets logic and returned the correct output
  - Pixel 2 API 29: used visible bound height diff and returned the correct output

Both 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2284)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/DeviceUtils.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/DeviceUtils.kt
@@ -4,9 +4,11 @@ import android.app.Activity
 import android.content.Context
 import android.graphics.Rect
 import android.net.ConnectivityManager
+import android.os.Build
 import android.telephony.TelephonyManager
 import android.util.DisplayMetrics
 import android.view.View
+import androidx.core.view.WindowInsetsCompat
 import java.lang.ref.WeakReference
 
 object DeviceUtils {
@@ -14,7 +16,6 @@ object DeviceUtils {
 
     /**
      * Check if the keyboard is currently being shown.
-     * Does not work for cases when keyboard is full screen.
      */
     fun isKeyboardUp(activityWeakReference: WeakReference<Activity?>): Boolean {
         val metrics = DisplayMetrics()
@@ -28,8 +29,15 @@ object DeviceUtils {
             window.windowManager.defaultDisplay.getMetrics(metrics)
         }
         if (view != null) {
-            val heightDiff = metrics.heightPixels - visibleBounds.bottom
-            isOpen = heightDiff > MARGIN_ERROR_PX_SIZE
+            isOpen =
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    val imeInsets = view.rootWindowInsets
+                    imeInsets.isVisible(WindowInsetsCompat.Type.ime())
+                } else {
+                    // Does not work for cases when keyboard is full screen for Android 9 and below
+                    val heightDiff = metrics.heightPixels - visibleBounds.bottom
+                    heightDiff > MARGIN_ERROR_PX_SIZE
+                }
         }
         return isOpen
     }


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/DeviceUtils.kt
```diff
@@ -28,8 +30,15 @@ object DeviceUtils {
             window.windowManager.defaultDisplay.getMetrics(metrics)
         }
         if (view != null) {
-            val heightDiff = metrics.heightPixels - visibleBounds.bottom
-            isOpen = heightDiff > MARGIN_ERROR_PX_SIZE
+            isOpen =
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    val imeInsets = view.rootWindowInsets.getInsets(WindowInsetsCompat.Type.ime())
+                    // ensure input method size is 0
+                    !(imeInsets.left == 0 && imeInsets.top == 0 && imeInsets.right == 0 && imeInsets.bottom == 0)
```
nan-li, line 37: Is this logic necessary? Can we use the `isVisible` AP at https://developer.android.com/reference/android/view/WindowInsets#isVisible(int) ?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/DeviceUtils.kt
```diff
@@ -28,8 +30,15 @@ object DeviceUtils {
             window.windowManager.defaultDisplay.getMetrics(metrics)
         }
         if (view != null) {
-            val heightDiff = metrics.heightPixels - visibleBounds.bottom
-            isOpen = heightDiff > MARGIN_ERROR_PX_SIZE
+            isOpen =
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                    val imeInsets = view.rootWindowInsets.getInsets(WindowInsetsCompat.Type.ime())
+                    // ensure input method size is 0
```
jkasten2, line 36: This comment isn't clear to me, can you update it or remove it?

jkasten2 (changes requested):
There is a comment "Does not work for cases when keyboard is full screen.", is this still true with the new API used in this PR?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/DeviceUtils.kt
jinliu9508: Good point, I have modified the code to use isVisible instead.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/DeviceUtils.kt
jinliu9508: I moved the comment to the code block for Android 9 and below to better explain the change. Confirm that the new API works with a full screen keyboard.

jkasten2 (approved):

PR #2287, opened 2025-04-25
Fix: IAM not shown when calling login immediately after init

jinliu9508:
# Description
## One Line Summary
Fix InAppMessagesManager so it now adds a IAM fetch condition in the event of user change instead of start(), to prevent IAM fetch conditions being created with a local onesignal ID.

### Motivation
Fix the issue when IAM is not showing on fresh install if Login is called immediately after initWithContext. This is caused by InAppMessagesManager creating IAM fetch conditions with local OneSignalID and not recognizing an updated ID to resolve. The PR adds a user change handler to add the IAM fetch condition in the event of retrieving a new OneSignalID. 

### Scope
Every time a OneSignalID is changed from a local ID to a backend ID, a IAM fetch condition will be created to attempt fetching IAMs. 

# Testing
## Manual testing
Issue reproducible by a fresh install that is calling Login immediately after initWithContext. We could observe that IAM does not show on the first startup. 

After the fix, IAM shows successfully while the ryw rule stays effective.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2287)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/backend/IIdentityBackendService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/backend/IIdentityBackendService.kt
@@ -38,7 +38,7 @@ interface IIdentityBackendService {
     )
 }
 
-internal object IdentityConstants {
+object IdentityConstants {
     /**
      * The alias label for the external ID alias.
      */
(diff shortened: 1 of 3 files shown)


jkasten2 (changes requested):
You noted in the "Manual testing" section of the PR you tested "... a fresh install that is calling Login immediately after initWithContext.". Can you test the following scenarios as well?
* Fresh install without calling login
* Restarting the app after login
* Restarting the app (login never called).

jinliu9508:
> You noted in the "Manual testing" section of the PR you tested "... a fresh install that is calling Login immediately after initWithContext.". Can you test the following scenarios as well?
> 
> * Fresh install without calling login
> * Restarting the app after login
> * Restarting the app (login never called).

Good point. Here the the test result:
  * Fresh install without calling login: pass
  * Restarting the app login never called: pass
  * Restarting the app after login: pass

Additionally, I also test with IAM with triggers:
  * Fresh install without calling login, then add trigger: pass
  * Restarting the app login never called, then add trigger: pass
  * Restarting the app after login, call the segment specific trigger: pass

jkasten2 (approved):

PR #2290, opened 2025-05-02
Maintain: fix flaky test unit caused by enqueue not complete instantly

jinliu9508:
Since OperationRepo.enqueue() was changed to internal enqueue in a separate coroutine scope, we can no longer expect calling OperationRepo.enqueue() can instantly insert the operation. Adding a delay after each enqueue call in the unit test to give time for the internal enqueuing to actually complete.

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2290)


jkasten2 (changes requested):
We shouldn't use delays and timeouts when it is feasible for tests. The root cause of this  flaky test is going to be common, so let's invest in a better way to wait for the right event, instead of time based.

Issue #2279, opened 2025-04-10
[Bug]: addAlias with some values throws, example org.json.JSONException: Value 1585567 at Al of type java.lang.String cannot be converted to boolean


KhushbuDX:
### How can we help?

I also facing the Issue while using addAlias with "Al" as label
It will allow user to login first time after reopen the app It will crash
Below are the crash Logs

java.lang.reflect.InvocationTargetException
at java.lang.reflect.Constructor.newInstance0(Native Method)
at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:89)
at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:82)
at com.onesignal.common.services.ServiceProvider.getAllServices(ServiceProvider.kt:55)
at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:72)
at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
at com.onesignal.internal.OneSignalImp.initWithContext(OneSignalImp.kt:554)
at com.onesignal.OneSignal.initWithContext(OneSignal.kt:135)
at AppController.setupOneSignal(AppController.kt:140)
at application.AppController.onCreate(AppController.kt:111)
at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1266)
at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7619)
at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2400)
at android.os.Handler.dispatchMessage(Handler.java:106)
at android.os.Looper.loopOnce(Looper.java:226)
at android.os.Looper.loop(Looper.java:313)
at android.app.ActivityThread.main(ActivityThread.java:8762)
at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #2296, opened 2025-05-09
Fix crash when adding alias with some labels

jkasten2:
# Description
## One Line Summary
When passing some labels to `User.addAlias`, such as "AL", would cause a crash when starting the app the next time.

## Details
This is due to casting the wrong type as initializeFromJson's code used a contains instead of an equals which can result in finding the wrong function.

To correct this switched this to equals("get$property", true) instead and also added a check for it.parameterCount == 0 as we only want to cast based on getters.

Lastly also removed the if (method == null) check to simplify the code, as the when statement can handle this case.

### Related
This address issue #2279

### Motivation
Valid input values to the SDK should not cause crashes.

### Scope
Affects casting logic loading Models from storage on app startup.

# Testing
## Unit testing
Also added tests for IdentityModel with a number of labels to ensure this code changes covers common label values.

## Manual testing
Tested on an Android 14 emulator, added an alias with the label "AL", reproduced crash, and ensured no longer crashes after the fix.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2296)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
@@ -94,28 +94,25 @@ open class Model(
                         data[property] = listOfItems
                     }
                 } else {
+                    // Cast JSON value based on class's property getter name return type
                     val method =
                         this.javaClass.methods.firstOrNull {
-                            it.returnType !=
-                                Void::class.java &&
-                                it.name.contains(
-                                    property,
+                            it.returnType != Void::class.java &&
+                                it.parameterCount == 0 &&
+                                it.name.equals(
+                                    "get$property",
                                     true,
                                 )
                         }
 
-                    if (method == null) {
-                        data[property] = jsonObject.get(property)
-                    } else {
-                        when (method.returnType) {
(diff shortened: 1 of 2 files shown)


jinliu9508 (approved):

jkasten2:
Was thinking about switching the implementation to use `getDeclaredMethod` to simply and improve this code performance. However since we depend on case-insensitivity this can't be used on it's own. It's possible we could use `getDeclaredMethod` if we can accurately generate the correct case, however validating this with all fields would take sometime and the bigger one, would make things more fragile.

On the performance front, calling `this.javaClass.methods` is fairly expensive (from memory when we did some performance testing a while back), this could be cached. However doing this change later would be better, so we don't increase the risk of breaking something and to get this bug fix shipped faster.

PR #2304, opened 2025-05-20
Fix: NotificationOpenedActivity "freeze" when a large amount of notification is clicked

jinliu9508:
# Description
## One Line Summary
Move the backend call `updateNotificationsAsOpened` to a background thread to prevent blocking the next activity from opening when many notifications are clicked.

### Motivation
Fix a frozen screen issue that occurs when a large number of notifications are clicked simultaneously.

### Scope
- The `updateNotificationsAsOpened` operation is now performed silently in the background. This means we no longer wait for its response before opening the destination activity upon a notification click. As a result, the order of this operation relative to other operations is no longer guaranteed. A potential risk is that IDs may change before `updateNotificationsAsOpened` completes on the backend. However, since all data is finalized at the time the requests are created, data integrity is maintained without blocking the UI.
- OSPrimaryCoroutineScope is also used by other backend operations like `createUser` and `updateSubscription`. As a result, these operations could potentially be delayed if too many `updateNotificationsAsOpened` calls are made.
One possible improvement would be to introduce a `SecondaryCoroutineScope` to handle lower-priority updates, but since they’re all backend calls, the overall time required to complete them remains roughly the same.
- ⚠️ Important: If the app is force-closed shortly after a notification click, some `updateNotificationsAsOpened` calls may be interrupted, and those clicks might not be updated on the backend.

# Testing
## Unit testing
A new unit test ensures that `openDestinationActivity` is not blocked when clicking on a large number of notifications.

## Manual testing
Tested on a Pixel 9 (API 35) emulator. Clicking a summary notification that groups 50+ individual notifications opens the app seamlessly. The `updateNotificationsAsOpened` calls are successfully processed in the background after the app is opened.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [x] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2304)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
@@ -7,6 +7,7 @@ import com.onesignal.common.JSONUtils
 import com.onesignal.common.events.CallbackProducer
 import com.onesignal.common.events.EventProducer
 import com.onesignal.common.exceptions.BackendException
+import com.onesignal.common.threading.OSPrimaryCoroutineScope
 import com.onesignal.core.internal.application.AppEntryAction
 import com.onesignal.core.internal.application.IApplicationService
 import com.onesignal.core.internal.config.ConfigModelStore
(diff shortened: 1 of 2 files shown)


jkasten2 (commented):
Do we need `suspend` on `notificationOpened` any more since we moved the network call to `OSPrimaryCoroutineScope`?

jinliu9508:
> Do we need `suspend` on `notificationOpened` any more since we moved the network call to `OSPrimaryCoroutineScope`?

Good question! I think we still need it because openDestinationActivity is suspending function that is also called in notificationOpened() and it is used on two different places. So refactoring this will be out of the scope of this PR. I also checked updateNotificationAsOpened and it looks like it is still suspending. We just need to make it suspend in a background thread so its run time doesn't scale up in the main thread.

jkasten2 (changes requested):

jkasten2 (approved):

PR #2311, opened 2025-06-03
Feat: add public log listener methods

jkasten2:
# Description
## One Line Summary
Adds the following new public methods to allow an app developer to capture logs as strings at runtime.

## Details
Adds the following new public methods:
* `OneSignal.Debug.addLogListener`
* `OneSignal.Debug.removeLogListener`

These new methods provide a way to capture all SDK log entires at runtime, allowing the app developer to store these and/or send them to their server.

The log listener is independent of logLevel, all message are always sent to listeners.

### Motivation
App developers want a way to capture all SDK logs and send them to their server to debug issues in production.

### Scope
Only add new methods to OneSignal.Debug to allow of capturing logs at runtime.

# Testing
## Unit testing
Added a new `LoggingTests` class to cover the new public methods.

## Manual testing
Tested on an Android 15 emulator.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [X] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing
       - Failing due to unrelated flaky OperationRepoTests test.

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2311)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/IDebugManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/IDebugManager.kt
@@ -3,17 +3,29 @@ package com.onesignal.debug
 /**
  * Access to debug the SDK in the event additional information is required to diagnose any
  * SDK-related issues.
- *
- * WARNING: This should not be used in a production setting.
  */
 interface IDebugManager {
     /**
      * The log level the OneSignal SDK should be writing to the Android log. Defaults to [LogLevel.WARN].
+     * WARNING: This should not be set higher than LogLevel.WARN in a production setting.
      */
     var logLevel: LogLevel
 
     /**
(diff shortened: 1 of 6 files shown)


jkasten2:
> I added Logging.log inside the onLogEvent and it looks like it is creating logs non-stop. Should we add a note to not calling log inside the log event?

* `Logging` should be used outside of the OneSignal SDK, it's been public since the 5.0.0 release but it was not indented and isn't documented.
* The logging listener isn't intended to be used inside the SDK, only externally by apps.

Given the above I don't think we need account for this.

PR #2318, opened 2025-06-17
Maintain: fix flaky tests in operationRepoTests

jinliu9508:
# Description
## One Line Summary
Fixed some flaky test units in operationRepoTests due to previous commits

### Motivation
Fix the flaky test units that have been blocking the runner to complete successfully

## Other
This PR also added a waitForIdle() method for testing in OSPrimaryCoroutineScope to allow thread to wait for the completion of operation enqueue. 

# Testing
## Unit testing
The test units are fixed and passed.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2318)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OSPrimaryCoroutineScope.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OSPrimaryCoroutineScope.kt
@@ -16,4 +16,6 @@ object OSPrimaryCoroutineScope {
             block()
         }
     }
+
+    suspend fun waitForIdle() = mainScope.launch { }.join()
 }
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
@@ -1,5 +1,6 @@
 package com.onesignal.core.internal.operations
 
+import com.onesignal.common.threading.OSPrimaryCoroutineScope
 import com.onesignal.common.threading.Waiter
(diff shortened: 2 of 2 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OSPrimaryCoroutineScope.kt
```diff
@@ -1,19 +1,50 @@
...
+            val completionJob =
+                mainScope.launch {
+                    activeJobs.forEach { it.join() }
+                    continuation.resume(Unit)
+                }
+
+            continuation.invokeOnCancellation {
+                completionJob.cancel()
+            }
+        }
```
jkasten2: I don't think we need to track a list of active jobs. Instead in `waitForIdle()`, why not just add a job and wait for it to be completed?

I think something like this would work:
```Kotlin
suspend fun waitForIdle() = mainScope.launch { }.join()
```

The blank job will only finish once all jobs before it are done, so we don't need to also track all the jobs with our own set/list.

jinliu9508: Good idea! Was trying to resume the blocking until completion, but didn't realize that mainScope is all we need to look after. I made the change in the fixup commit.

jkasten2 (approved):

PR #2326, opened 2025-06-23
chore: add github-to-asana workflows

sherwinski:
# Description
## Summary
Adds three new GitHub Actions that will create/modify Asana tasks:
1. Create a new Asana task when a new GitHub issue is created
2. Add a comment to an existing Asana task when a GitHub user comments on an issue
3. Add a comment to an existing Asana task when any action is performed on a GitHub issue (e.g. assignee changed, labels added/removed, etc.)

### Motivation
This helps capture activity and potential bugs from this repo in Asana for better organization and project maintenance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2326)


.github/workflows/asana-add-comment.yml
```diff
@@ -0,0 +1,44 @@
...
+    permissions:
+      issues: read
+    steps:
+        - name: Get Asana Task Corresponding to Issue
+          env:
+            ISSUE_ID: ${{ github.event.issue.id }}
+            WORKSPACE_ID: "780103692902078"
+          run: |
+            curl --request GET \
+                 --url "https://app.asana.com/api/1.0/workspaces/$WORKSPACE_ID/tasks/search?opt_fields=notes&text=$ISSUE_ID&sort_by=modified_at&sort_ascending=false" \
```
jkasten2: In relation to `opt_fields=notes&text=$ISSUE_ID` is the `github.event.issue.id` value a more unique value than the public facing "number"? If so, what format is the "id"? If it isn't something that is universally unique, like a UUID, we should include the repo name + this "id" in this search, otherwise it's going to find false positives.

sherwinski: The ID looks something like `3147054299`. I don't know if it's a UUID specifically, but here is the [description](https://docs.github.com/en/rest/using-the-rest-api/issue-event-types?apiVersion=2022-11-28#issue-event-object-common-properties) on GitHub's docs:

> The unique identifier of the event.

So I figured it would be safe to use for this case.

sherwinski: I also explored building an extra layer of search to narrow it down by the `DSA Repo Ticket Url` field, but it looks like you can't combine string search terms with an AND operator in that way.

jkasten2: It seems `github.event.issue.id` is a 10 digit number from some examples I have seen, I can see us running into collisions in some rare cases. Can we add the repo name to the start of the number to avoid that?
* The notes field can be pretty long, and it could contain a log with timestamps (which are also 10 digits) so I am also afraid of false positives there.

sherwinski: I tested prepending the full repo name to the unique issue ID, e.g. `sherwinski/github-to-asana#3176512565`, which is still searchable by the API. If that helps add some reassurance against possible collisions then I can add that into this PR.

jkasten2: Ya that works, let me know when you add that and I can approve this PR.

jkasten2 (approved):

PR #2327, opened 2025-06-26
fix: logout incorrectly uses the old subscription ID

jinliu9508:
# Description
## One Line Summary
Fixes a bug where the subscription status is lost when a subscription is deleted via the OneSignal dashboard after 'login()' is called.

### Motivation
There is a use case where inactive subscriptions are manually removed from the dashboard. If the SDK subsequently calls 'login()' or 'logout()' after such deletion, a new subscription is created, but its status defaults to 'NEVER_SUBSCRIBE' — even though the user had previously subscribed. We expect the newly created subscription to inherit the setup and subscription state of the original.

### OTHER - Root Cause Explained
When 'logout()' is called, the SDK sends a 'login-user' request that includes a 'transfer-subscription' operation. Normally, the backend returns a subscription that matches what is stored locally. However, if the original subscription has been deleted from the dashboard, the backend instead returns a newly created default subscription, which lacks prior subscription attributes such as the subscription status.

### Scope
The 'TransferSubscription' operation is now only added when an 'externalID' is present.
  * If 'externalID' is not present (i.e., during a 'logout()' or first-time 'login()' as an anonymous user), we now use a 'CreateSubscription' operation instead. This will also be consistent with the behavior in our iOS SDK.
  * This change prevents the system from attempting to transfer a deleted subscription, which would otherwise result in a new subscription without inherited subscription state.

# Testing
## Unit testing
Not feasible, as the process involves manual deletion via the dashboard, which cannot be automated in unit tests.

## Manual testing

- [x] Test Case 1 – 'subscribe()' → 'login()' → delete subscription → 'logout()'
  * Expected: A new subscription is created after 'logout()' that retains the original subscription state, including the subscription status.
  * Actual: Matches expected behavior.
  
- [x] Test Case 2 – 'subscribe()' → 'login()' → 'logout()' → 'login()' (same 'externalID')
  * After 'logout()', the 'externalID' is successfully cleared on the dashboard. The push status remains unchanged.
  * On re-login with the same 'externalID', it is correctly re-associated with the subscription.
  
- [x] Test Case 3 – 'subscribe()' → 'login()' → 'logout()' → 'login()' (different 'externalID')
  * The SDK successfully transfers the subscription to the new 'externalID'.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2327)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -467,8 +467,10 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         // to the queue.
         val currentPushSubscription = subscriptionModelStore!!.list().firstOrNull { it.id == configModel!!.pushSubscriptionId }
         val newPushSubscription = SubscriptionModel()
+        val localSubscriptionID = IDManager.createLocalId()
+        val currentSubscriptionID = currentPushSubscription?.id ?: localSubscriptionID
 
-        newPushSubscription.id = currentPushSubscription?.id ?: IDManager.createLocalId()
+        newPushSubscription.id = currentSubscriptionID
         newPushSubscription.type = SubscriptionType.PUSH
         newPushSubscription.optedIn = currentPushSubscription?.optedIn ?: true
         newPushSubscription.address = currentPushSubscription?.address ?: ""
@@ -478,11 +480,6 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         newPushSubscription.carrier = DeviceUtils.getCarrierName(services.getService<IApplicationService>().appContext) ?: ""
         newPushSubscription.appVersion = AndroidUtils.getAppVersion(services.getService<IApplicationService>().appContext) ?: ""
 
-        // ensure we always know this devices push subscription ID
-        configModel!!.pushSubscriptionId = newPushSubscription.id
-
-        subscriptions.add(newPushSubscription)
-
         // The next 4 lines makes this user the effective user locally.  We clear the subscriptions
         // first as a `NO_PROPOGATE` change because we don't want to drive deleting the cleared subscriptions
         // on the backend.  Once cleared we can then setup the new identity/properties model, and add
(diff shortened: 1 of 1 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -477,10 +478,6 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         newPushSubscription.deviceOS = Build.VERSION.RELEASE
         newPushSubscription.carrier = DeviceUtils.getCarrierName(services.getService<IApplicationService>().appContext) ?: ""
         newPushSubscription.appVersion = AndroidUtils.getAppVersion(services.getService<IApplicationService>().appContext) ?: ""
-
-        // ensure we always know this devices push subscription ID
-        configModel!!.pushSubscriptionId = newPushSubscription.id
-
         subscriptions.add(newPushSubscription)
```
jkasten2: `newPushSubscription` doesn't have it's `id` set yet, it would be more clear if we added towards the bottom of this method, after it's `id` is set.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -492,13 +489,22 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+        } else if (currentPushSubscription != null && identityModel.externalId != null) {
+            // case for switching to another user
+            newPushSubscription.id = currentSubscriptionID
             operationRepo!!.enqueue(TransferSubscriptionOperation(configModel!!.appId, currentPushSubscription.id, sdkId))
             subscriptionModelStore!!.replaceAll(subscriptions, ModelChangeTags.NO_PROPOGATE)
         } else {
+            // case for first time login or logout
+            // always use a local subscription ID to indicate a newly created subscription
+            newPushSubscription.id = localSubscriptionID
             subscriptionModelStore!!.replaceAll(subscriptions)
```
jkasten2: Can this also be `ModelChangeTags.NO_PROPOGATE` to match the other two if/if-else blocks above? If so, then we call this once and move this outside of these `if` statements to simplify this code.

jkasten2 (changes requested):
Left two comments to address on specific lines of code.

It would be beneficial to add tests for the code, however since there are pretty significant changes around this code in the 5.2.0 branch it would be a better fit there, outside of this PR.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: I reorganized the code a bit in the fixup commit so 'newPushSubscription' is added after the id is set

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Unfortunately this part requires a NORMAL tag to work properly. Otherwise there will be no create-subscription to be added. I reorganized the code a bit for more clarity.

jkasten2 (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jkasten2: I see. Thanks for explaining and making the code more clear

PR #2329, opened 2025-06-27
[Fix] Android 7 and lower crashes with getParameterCount calls, when the app is built with AGP 7 or older

### LINKED ISSUE #2323: [Bug]: Crash on Android 6-7 when using SDK version 5.1.34 with a Gradle 7 project
### What happened?

A lot of crashes happened recently when OneSignal Android SDK is updated. Seems like 
[Model.initializeFromJson()](https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt#L101) has code causing this crash.

Image

Image

Image

### Steps to reproduce?

```Markdown
1. Install SDK v5.1.34.
2. Launch app normally.
3. Send push to device.
4. Consecutive launches are failed with crash.
```

### What did you expect to happen?

I expected no crash at startup

### OneSignal Android SDK version

5.1.34

### Android version

7.1, 7.0, 6

### Specific Android models

```Markdown

```

### Relevant log output

```Shell
Caused by java.lang.NoSuchMethodError: No virtual method getParameterCount()I in class Ljava/lang/reflect/Method; or its super classes (declaration of 'java.lang.reflect.Method' appears in /system/framework/core-libart.jar)
       at com.onesignal.common.modeling.Model.initializeFromJson(Model.kt:101)
       at com.onesignal.common.modeling.SimpleModelStore.create(SimpleModelStore.kt:30)
       at com.onesignal.common.modeling.ModelStore.load(ModelStore.kt:175)
       at com.onesignal.common.modeling.SimpleModelStore.<init>(SimpleModelStore.kt:23)
       at com.onesignal.core.internal.config.ConfigModelStore.<init>(ConfigModelStore.kt:8)
       at java.lang.reflect.Constructor.newInstance(Constructor.java)
       at com.onesignal.common.services.ServiceRegistrationReflection.resolve(ServiceRegistration.kt:89)
       at com.onesignal.common.services.ServiceProvider.getServiceOrNull(ServiceProvider.kt:79)
       at com.onesignal.common.services.ServiceProvider.getService(ServiceProvider.kt:67)
       at com.onesignal.internal.OneSignalImp.initWithContext(OneSignalImp.kt:570)
       at com.onesignal.OneSignal.initWithContext(OneSignal.java:135)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jkasten2:

# Description
## One Line Summary
Fixes startup crashes with Android 7, on `getParameterCount()`, when the app is built with AGP 7 or older.

## Details
This method isn't compatible with Android 7 and older devices and causes a crash.
AGP 8 (Android Gradle Plugin) added compatibility for this, however its not uncommon for app developers to still be on 7.

The `getParameterCount()` was added to be cautious, however there isn't a known case where this is needed so we removed it in this PR.

### Motivation
We want to still support Android 5-7 versions and Gradle 7.

### Scope
Effects deserialization of JSON into model objects.

### Related
This fixes #2323

# Testing
## Unit testing
No unit test changes were needed.

## Manual testing
Tested on an Android 5 emulator.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
@@ -98,7 +98,6 @@ open class Model(
                     val method =
                         this.javaClass.methods.firstOrNull {
                             it.returnType != Void::class.java &&
-                                it.parameterCount == 0 &&
                                 it.name.equals(
                                     "get$property",
                                     true,


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
```diff
@@ -98,7 +98,6 @@ open class Model(
                     val method =
                         this.javaClass.methods.firstOrNull {
                             it.returnType != Void::class.java &&
-                                it.parameterCount == 0 &&
```
jinliu9508, line 101: I think there may potentially be a behavior change here. When a method has no parameters, it was previously assigned as an object in the old code, but in the changed code, it would be assigned based on its actual return type.

I see that parameterCount was first added in this [PR](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2296/files). Is it necessary to check parameterCount in order to fix that crash? If so, this change of behavior may affect the previous fix for the crash. It will be good to double check with that crash.

jinliu9508 (commented):
I think there may potentially be a behavior change here. When a method has no parameters, it was previously assigned as an object in the old code, but in the changed code, it would be assigned based on its actual return type.

I see that parameterCount was first added in this [PR](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2296/files). Is it necessary to check parameterCount in order to fix that crash? If so, this change of behavior may affect the previous fix for the crash. It will be good to double check with that crash.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt
jkasten2: The main change from that PR was going from `contains` to `equals` to avoid picking the wrong method. While it's possible to have getPropertyX and getPropertyX(param1) this isn't something that should be in our codebase as this is against our coding conventions.  Also, all the tests still pass, that did once fail before code changes in the PR you point out.

PR #2339, opened 2025-07-10
Feature: Custom Event

jinliu9508:
# Description
## One Line Summary
Introduce a new SDK feature: 'User.trackEvent', enabling clients to track custom events at the user level.

### Motivation
Clients often need to track events with custom data tied to individual users at any point during the app lifecycle. This feature addresses that need by providing a streamlined API, enhancing workflow flexibility and allowing for more precise user behavior analysis.

### Scope
* Added a new API: User.trackEvent(name: String, properties: Map<String, Any>), which allows tracking a custom event with associated properties for the current user.
* Custom event operations are cached and retried following the same OperationRepo rules as other user-level operations. However, each trackEvent invocation triggers an immediate network request. Event batching by user is not currently supported but may be considered in a future update.
* Custom events are not persisted locally outside the operation system. They are not retrievable through the SDK after being sent. Clients may need to maintain their own copy if needed.
* Events exceeding 1MB in payload size are rejected and discarded. Payload chunking is not implemented in this alpha version.

# Testing
## Unit testing
Added 3 test units ensuring that calling the new API results in a successful backend operation.
* UserManagerTests: ensure the custom event controller call sendCustomEvent() with the correct name and properties.
* CustomEventBackendServiceTests: ensure the backend service sends the correct data for the testing event
* CustomEventOperationExecutorTests: ensure the executor correctly executes the custom event operation

## Manual testing
- [x] Test case: calling trackEvent with a null properties
```
OneSignal.getUser().trackEvent("empty event", null);
```
* Expect: Event tracked with only the os_sdk in properties

- [x] Test case: calling trackEvent with an empty property map
```
Map<String, Object> empty = new HashMap<>();
OneSignal.getUser().trackEvent("empty event", empty);
```
* Expect: Event tracked with only the os_sdk in properties

- [x] Test case: a simple event property
```
Map<String, Object> simple = Collections.singletonMap("promo_code", "NEW50");
OneSignal.getUser().trackEvent(name, simple);
```
* Expect: Event tracked with the correct properties

- [x] Test case with complex nested properties
```
Map<String, Object> simple = Collections.singletonMap("promo_code", "NEW50");
Map<String, Object> nested = new HashMap<>();
nested.put("signed_up_with_email", true);
nested.put("products_viewed_count", 15);
nested.put("names", Map.of("a", Map.of("b", 5.2)));

Map<String, Object> complex = new HashMap<>();
complex.put("promo_code", "NEW50");
complex.put("additional_data", nested);
OneSignal.getUser().trackEvent(name, complex);
```
  * Expect: Event tracked with all properties under a correct hierarchy
Custom event data recorded on the dashboard:
```
"properties": {
    "additional_data": {
      "signed_up_with_email": true,
      "products_viewed_count": 15,
      "names": {
        "a": {
          "b": 5.2
        }
      }
    },
    "promo_code": "NEW50",
    "os_sdk": {
      "sdk": "050135",
      "device_os": "15",
      "type": "AndroidPush",
      "device_model": "sdk_gphone64_arm64",
      "device_type": "Android",
      "app_version": "1"
    }
  }
```

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2339)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/TrackEventOperation.kt
```diff
@@ -0,0 +1,77 @@
+package com.onesignal.user.internal.operations
+
+import com.onesignal.common.IDManager
+import com.onesignal.core.internal.operations.GroupComparisonType
+import com.onesignal.core.internal.operations.Operation
+import com.onesignal.user.internal.customEvents.impl.CustomEvent
+import com.onesignal.user.internal.operations.impl.executors.CustomEventOperationExecutor
+
+class TrackEventOperation() : Operation(CustomEventOperationExecutor.CUSTOM_EVENT) {
```
nan-li: Recommend to rename this class as TrackCustomEventOperation: **"event"** alone is a loaded term, and we may track other types of events in the future.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/customEvents/impl/CustomEventBackendService.kt
```diff
@@ -0,0 +1,46 @@
...
+        appId: String,
+        onesignalId: String,
+        externalId: String?,
+        timestamp: Long,
+        customEvent: CustomEvent,
+        metadata: CustomEventMetadata,
+    ): ExecutionResponse {
+        val body = JSONObject()
+        body.put("name", customEvent.name)
+        body.put("app_id", appId)
```
nan-li: No app_id in body

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/TrackEventOperation.kt
```diff
@@ -0,0 +1,77 @@
...
+     */
+    var event: CustomEvent
+        get() = getAnyProperty(::event.name) as CustomEvent
+        set(value) {
+            setAnyProperty(::event.name, value)
+        }
+
+    override val createComparisonKey: String get() = "$appId.User.$onesignalId"
+    override val modifyComparisonKey: String get() = "$appId.User.$onesignalId.CustomEvent.$name"
+
```
nan-li: I think this wouldn't be right. All of these comparison keys should be NONE.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/TrackEventOperation.kt
```diff
@@ -0,0 +1,77 @@
...
+    var timeStamp: Long
+        get() = getLongProperty(::timeStamp.name)
+        private set(value) {
+            setLongProperty(::timeStamp.name, value)
+        }
+
+    /**
+     * The custom event instance containing the event name and properties.
+     */
+    var event: CustomEvent
```
nan-li: To make caching easier, I would recommend turning the user's custom event into a jsonobject as early as possible, probably in the custom event controller? Then you probably don't need CustomEvent or ICustomEvent, to simplify the codebase.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/TrackEventOperation.kt
jinliu9508: Addressed in the fixup commit

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/customEvents/impl/CustomEventBackendService.kt
jinliu9508: Addressed in the fixup commit

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/TrackEventOperation.kt
jinliu9508: Addressed in the fixup commit. I use the unique event name for the ID so they don't get batched with other operations that have null keys, in the future.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/TrackEventOperation.kt
jinliu9508: Addressed in the fixup commit. The controller will now accept the map and directly convert it into a JSONObject.

jinliu9508:
Another change is made for using the new endpoint `/apps/:app_id/custom_events`

PR #2341, opened 2025-07-30
fix(subs): Fix parsing subscriptions from server, and improve handling of deleted push subscriptions

nan-li:
# Description
## One Line Summary
When multiple subscriptions are in the create user payload, parse the subscriptions correctly, and this also corrects the handling of deleted push subscriptions.

## Details
The SDK was relying on subscription order in the sent request and response to match subscriptions, but the order is not maintained. We fix this by deprecating `transfer-subscription` in favor of always using `create-subscription`. Create-subscription includes more properties that we can use to parse and match the subscriptions. By using `create-subscription`, this also fixes problems related to push subscriptions deleted and losing the subscription data, now that we always send all the push data. See https://github.com/OneSignal/OneSignal-Android-SDK/pull/2327 for an example.

### Motivation
- Fix bug with wrong subscription ID being used for push
- The fix also improves situations where the push subscription is deleted

### Scope
1. Reverted changes in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2327
2. Change all transfer subscription calls to create subscription calls, which will include all push sub properties instead of just the ID.
3. No longer rely on the ordering of subscriptions in responses. Instead, parse the subscription data to match the local subscription with the remote subscription.
4. If push subscription ID returned by the server is different, the SDK updates its push subscription ID (this is existing behavior).

# Testing
## Unit testing
None added

## Manual testing
Android emulator on API 35

**New install, add email immediately, delete the anon user, login**
1. SDK calls create user for the anon user with push sub and email sub in payload
2. Subscription responses are parsed correctly
3. Delete the user via the dashboard
4. Call login(external_id)
5. SDK calls identify user and gets 404
6. SDK recovers by calling create user with the whole push sub in payload
7. SDK parses subscription response correctly, push sub state is correct

**Current user is anon and has an email sub, delete the push sub, call login to existing user**
1. Current user is anon and has an email sub (prevents the user from being deleted)
2. Delete the push sub via the dashboard
3. Call login(existing_external_id)
4. Identify user returns 409
5. SDK calls create user with the external ID and the whole push sub in the payload
6. The response returns a different push subscription ID and the SDK replaces the existing ID with the new ID correctly
7. SDK state is correct, push sub state is correct

**Current user is anon and has an email sub, delete the push sub, call login to new user**
1. Current user is anon and has an email sub (prevents the user from being deleted)
2. Delete the push sub via the dashboard
3. Call login(brand_new_external_id)
4. Identify user returns success
5. SDK calls create subscription and includes the whole push sub in the payload
6. The response returns a different push subscription ID and the SDK replaces the existing ID with the new ID correctly
7. SDK state is correct, push sub state is correct

**Current user is identified, delete the push sub, then login + add email**
1. SDK has an identified user
2. Delete only the push subscription via the dashboard
3. Login(external_id) and immediately add email
4. SDK calls create user with the whole push sub and email sub in payload
5. The response returns a different push subscription ID and the SDK replaces the existing ID with the new ID correctly
6. SDK state is correct, push sub state is correct

**Current user is identified, delete the push sub, then logout**
1. SDK has an identified user
2. Delete only the push subscription via the dashboard
3. Logout
4. SDK calls create user for the anon user with the whole push sub in the payload
5. The response returns a different push subscription ID and the SDK replaces the existing ID with the new ID correctly
6. SDK state is correct, push sub state is correct

**Delete the push subscription, call optIn/optOut to trigger sub update**
1. Delete the push subscription via the dashboard
2. Call optIn or optOut
3. SDK sends an update subscription call and receives a 404
4. SDK recovers by calling create subscription
5. Receive a new push subscription ID and the SDK replaces the existing ID with the new ID correctly
6. SDK state is correct, push sub state is correct

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2341)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/operations/LoginUserOperationExecutorTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/operations/LoginUserOperationExecutorTests.kt
@@ -739,4 +739,124 @@ class LoginUserOperationExecutorTests : FunSpec({
             exactly = 0,
         ) { mockUserBackendService.createUser(appId, any(), any(), any()) }
     }
+
+    test("create user maps subscriptions when backend order is different (match by id/token)") {
+        // Given
+        val mockUserBackendService = mockk<IUserBackendService>()
+        // backend returns EMAIL first (with token), then PUSH — out of order
+        coEvery { mockUserBackendService.createUser(any(), any(), any(), any()) } returns
+            CreateUserResponse(
+                mapOf(IdentityConstants.ONESIGNAL_ID to remoteOneSignalId),
+                PropertiesObject(),
+                listOf(
+                    SubscriptionObject(id = remoteSubscriptionId2, type = SubscriptionObjectType.EMAIL, token = "name@company.com"),
+                    SubscriptionObject(id = remoteSubscriptionId1, type = SubscriptionObjectType.ANDROID_PUSH, token = "pushToken2"),
+                ),
+            )
+
+        val mockIdentityOperationExecutor = mockk<IdentityOperationExecutor>()
+        val mockIdentityModelStore = MockHelper.identityModelStore()
+        val mockPropertiesModelStore = MockHelper.propertiesModelStore()
+        val mockSubscriptionsModelStore = mockk<SubscriptionModelStore>()
+        every { mockSubscriptionsModelStore.get(any()) } returns null
+
+        val executor =
+            LoginUserOperationExecutor(
+                mockIdentityOperationExecutor,
+                AndroidMockHelper.applicationService(),
+                MockHelper.deviceService(),
+                mockUserBackendService,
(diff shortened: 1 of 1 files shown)


jkasten2 (changes requested):
The code changes themselves look good, but can we add some tests?

jinliu9508:
@jkasten2 I have added two test cases for this change, verifying the mapping of subscriptions is correct and can handle the situation when push sub is deleted.

jkasten2 (approved):

Issue #2215, opened 2024-11-19
[Bug]: NullPointerException in PermissionsActivity.reregisterCallbackHandlers


koral--:
### What happened?

App crashed due to the exception attached in the log.

It seems that the issue is here: https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt#L58
and `extras` turned out to be null.
At the line above the affected one there is a nullable parameter: https://github.com/OneSignal/OneSignal-Android-SDK/blob/main/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt#L57

Regardless of the root cause it does not make sense to have a nullable parameter and then immediately use a non-null assertion operator on it. The parameter should be either non-nullable or there should be a nullcheck.

### Steps to reproduce?

```Markdown
Unknown, it's an automated crash report.
From the stacktrace I guess it is related to app start.
```

### What did you expect to happen?

No crash.

### OneSignal Android SDK version

5.1.23

### Android version

12

### Specific Android models

```Markdown
* Pixel 3a
```

### Relevant log output

```Shell
java.lang.RuntimeException: Unable to start activity ComponentInfo{<stripped app id>/com.onesignal.core.activities.PermissionsActivity}: java.lang.NullPointerException
    android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3635)

(stripped stacktrace entries from android framework)

Caused by: java.lang.NullPointerException: 
    com.onesignal.core.activities.PermissionsActivity.reregisterCallbackHandlers(PermissionsActivity.kt)
    com.onesignal.core.activities.PermissionsActivity.handleBundleParams(PermissionsActivity.kt)
    com.onesignal.core.activities.PermissionsActivity.onCreate(PermissionsActivity.kt:32)
    android.app.Activity.performCreate(Activity.java:8051)

(stripped stacktrace entries from android framework)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

yeldar-nurpeissov:
Hello, is there currently a fix available for this issue?

yeldar-nurpeissov:
Hi, thanks for following up.

This was during closed testing on Google Play Console with community testers to get production access.
On Firebase, I see 3 crashes in one release and 1 crash on the latest release, with 2–3 releases in between those occurrences.
[...]

PR #2349, opened 2025-08-13
fix(crash): NPE in PermissionsActivity

nan-li:
# Description
## One Line Summary
Fix null pointer exception when PermissionsActivity is being created.

## Details
* This should never happen as the SDK does send extras for this Activity, and no other entity should be creating PermissionsActivity.
* However, we have seen crash reports from our users that primarily seems to affect Nexus 5 and Pixel 3, in low numbers.
* One theory is that automatic testing done by the google play store is explicitly driving the PermissionsActivity.

### Motivation
- Address https://github.com/OneSignal/OneSignal-Android-SDK/issues/2215

### Scope
- This appears to be a crash in very low numbers. If we detect `extras` is `null`, then we return early, which is safe to do as it doesn't make sense to show the Activity anyway.

# Testing
## Unit testing
None

## Manual testing
None, unable to repro

**From Josh:**

> The only odd case I can think of is if the end-user backgrounds the app when they see the permission prompt. Then if the app is kicked out of memory and the end-user decides to come back to the app the Activity will be recreated. This is not a flow we normally test, but I just did so now and not seeing any issues.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
@@ -26,6 +26,11 @@ class PermissionsActivity : Activity() {
             return
         }
 
+        if (intent.extras == null) {
+            // This should never happen, but extras is null in rare crash reports
+            return
+        }
+
         requestPermissionService = OneSignal.getService()
         preferenceService = OneSignal.getService()
 


sherwinski (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
```diff
@@ -26,6 +26,11 @@ class PermissionsActivity : Activity() {
             return
         }
 
+        if (intent.extras == null) {
+            // This should never happen, but extras is null in rare crash reports
+            return
```
jinliu9508, line 33: Returning early keeps the Activity alive (it will still hit onStart/onResume), leaves it on the back stack, and can cause a blank/flickery frame. Even though this was most likely triggered by tests passing null extras, we should harden production code.

I would recommend reusing the existing exit path at line 49 so behavior is deterministic.

nan-li: This is because we already called `super.onCreate()` so we need to call `finish()`?
Should the initWithContext check do that as well? I think it is more common for initWithContext to return false, than for the extras to be null?

jinliu9508: In practice, I suspect the initWithContext check never fail: showing the permission prompt requires `OneSignal.getNotification()` or `OneSignal.getLocation()`, and if initWithContext wasn’t invoked earlier, the `“must call initWithContext before use”` exception would already be thrown.

For the case of the automated tests, initializing OneSignal by calling initWithContext in `onCreate()` here seems to work, so was continued to `handleBundleParams()`.

nan-li: Let's call finish on all early returns, updated the PR.

jinliu9508 (approved):

PR #2353, opened 2025-08-22
fix: Finishing the activity on main thread when a notification is opened

abdulraqeeb33:
# Description
## One Line Summary
Crash occurring when a notification was opened. 
### Motivation
The below crash was happening when a notification was received and clicked to open.

```Fatal Exception: java.lang.IllegalArgumentException: reportSizeConfigurations: ActivityRecord not found for: Token{da404ff ActivityRecord{d0ab41e u0 mnn.Android/com.onesignal.notifications.activities.NotificationOpenedActivity t-1 f}}
       at android.os.Parcel.createException(Parcel.java:1970)
       at android.os.Parcel.readException(Parcel.java:1934)
       at android.os.Parcel.readException(Parcel.java:1884)
       at android.app.IActivityManager$Stub$Proxy.reportSizeConfigurations(IActivityManager.java:8625)
       at android.app.ActivityThread.reportSizeConfigurations(ActivityThread.java:3327)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3286)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1970)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7156)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
```
### Scope
The activity was finished by the background thread earlier. So even before all the internal methods of onCreate were actually complete (on main thread), the finish method would be called by the background thread causing a crash.

### OPTIONAL - Other

# Testing
## Unit testing
Added couple of tests

- Testing finish being called on main thread
- Making sure processIntent is called

## Manual testing
Tested opening a notification while the app was foregrounded, app build with Android Studio 2025.1.2 with a fresh install of the OneSignal example app on a emulator running Api 36.0

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [x] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

https://github.com/user-attachments/assets/a59f51f7-9b2a-4817-95d9-191ad9d9c57f

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2353)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211106741376738

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
@@ -72,7 +72,17 @@ fun suspendifyOnThread(
  * call suspending functions.  This is a nonblocking call, which
  * means the scope will run on a background thread.  This will
  * return immediately!!! Also provides an optional onComplete.
- */
+ *
+ * @param priority The priority of the background thread. Default is -1.
+ *                 Higher values indicate higher thread priority.
+ *
+ * @param block A suspending lambda to be executed on the background thread.
+ *              This is where you put your suspending code.
+ *
+ * @param onComplete An optional lambda that will be invoked on the same
+ *                   background thread after [block] has finished executing.
+ *                   Useful for cleanup or follow-up logic.
+ **/
 fun suspendifyOnThread(
     priority: Int = -1,
     block: suspend () -> Unit,


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
```diff
@@ -44,18 +45,26 @@ abstract class NotificationOpenedActivityBase : Activity() {
...
+                // startActivity() must be called BEFORE finish(), otherwise
+                // the app is never foregrounded.
+            },
+            onComplete = {
+                // Safely finish the activity on the main thread after processing is complete.
+                // This gives the system enough time to complete rendering before closing the Trampoline activity.
+                runOnUiThread {
+                    AndroidUtils.finishSafely(this)
+                }
+            },
```
jkasten2, line 67: Why add this new `onComplete` instead of just adding this to the existing `block` above? This adds complexity that doesn't' seems to be needed.

jkasten2 (changes requested):

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
abdulraqeeb33: Thanks for the comment! The onComplete is part of the suspendifyOnThread method to give callers a clean hook after the block completes - so as to run any other task they may want to on whatever thread they prefer. 
It’s only invoked once, immediately after the block finishes, and stays self-contained within the same thread. I don’t believe this adds unnecessary complexity, but happy to clarify or adjust if there’s a preferred pattern.

This also promotes the developer to think of moving code on different threads. In this case the main thread.

jkasten2: Thinking about it again, the pattern of `onComplete` make sense to split up code responsibilities. The `block` being the main work we want to do, and the `onComplete` to "clean up state" that isn't directly related to the code in the `block`.

If the above is correct can you add parameter comments for both  `block` and  `onComplete` parameters in `ThreadUtils.kt` to explain the recommended usage pattern?

abdulraqeeb33: Thank you for looking at it again. I have added comments.

jkasten2 (approved):

PR #2354, opened 2025-08-26
fix(subs): Fix when multiple create subscription operations exist without token

nan-li:
# Description
## One Line Summary
Add the non-local push subscription ID to a create subscription operation, so subscriptions without a push token can all be associated to the same subscription.

### Motivation
After https://github.com/OneSignal/OneSignal-Android-SDK/pull/2341 swapped out all `transfer-subscription` operations with `create-subscription` operations, the latter did not include the existing push subscription ID. With this replacement, we "lose" the subscription ID in the payload. Usually, there is a push token for the backend to fetch the existing push sub, but if the SDK doesn't have a token yet, there is no identifier. This causes a problem if the SDK does not have a push token yet and multiple `create-subscriptions` are sent with neither ID nor token for the backend to identify the existing subscription to use. Since each create-subscription did not include a token nor an ID, the backend returned a brand new subscription each time. Now, the create-subscription will include the push subscription ID if it is non-local so the backend can look up the subscription properly.

### Scope
- Create subscription operations now include the push subscription ID, if  non-local.
- It already includes all other properties such as notification_types, token, enabled, etc.
- If this push subscription has since been deleted, a new subscription ID will be returned, and the SDK hydrates the new ID properly.

# Testing
## Unit testing
Added a test to differentiate when local vs valid push subscription exists, for create subscription.

## Manual testing
Pixel Emulator API 35
- reproduced issue by turning off wifi on a new install, and calling multiple logins (no push token is available yet)
- after PR, this issue is fixed
- also tested what happens if the push subscription has been deleted on the server, and SDK handles properly

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2354)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/CreateSubscriptionOperation.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/CreateSubscriptionOperation.kt
@@ -99,11 +99,7 @@ class CreateSubscriptionOperation() : Operation(SubscriptionOperationExecutor.CR
     }
 
     override fun translateIds(map: Map<String, String>) {
-        if (map.containsKey(onesignalId)) {
-            onesignalId = map[onesignalId]!!
-        }
-        if (map.containsKey(subscriptionId)) {
-            subscriptionId = map[subscriptionId]!!
-        }
+        map[onesignalId]?.let { onesignalId = it }
+        map[subscriptionId]?.let { subscriptionId = it }
     }
 }


jinliu9508 (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/CreateSubscriptionOperation.kt
```diff
@@ -102,5 +102,8 @@ class CreateSubscriptionOperation() : Operation(SubscriptionOperationExecutor.CR
         if (map.containsKey(onesignalId)) {
             onesignalId = map[onesignalId]!!
         }
+        if (map.containsKey(subscriptionId)) {
+            subscriptionId = map[subscriptionId]!!
```
abdulraqeeb33: why do we need the !!? are we just following the pattern above?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
```diff
@@ -87,11 +88,12 @@ internal class SubscriptionOperationExecutor(
         val enabled = lastUpdateOperation?.enabled ?: createOperation.enabled
         val address = lastUpdateOperation?.address ?: createOperation.address
         val status = lastUpdateOperation?.status ?: createOperation.status
+        val subId = if (!IDManager.isLocalId(createOperation.subscriptionId)) createOperation.subscriptionId else null
```
abdulraqeeb33, line 91: I think we should write a unit test for this to make sure we are checking against the IDManager.isLocalId or atleast include testing this in the current unit tests.

abdulraqeeb33 (approved):
couple of nits but otherwise looks good.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
nan-li: Do you mean as a finer unit test for just this piece of functionality? I semi incorporated this in the `SubscriptionOperationExecutorTests.kt` changes. 
- test("create subscription successfully creates subscription, with local ID") was adjusted to add a check that if the ID is local, it eventually goes to the backend executor as null.
- test("create subscription includes the subscription ID if non-local") was added to test the behavior when ID is non-local

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/CreateSubscriptionOperation.kt
nan-li: Even though we check `map.containsKey`, my IDE at least can't pick up that `map[subscriptionId]` is a `String`, not a `String?`. Removing the `!!` raises a Type mismatch error: Required: String, Found: String?, and then prompts me to add non-null assertion.

abdulraqeeb33: Right, thats because its possible that the value in map[subscriptionId] could be null.
Kotlin provides a safe way to refer such values which may be null. https://kotlinlang.org/docs/null-safety.html# here by using the ? and let, we make sure that a Null pointer exception will never occur here.


```suggestion
            if (map.containsKey(subscriptionId)) {
            map[subscriptionId]?.let { subscriptionId = it }
        }
```

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
abdulraqeeb33: You are right. I should have taken a closer look at it. Thanks.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/CreateSubscriptionOperation.kt
```diff
@@ -99,8 +99,7 @@ class CreateSubscriptionOperation() : Operation(SubscriptionOperationExecutor.CR
     }
 
     override fun translateIds(map: Map<String, String>) {
-        if (map.containsKey(onesignalId)) {
-            onesignalId = map[onesignalId]!!
-        }
+        map[onesignalId]?.let { onesignalId = it }
```
abdulraqeeb33, line 102: 💯

PR #2358, opened 2025-09-02
Automate Version Number

abdulraqeeb33:
# Description
## One Line Summary
This PR is centralizing the storage of version number for the SDK. Later that is updated from the Github action.

### Motivation
Part 1 of automation

### Scope
Automating the release, should not effect the normal running of the app.

# Testing
## Unit testing
Added a bunch of unit tests.

## Manual testing
Tested the change by running the Example app and logged the version number.

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install GitHub CLI
+        run: sudo apt-get install -y gh
+
+      - name: Authenticate GitHub CLI
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | gh auth login --with-token
+
+      - name: Create release branch from main (delete it exists)
```
sherwinski:
```suggestion
      - name: Create release branch from main (delete it first, if it already exists)
```

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,82 @@
...
+          sed -i.bak "s/^SDK_VERSION=.*/SDK_VERSION=$VERSION/" Examples/OneSignalDemo/gradle.properties
+          rm Examples/OneSignalDemo/gradle.properties.bak
+          
+      - name: Configure Git and commit changes
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
+          BRANCH="rel/$VERSION"
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
sherwinski, line 49: Can L68 be excluded?

sherwinski (commented):
From a GitHub actions viewpoint, this looks good to me. I will let others comment on the version number automation part.

Only a few thoughts come to mind:
- Making sure it's clear that when kicking off the action, the version number is not prepended with `v`. I only mention this because it has been an issue in other repos and cause problems with version ordering.
- Does maven central reject releases with the same version number? If not, it may be helpful to add guardrails

.github/workflows/create-release-pr.yml
abdulraqeeb33: its better to keep this since this is linked to the github-actions[bot], it will display a clean commit.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install GitHub CLI
+        run: sudo apt-get install -y gh
```
jkasten2: This should be installed by default for Github runners.
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+
+      - name: Authenticate GitHub CLI
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | gh auth login --with-token
+
+      - name: Create release branch from main (delete it first, it it already exists)
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
+          BRANCH="rel/$VERSION"
+
+          git fetch origin
```
jkasten2: Doesn't `actions/checkout@v3` already do this?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+          
+          # Delete local branch if it exists
+          if git show-ref --quiet refs/heads/"$BRANCH"; then
+            echo "Deleting local branch $BRANCH"
+            git branch -D "$BRANCH"
+          fi
```
jkasten2: There shouldn't be any local branches as the Github action will always be run from a clean clone of the repo.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+            echo "Deleting remote branch $BRANCH"
+            git push origin --delete "$BRANCH"
+          fi
+
+          # Create branch fresh from origin/main
+          git checkout -b "$BRANCH" origin/main
+
+      - name: Update gradle.properties files with SDK_VERSION
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
```
jkasten2: VERSION is set in the previous step, do we need to repeat this here?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+          sed -i.bak "s/^SDK_VERSION=.*/SDK_VERSION=$VERSION/" OneSignalSDK/gradle.properties
+          rm OneSignalSDK/gradle.properties.bak
+          
+          sed -i.bak "s/^SDK_VERSION=.*/SDK_VERSION=$VERSION/" Examples/OneSignalDemo/gradle.properties
+          rm Examples/OneSignalDemo/gradle.properties.bak
+          
+      - name: Configure Git and commit changes
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
+          BRANCH="rel/$VERSION"
```
jkasten2: VERSION and BRANCH are set in the previous step, do we need to repeat this here?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+          git commit -am "chore: bump SDK_VERSION to $VERSION"
+          git push origin "$BRANCH"
+          
+      - name: Generate Release Notes from Commits
+        id: generate_notes
+        run: |
+          echo "## 🔖 Auto-Generated Release Notes" > pr_body.md
+          echo "" >> pr_body.md
+
+          VERSION="${{ github.event.inputs.version }}"
```
jkasten2: VERSION is set in the previous step, do we need to repeat this here?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+          if [[ -z "$COMMITS" ]]; then
+            echo "### ⚠️ No commits found" >> pr_body.md
+            echo "_No commits were found since the last release merge._" >> pr_body.md
+            echo "Release manager to verify and add appropriate release notes." >> pr_body.md
+          fi
+
+      - name: Create Pull Request
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
+          BRANCH="rel/$VERSION"
```
jkasten2: VERSION and BRANCH are set in the previous step, do we need to repeat this here?

OneSignalSDK/settings.gradle
```diff
@@ -3,7 +3,7 @@
 gradle.rootProject {
     allprojects {
         group = 'com.onesignal'
-        version = '5.1.37'
+        version = gradle.rootProject.hasProperty("SDK_VERSION") ? gradle.rootProject.getProperty("SDK_VERSION") : "x.x.x"
```
jkasten2: Why have a fallback to "x.x.x" here? If `SDK_VERSION` can't be found then it would be better to fail the build then silently move on.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/OneSignalUtils.kt
```diff
@@ -1,12 +1,28 @@
...
+     * User-Agent strings and other uses where lexicographical order matters.
+     * Its being picked up from the Gradle build config field `SDK_VERSION`.
+     * Also its calculated lazily once when first accessed and then cached after that.
      */
-    const val SDK_VERSION: String = "050137"
+    val sdkVersion: String by lazy {
+        formatVersion(BuildConfig.SDK_VERSION)
+    }
+
+    internal fun formatVersion(version: String): String {
```
jkasten2: nit: Add a short description of the format conversion of the version number happing here.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+          echo "### 🐛 Bug Fixes" >> pr_body.md
+          grep -E '^bug' all_commits.txt | sed 's/^/- /' >> pr_body.md || echo "- _None_" >> pr_body.md
+
+          echo "" >> pr_body.md
+          echo "### 🔧 Improvements" >> pr_body.md
+          grep -E '^(perf|refactor)' all_commits.txt | sed 's/^/- /' >> pr_body.md || echo "- _None_" >> pr_body.md
+
+          echo "" >> pr_body.md
+          echo "### 📝 Uncategorized Commits" >> pr_body.md
+          grep -v -E '^(feat|bug|perf|refactor)' all_commits.txt | sed 's/^/- /' >> pr_body.md || echo "- _None_" >> pr_body.md
```
jkasten2: Can we use the PR title instead of each commit for the release notes? There can be multiple commits per PR and commit titles are normally more technical and granular, making it hard to understand what changed at a high level.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,143 @@
...
+          echo "$VERSION" >> pr_body.md
+          
+          echo "" >> pr_body.md
+
+          # Final fallback: check if any commits were found at all
+          if [[ -z "$COMMITS" ]]; then
+            echo "### ⚠️ No commits found" >> pr_body.md
+            echo "_No commits were found since the last release merge._" >> pr_body.md
+            echo "Release manager to verify and add appropriate release notes." >> pr_body.md
+          fi
```
jkasten2: If there are no commits there is nothing to do, shouldn't we stop here and fail the Github action?

.github/workflows/create-release-pr.yml
abdulraqeeb33: yeah i think i was just being more safe. but makes sense

.github/workflows/create-release-pr.yml
abdulraqeeb33: so it basically gets the current commit of the [default branch](https://github.com/actions/checkout), but since we need full git history to gather the logs, we needed to do a fetch origin. The alternative is to 

```with:
    fetch-depth: 0 
    fetch-tags: true```

File: .github/workflows/create-release-pr.yml (hunk 13 above)
Comment (abdulraqeeb33) [reply .github/workflows/create-release-pr.yml]: yes you are right.

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/OneSignalUtils.kt (hunk 11 above)
Comment (jkasten2) [reply OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/OneSignalUtils.kt]: I see this was resolved in another PR:
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2362/files#diff-d87b0e41bc1429f5344d3886c20a3d85366ae4383e03f2054d0488aeb4747f86R18-R20

File: .github/workflows/create-release-pr.yml [hunk 14]
```diff
@@ -0,0 +1,147 @@
...
+
+          git checkout -b "$BRANCH" origin/main
+
+      - name: Update SDK_VERSION in gradle.properties
+        run: |
+          sed -i.bak "s/^SDK_VERSION=.*/SDK_VERSION=$VERSION/" OneSignalSDK/gradle.properties
+          rm OneSignalSDK/gradle.properties.bak
+          
+          sed -i.bak "s/^SDK_VERSION=.*/SDK_VERSION=$VERSION/" Examples/OneSignalDemo/gradle.properties
+          rm Examples/OneSignalDemo/gradle.properties.bak
```
jkasten2: Can we use `sed` to modify the file without creating a temp file and having to clean it up with `rm`?

PR #2359, opened 2025-09-02
ci: publish job

abdulraqeeb33:
# Description
## One Line Summary
Create and publish the latest release from the main branch.

## Details
We want this GitHub Action (GA) to do the following

- Checkout the branch (dropdown). Manually enter the release tag you want to publish, if empty - pick up from gradle.properties
- Assemble
- Publish to Maven
- Create a release tag and push to Github

image

image

### Motivation
Ease of deploying new version of a release.

### Scope
Updated the release flow.

# Testing
## Unit testing
No unit tests as its purely a GA

## Manual testing
Tested by running the jobs, attached pics for reference.

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2359)


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,150 @@
...
+          gpg --list-packets "$GPG_FILE_PATH" || echo "Invalid key file!"
+
+      - name: Resolve release version
+        id: resolve_version
+        run: |
+          if [[ -n "${{ github.event.inputs.version }}" ]]; then
+            VERSION="${{ github.event.inputs.version }}"
+          else
+            VERSION=$(grep '^SDK_VERSION=' OneSignalSDK/gradle.properties | cut -d '=' -f2)
+          fi
```
jkasten2: It looks like setting `github.event.inputs.version` will result in the version value in `gradle.properties` not matching. Is there a case you are thinking of where this would be valid?

abdulraqeeb33: this is for a manual run. We could technically get rid of it. I added it just as a option to run manually.
Do you think its worth keeping it?

jkasten2: I think setting it is currently misleading, as it will result in the version number that is published being different that what is shown on Github. Given that, I would rather not have this option, instead a new release PR should be made if the version really does need to change.

abdulraqeeb33: makes sense. i will get rid of it.

PR #2368, opened 2025-09-11
Refactor: move part of initWithContext off main thread

jinliu9508:
# Description
## One Line Summary
Move slow/IO-bound parts of initWithContext off the main thread and add an async overload with a completion callback for internal usage. 

### Motivation
Avoid main-thread stalls and potential ANRs during initialization, especially when SharedPreferences or other disk/network operations are involved. Provide an async path so app components (Activities/Receivers) can trigger initialization without blocking UI or risking Receiver timeouts.

### Scope
* initWithContext(contex, appId): now move all initialization to background and return immediately. 
* initWithContext(contex) is now suspend and can only be called from a coroutine. This method should used by internal classes only. It will remain suspend until all initialization is completed.
* Accessors remain blocked if called immediately after initWithContext and before initialization is fully completed. This is to ensure no breaking change or expectation on existing consumer use cases.
* Receivers will now call initWithContext(contex) in background, and switch back to main on tasks that must be executed in main thread, eg, finish(), startActivity() and other app lifecycle or UI operations. 
* Workers run doWork() in background will now directly call initWithContext(contex)
* Retain the existing synchronous overload for backward compatibility; the new overload provides a non-blocking path with a completion callback.
* Introduces a centralized ANR timeout constant. 
* No behavioral change to notification delivery formats, APIs, or user-visible features.

# Testing
## Unit testing
* New Robolectric/Kotest tests cover initWithContext in different scenarios. Ex:
  * Accessors throw before initialization.
  * Init does not block even when SharedPreferences access is artificially delayed.
  * Tags, local onesignalId, push subscription, and immediate login right after init. 
* LatchAwaiterTests testing the genetic functionalities of the waiter.

## Manual testing
* Cold-start the sample app with logging at VERBOSE. Ensure all initial operations are created in the same order.
* Trigger init from an Activity and verify UI remains responsive.
* For Receivers: simulate a boot or FCM broadcast; verify the receiver returns promptly (no >10s wait), and OneSignal completes initialization shortly after.
* Notification click when app is killed / backgrounded. 
* Login, logout, addTag, addTrigger, permission update, IAM interaction
* Environment: Android Studio Hedgehog or later; Pixel 6 emulator API 33; fresh install of the demo app.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2368)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
@@ -85,8 +85,15 @@ interface IOneSignal {
      */
     fun initWithContext(
         context: Context,
-        appId: String?,
+        appId: String,
     ): Boolean
+
+    /**
+     * Initialize the OneSignal SDK, suspend until initialization is completed
+     *
+     * @param context The Android context the SDK should use.
+     */
+    suspend fun initWithContext(context: Context): Boolean
 
     /**
      * Login to OneSignal under the user identified by the [externalId] provided. The act of
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
@@ -204,8 +204,8 @@ object OneSignal {
      * THIS IS AN INTERNAL INTERFACE AND SHOULD NOT BE USED DIRECTLY.
      */
     @JvmStatic
-    fun initWithContext(context: Context): Boolean {
-        return oneSignal.initWithContext(context, null)
+    suspend fun initWithContext(context: Context): Boolean {
+        return oneSignal.initWithContext(context)
     }
 
     /**
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
@@ -1,0 +1,82 @@
+package com.onesignal.common.threading
+
+import com.onesignal.common.AndroidUtils
+import com.onesignal.debug.internal.logging.Logging
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+/**
+ * This class allows blocking execution until asynchronous initialization or completion is signaled, with support for configurable timeouts and detailed logging for troubleshooting.
+ * It is designed for scenarios where certain tasks, such as SDK initialization, must finish before continuing.
+ * When used on the main/UI thread, it applies a shorter timeout and logs a thread stack trace to warn developers, helping to prevent Application Not Responding (ANR) errors caused by blocking the UI thread.
+ *
+ * Usage:
+ *   val awaiter = LatchAwaiter("OneSignal SDK Init")
+ *   awaiter.release() // when done
+ */
+class LatchAwaiter(
+    private val componentName: String = "Component",
+) {
+    companion object {
+        const val DEFAULT_TIMEOUT_MS = 30_000L // 30 seconds
+        const val ANDROID_ANR_TIMEOUT_MS = 4_800L // Conservative ANR threshold
+    }
+
+    private val latch = CountDownLatch(1)
+
+    /**
+     * Releases the latch to unblock any waiting threads.
+     */
+    fun release() {
+        latch.countDown()
+    }
+
+    /**
+     * Wait for the latch to be released with an optional timeout.
+     *
+     * @return true if latch was released before timeout, false otherwise.
+     */
+    fun await(timeoutMs: Long = getDefaultTimeout()): Boolean {
+        val completed =
+            try {
+                latch.await(timeoutMs, TimeUnit.MILLISECONDS)
+            } catch (e: InterruptedException) {
+                Logging.warn("Interrupted while waiting for $componentName", e)
+                logAllThreads()
+                false
+            }
+
+        if (!completed) {
+            val message = createTimeoutMessage(timeoutMs)
+            Logging.warn(message)
+        }
+
+        return completed
+    }
+
+    private fun getDefaultTimeout(): Long {
+        return if (AndroidUtils.isRunningOnMainThread()) ANDROID_ANR_TIMEOUT_MS else DEFAULT_TIMEOUT_MS
+    }
+
+    private fun createTimeoutMessage(timeoutMs: Long): String {
+        return if (AndroidUtils.isRunningOnMainThread()) {
+            "Timeout waiting for $componentName after ${timeoutMs}ms on the main thread. " +
+                "This can cause ANRs. Consider calling from a background thread."
+        } else {
+            "Timeout waiting for $componentName after ${timeoutMs}ms."
+        }
+    }
+
+    private fun logAllThreads(): String {
+        val allThreads = Thread.getAllStackTraces()
+        val sb = StringBuilder()
+        for ((thread, stack) in allThreads) {
+            sb.append("ThreadDump Thread: ${thread.name} [${thread.state}]\n")
+            for (element in stack) {
+                sb.append("\tat $element\n")
+            }
+        }
+
+        return sb.toString()
+    }
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
@@ -8,11 +8,14 @@ import android.os.Bundle
 import android.os.Handler
 import androidx.core.app.ActivityCompat
 import com.onesignal.OneSignal
+import com.onesignal.common.threading.suspendifyOnThread
 import com.onesignal.core.R
 import com.onesignal.core.internal.permissions.impl.RequestPermissionService
 import com.onesignal.core.internal.preferences.IPreferencesService
 import com.onesignal.core.internal.preferences.PreferenceOneSignalKeys
 import com.onesignal.core.internal.preferences.PreferenceStores
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
 class PermissionsActivity : Activity() {
     private var requestPermissionService: RequestPermissionService? = null
@@ -22,21 +25,29 @@ class PermissionsActivity : Activity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        if (!OneSignal.initWithContext(this)) {
-            finishActivity()
-            return
-        }
-
         if (intent.extras == null) {
             // This should never happen, but extras is null in rare crash reports
             finishActivity()
             return
         }
 
-        requestPermissionService = OneSignal.getService()
-        preferenceService = OneSignal.getService()
+        // init in background
+        suspendifyOnThread {
+            val initialized = OneSignal.initWithContext(this)
 
-        handleBundleParams(intent.extras)
+            // finishActivity() and handleBundleParams must be called from main
+            withContext(Dispatchers.Main) {
+                if (!initialized) {
+                    finishActivity()
+                    return@withContext
+                }
+
+                requestPermissionService = OneSignal.getService()
+                preferenceService = OneSignal.getService()
+
+                handleBundleParams(intent.extras)
+            }
+        }
     }
 
     override fun onNewIntent(intent: Intent) {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
@@ -35,13 +35,14 @@ import com.onesignal.debug.internal.logging.Logging
 
 class SyncJobService : JobService() {
     override fun onStartJob(jobParameters: JobParameters): Boolean {
-        if (!OneSignal.initWithContext(this)) {
-            return false
-        }
-
-        var backgroundService = OneSignal.getService<IBackgroundManager>()
-
         suspendifyOnThread {
+            // init OneSignal in background
+            if (!OneSignal.initWithContext(this)) {
+                jobFinished(jobParameters, false)
+                return@suspendifyOnThread
+            }
+
+            val backgroundService = OneSignal.getService<IBackgroundManager>()
             backgroundService.runBackgroundServices()
 
             Logging.debug("LollipopSyncRunnable:JobFinished needsJobReschedule: " + backgroundService.needsJobReschedule)
@@ -49,7 +50,6 @@ class SyncJobService : JobService() {
             // Reschedule if needed
             val reschedule = backgroundService.needsJobReschedule
             backgroundService.needsJobReschedule = false
-
             jobFinished(jobParameters, reschedule)
         }
 
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -14,6 +14,7 @@ import com.onesignal.common.safeString
 import com.onesignal.common.services.IServiceProvider
 import com.onesignal.common.services.ServiceBuilder
 import com.onesignal.common.services.ServiceProvider
+import com.onesignal.common.threading.LatchAwaiter
 import com.onesignal.common.threading.suspendifyOnThread
 import com.onesignal.core.CoreModule
 import com.onesignal.core.internal.application.IApplicationService
@@ -53,8 +54,16 @@ import com.onesignal.user.internal.subscriptions.SubscriptionType
 import org.json.JSONObject
 
 internal class OneSignalImp : IOneSignal, IServiceProvider {
+    @Volatile
+    private var latchAwaiter = LatchAwaiter("OneSignalImp")
+
+    @Volatile
+    private var initState: InitState = InitState.NOT_STARTED
+
     override val sdkVersion: String = OneSignalUtils.sdkVersion
-    override var isInitialized: Boolean = false
+
+    override val isInitialized: Boolean
+        get() = initState == InitState.SUCCESS
 
     override var consentRequired: Boolean
         get() = configModel?.consentRequired ?: (_consentRequired == true)
@@ -83,46 +92,25 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
 
     // we hardcode the DebugManager implementation so it can be used prior to calling `initWithContext`
     override val debug: IDebugManager = DebugManager()
-    override val session: ISessionManager get() =
-        if (isInitialized) {
-            services.getService()
-        } else {
-            throw Exception(
-                "Must call 'initWithContext' before use",
-            )
-        }
-    override val notifications: INotificationsManager get() =
-        if (isInitialized) {
-            services.getService()
-        } else {
-            throw Exception(
-                "Must call 'initWithContext' before use",
-            )
-        }
-    override val location: ILocationManager get() =
-        if (isInitialized) {
-            services.getService()
-        } else {
-            throw Exception(
-                "Must call 'initWithContext' before use",
-            )
-        }
-    override val inAppMessages: IInAppMessagesManager get() =
-        if (isInitialized) {
-            services.getService()
-        } else {
-            throw Exception(
-                "Must call 'initWithContext' before use",
-            )
-        }
-    override val user: IUserManager get() =
-        if (isInitialized) {
-            services.getService()
-        } else {
-            throw Exception(
-                "Must call 'initWithContext' before use",
-            )
-        }
+    override val session: ISessionManager
+        get() =
+            waitAndReturn { services.getService() }
+
+    override val notifications: INotificationsManager
+        get() =
+            waitAndReturn { services.getService() }
+
+    override val location: ILocationManager
+        get() =
+            waitAndReturn { services.getService() }
+
+    override val inAppMessages: IInAppMessagesManager
+        get() =
+            waitAndReturn { services.getService() }
+
+    override val user: IUserManager
+        get() =
+            waitAndReturn { services.getService() }
 
     // Services required by this class
     // WARNING: OperationRepo depends on OperationModelStore which in-turn depends
@@ -179,169 +167,218 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
         services = serviceBuilder.build()
     }
 
+    private fun initEssentials(context: Context) {
+        PreferenceStoreFix.ensureNoObfuscatedPrefStore(context)
+
+        // start the application service. This is called explicitly first because we want
+        // to make sure it has the context provided on input, for all other startable services
+        // to depend on if needed.
+        val applicationService = services.getService<IApplicationService>()
+        (applicationService as ApplicationService).start(context)
+
+        // Give the logging singleton access to the application service to support visual logging.
+        Logging.applicationService = applicationService
+
+        // get the current config model, if there is one
+        configModel = services.getService<ConfigModelStore>().model
+    }
+
+    private fun updateConfig() {
+        // if requires privacy consent was set prior to init, set it in the model now
+        if (_consentRequired != null) {
+            configModel!!.consentRequired = _consentRequired!!
+        }
+
+        // if privacy consent was set prior to init, set it in the model now
+        if (_consentGiven != null) {
+            configModel!!.consentGiven = _consentGiven!!
+        }
+
+        if (_disableGMSMissingPrompt != null) {
+            configModel!!.disableGMSMissingPrompt = _disableGMSMissingPrompt!!
+        }
+    }
+
+    private fun bootstrapServices(): StartupService {
+        sessionModel = services.getService<SessionModelStore>().model
+        operationRepo = services.getService<IOperationRepo>()
+
+        val startupService = StartupService(services)
+        // bootstrap all services
+        startupService.bootstrap()
+
+        return startupService
+    }
+
+    private fun initUser(forceCreateUser: Boolean) {
+        // create a new local user
+        if (forceCreateUser ||
+            !identityModelStore!!.model.hasProperty(IdentityConstants.ONESIGNAL_ID)
+        ) {
+            val legacyPlayerId =
+                preferencesService!!.getString(
+                    PreferenceStores.ONESIGNAL,
+                    PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
+                )
+            if (legacyPlayerId == null) {
+                Logging.debug("initWithContext: creating new device-scoped user")
+                createAndSwitchToNewUser()
+                operationRepo!!.enqueue(
+                    LoginUserOperation(
+                        configModel!!.appId,
+                        identityModelStore!!.model.onesignalId,
+                        identityModelStore!!.model.externalId,
+                    ),
+                )
+            } else {
+                Logging.debug("initWithContext: creating user linked to subscription $legacyPlayerId")
+
+                // Converting a 4.x SDK to the 5.x SDK.  We pull the legacy user sync values to create the subscription model, then enqueue
+                // a specialized `LoginUserFromSubscriptionOperation`, which will drive fetching/refreshing of the local user
+                // based on the subscription ID we do have.
+                val legacyUserSyncString =
+                    preferencesService!!.getString(
+                        PreferenceStores.ONESIGNAL,
+                        PreferenceOneSignalKeys.PREFS_LEGACY_USER_SYNCVALUES,
+                    )
+                var suppressBackendOperation = false
+
+                if (legacyUserSyncString != null) {
+                    val legacyUserSyncJSON = JSONObject(legacyUserSyncString)
+                    val notificationTypes =
+                        legacyUserSyncJSON.safeInt("notification_types")
+
+                    val pushSubscriptionModel = SubscriptionModel()
+                    pushSubscriptionModel.id = legacyPlayerId
+                    pushSubscriptionModel.type = SubscriptionType.PUSH
+                    pushSubscriptionModel.optedIn =
+                        notificationTypes != SubscriptionStatus.NO_PERMISSION.value && notificationTypes != SubscriptionStatus.UNSUBSCRIBE.value
+                    pushSubscriptionModel.address =
+                        legacyUserSyncJSON.safeString("identifier") ?: ""
+                    if (notificationTypes != null) {
+                        pushSubscriptionModel.status =
+                            SubscriptionStatus.fromInt(notificationTypes)
+                                ?: SubscriptionStatus.NO_PERMISSION
+                    } else {
+                        pushSubscriptionModel.status = SubscriptionStatus.SUBSCRIBED
+                    }
+
+                    pushSubscriptionModel.sdk = OneSignalUtils.sdkVersion
+                    pushSubscriptionModel.deviceOS = Build.VERSION.RELEASE
+                    pushSubscriptionModel.carrier = DeviceUtils.getCarrierName(
+                        services.getService<IApplicationService>().appContext,
+                    ) ?: ""
+                    pushSubscriptionModel.appVersion = AndroidUtils.getAppVersion(
+                        services.getService<IApplicationService>().appContext,
+                    ) ?: ""
+
+                    configModel!!.pushSubscriptionId = legacyPlayerId
+                    subscriptionModelStore!!.add(
+                        pushSubscriptionModel,
+                        ModelChangeTags.NO_PROPOGATE,
+                    )
+                    suppressBackendOperation = true
+                }
+
+                createAndSwitchToNewUser(suppressBackendOperation = suppressBackendOperation)
+
+                operationRepo!!.enqueue(
+                    LoginUserFromSubscriptionOperation(
+                        configModel!!.appId,
+                        identityModelStore!!.model.onesignalId,
+                        legacyPlayerId,
+                    ),
+                )
+                preferencesService!!.saveString(
+                    PreferenceStores.ONESIGNAL,
+                    PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
+                    null,
+                )
+            }
+        } else {
+            Logging.debug("initWithContext: using cached user ${identityModelStore!!.model.onesignalId}")
+        }
+    }
+
     override fun initWithContext(
         context: Context,
-        appId: String?,
+        appId: String,
     ): Boolean {
         Logging.log(LogLevel.DEBUG, "initWithContext(context: $context, appId: $appId)")
 
+        // do not do this again if already initialized or init is in progress
         synchronized(initLock) {
-            // do not do this again if already initialized
-            if (isInitialized) {
-                Logging.log(LogLevel.DEBUG, "initWithContext: SDK already initialized")
+            if (initState.isSDKAccessible()) {
+                Logging.log(LogLevel.DEBUG, "initWithContext: SDK already initialized or in progress")
                 return true
             }
 
-            Logging.log(LogLevel.DEBUG, "initWithContext: SDK initializing")
+            initState = InitState.IN_PROGRESS
+        }
 
-            PreferenceStoreFix.ensureNoObfuscatedPrefStore(context)
+        // init in background and return immediately to ensure non-blocking
+        suspendifyOnThread {
+            internalInit(context, appId)
+        }
+        initState = InitState.SUCCESS
+        return true
+    }
 
-            // start the application service. This is called explicitly first because we want
-            // to make sure it has the context provided on input, for all other startable services
-            // to depend on if needed.
-            val applicationService = services.getService<IApplicationService>()
-            (applicationService as ApplicationService).start(context)
+    /**
+     * Called from internal classes only. Remain suspend until initialization is fully completed.
+     */
+    override suspend fun initWithContext(context: Context): Boolean {
+        Logging.log(LogLevel.DEBUG, "initWithContext(context: $context)")
 
-            // Give the logging singleton access to the application service to support visual logging.
-            Logging.applicationService = applicationService
+        // do not do this again if already initialized or init is in progress
+        synchronized(initLock) {
+            if (initState.isSDKAccessible()) {
+                Logging.log(LogLevel.DEBUG, "initWithContext: SDK already initialized or in progress")
+                return true
+            }
 
-            // get the current config model, if there is one
-            configModel = services.getService<ConfigModelStore>().model
-            sessionModel = services.getService<SessionModelStore>().model
-            operationRepo = services.getService<IOperationRepo>()
+            initState = InitState.IN_PROGRESS
+        }
 
-            var forceCreateUser = false
+        val result = internalInit(context, null)
+        initState = if (result) InitState.SUCCESS else InitState.FAILED
+        return result
+    }
 
-            // initWithContext is called by our internal services/receivers/activities but they do not provide
-            // an appId (they don't know it).  If the app has never called the external initWithContext
-            // prior to our services/receivers/activities we will blow up, as no appId has been established.
-            if (appId == null && !configModel!!.hasProperty(ConfigModel::appId.name)) {
+    private fun internalInit(
+        context: Context,
+        appId: String?,
+    ): Boolean {
+        initEssentials(context)
+
+        var forceCreateUser = false
+        if (appId != null) {
+            // If new appId is different from stored one, flag user recreation
+            if (!configModel!!.hasProperty(ConfigModel::appId.name) || configModel!!.appId != appId) {
+                forceCreateUser = true
+            }
+            configModel!!.appId = appId
+        } else {
+            // appId is null — fallback to legacy
+            if (!configModel!!.hasProperty(ConfigModel::appId.name)) {
                 val legacyAppId = getLegacyAppId()
                 if (legacyAppId == null) {
-                    Logging.warn("initWithContext called without providing appId, and no appId has been established!")
+                    Logging.warn("suspendInitInternal: no appId provided or found in legacy config.")
+                    initState = InitState.FAILED
+                    latchAwaiter.release()
                     return false
-                } else {
-                    Logging.debug("initWithContext: using cached legacy appId $legacyAppId")
-                    forceCreateUser = true
-                    configModel!!.appId = legacyAppId
                 }
+                forceCreateUser = true
+                configModel!!.appId = legacyAppId
             }
-
-            // if the app id was specified as input, update the config model with it
-            if (appId != null) {
-                if (!configModel!!.hasProperty(ConfigModel::appId.name) || configModel!!.appId != appId) {
-                    forceCreateUser = true
-                }
-                configModel!!.appId = appId
-            }
-
-            // if requires privacy consent was set prior to init, set it in the model now
-            if (_consentRequired != null) {
-                configModel!!.consentRequired = _consentRequired!!
-            }
-
-            // if privacy consent was set prior to init, set it in the model now
-            if (_consentGiven != null) {
-                configModel!!.consentGiven = _consentGiven!!
-            }
-
-            if (_disableGMSMissingPrompt != null) {
-                configModel!!.disableGMSMissingPrompt = _disableGMSMissingPrompt!!
-            }
-
-            val startupService = StartupService(services)
-
-            // bootstrap services
-            startupService.bootstrap()
-
-            if (forceCreateUser || !identityModelStore!!.model.hasProperty(IdentityConstants.ONESIGNAL_ID)) {
-                val legacyPlayerId =
-                    preferencesService!!.getString(
-                        PreferenceStores.ONESIGNAL,
-                        PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
-                    )
-                if (legacyPlayerId == null) {
-                    Logging.debug("initWithContext: creating new device-scoped user")
-                    createAndSwitchToNewUser()
-                    operationRepo!!.enqueue(
-                        LoginUserOperation(
-                            configModel!!.appId,
-                            identityModelStore!!.model.onesignalId,
-                            identityModelStore!!.model.externalId,
-                        ),
-                    )
-                } else {
-                    Logging.debug("initWithContext: creating user linked to subscription $legacyPlayerId")
-
-                    // Converting a 4.x SDK to the 5.x SDK.  We pull the legacy user sync values to create the subscription model, then enqueue
-                    // a specialized `LoginUserFromSubscriptionOperation`, which will drive fetching/refreshing of the local user
-                    // based on the subscription ID we do have.
-                    val legacyUserSyncString =
-                        preferencesService!!.getString(
-                            PreferenceStores.ONESIGNAL,
-                            PreferenceOneSignalKeys.PREFS_LEGACY_USER_SYNCVALUES,
-                        )
-                    var suppressBackendOperation = false
-
-                    if (legacyUserSyncString != null) {
-                        val legacyUserSyncJSON = JSONObject(legacyUserSyncString)
-                        val notificationTypes = legacyUserSyncJSON.safeInt("notification_types")
-
-                        val pushSubscriptionModel = SubscriptionModel()
-                        pushSubscriptionModel.id = legacyPlayerId
-                        pushSubscriptionModel.type = SubscriptionType.PUSH
-                        pushSubscriptionModel.optedIn =
-                            notificationTypes != SubscriptionStatus.NO_PERMISSION.value && notificationTypes != SubscriptionStatus.UNSUBSCRIBE.value
-                        pushSubscriptionModel.address =
-                            legacyUserSyncJSON.safeString("identifier") ?: ""
-                        if (notificationTypes != null) {
-                            pushSubscriptionModel.status = SubscriptionStatus.fromInt(notificationTypes) ?: SubscriptionStatus.NO_PERMISSION
-                        } else {
-                            pushSubscriptionModel.status = SubscriptionStatus.SUBSCRIBED
-                        }
-
-                        pushSubscriptionModel.sdk = OneSignalUtils.sdkVersion
-                        pushSubscriptionModel.deviceOS = Build.VERSION.RELEASE
-                        pushSubscriptionModel.carrier = DeviceUtils.getCarrierName(
-                            services.getService<IApplicationService>().appContext,
-                        ) ?: ""
-                        pushSubscriptionModel.appVersion = AndroidUtils.getAppVersion(
-                            services.getService<IApplicationService>().appContext,
-                        ) ?: ""
-
-                        configModel!!.pushSubscriptionId = legacyPlayerId
-                        subscriptionModelStore!!.add(
-                            pushSubscriptionModel,
-                            ModelChangeTags.NO_PROPOGATE,
-                        )
-                        suppressBackendOperation = true
-                    }
-
-                    createAndSwitchToNewUser(suppressBackendOperation = suppressBackendOperation)
-
-                    operationRepo!!.enqueue(
-                        LoginUserFromSubscriptionOperation(
-                            configModel!!.appId,
-                            identityModelStore!!.model.onesignalId,
-                            legacyPlayerId,
-                        ),
-                    )
-                    preferencesService!!.saveString(
-                        PreferenceStores.ONESIGNAL,
-                        PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
-                        null,
-                    )
-                }
-            } else {
-                Logging.debug("initWithContext: using cached user ${identityModelStore!!.model.onesignalId}")
-            }
-
-            // schedule service starts out of main thread
-            startupService.scheduleStart()
-
-            isInitialized = true
-            return true
         }
+
+        updateConfig()
+        val startupService = bootstrapServices()
+        initUser(forceCreateUser)
+        startupService.scheduleStart()
+        latchAwaiter.release()
+        return true
     }
 
     override fun login(
@@ -350,9 +387,11 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
     ) {
         Logging.log(LogLevel.DEBUG, "login(externalId: $externalId, jwtBearerToken: $jwtBearerToken)")
 
-        if (!isInitialized) {
-            throw Exception("Must call 'initWithContext' before 'login'")
+        if (!initState.isSDKAccessible()) {
+            throw IllegalStateException("Must call 'initWithContext' before 'login'")
         }
+
+        waitForInit()
 
         var currentIdentityExternalId: String? = null
         var currentIdentityOneSignalId: String? = null
@@ -402,9 +441,11 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
     override fun logout() {
         Logging.log(LogLevel.DEBUG, "logout()")
 
-        if (!isInitialized) {
-            throw Exception("Must call 'initWithContext' before 'logout'")
+        if (!initState.isSDKAccessible()) {
+            throw IllegalStateException("Must call 'initWithContext' before 'logout'")
         }
+
+        waitForInit()
 
         // only allow one login/logout at a time
         synchronized(loginLock) {
@@ -504,4 +545,29 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
     override fun <T> getServiceOrNull(c: Class<T>): T? = services.getServiceOrNull(c)
 
     override fun <T> getAllServices(c: Class<T>): List<T> = services.getAllServices(c)
+
+    private fun waitForInit() {
+        latchAwaiter.await()
+    }
+
+    private fun <T> waitAndReturn(getter: () -> T): T {
+        when (initState) {
+            InitState.NOT_STARTED -> {
+                throw IllegalStateException("Must call 'initWithContext' before use")
+            }
+            InitState.IN_PROGRESS -> {
+                Logging.debug("Waiting for init to complete...")
+                waitForInit()
+            }
+            InitState.FAILED -> {
+                throw IllegalStateException("Initialization failed. Cannot proceed.")
+            }
+            else -> {
+                // SUCCESS
+                waitForInit()
+            }
+        }
+
+        return getter()
+    }
 }
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
@@ -1,0 +1,88 @@
+package com.onesignal.common.threading
+
+import com.onesignal.common.AndroidUtils
+import com.onesignal.debug.LogLevel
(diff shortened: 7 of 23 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -51,10 +53,20 @@ import com.onesignal.user.internal.subscriptions.SubscriptionModelStore
...
 internal class OneSignalImp : IOneSignal, IServiceProvider {
+    @Volatile
+    private var initLatch = CountDownLatch(1)
+
+    @Volatile
+    private var isInitializedSuccess: Boolean = false
+
     override val sdkVersion: String = OneSignalUtils.sdkVersion
-    override var isInitialized: Boolean = false
+    override var isInitialized = false
```
jkasten2: The difference between `isInitialized` and `isInitializedSuccess` isn't clear enough based on the name alone. Also do we still need it if we also have `isInitStarted`?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,169 +170,218 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-            startupService.scheduleStart()
-
             isInitialized = true
+
             return true
         }
+
+        // If the appId is not supplied, then it is called from internal classes. We will need to return
+        // a Boolean indicating whether it is safe to proceed.
+        initEssentials(context)
```
jkasten2, line 352: This still does disk I/O on the main thread so it does look like this will solve ANRs that come from the FCM broadcast receiver. Can we also move this off the main thread?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,169 +170,218 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-                        configModel!!.pushSubscriptionId = legacyPlayerId
-                        subscriptionModelStore!!.add(
-                            pushSubscriptionModel,
-                            ModelChangeTags.NO_PROPOGATE,
-                        )
-                        suppressBackendOperation = true
-                    }
+        // if an appId is supplied, we can move all loading in background and the countdown latch
+        // will free up accessors when initialization is fully completed
+        if (appId != null) {
```
jkasten2: The code differences between an public init (non-null appId) and an internal init (null appId) is hard to read in a single method. Can we split this logic into different methods?

Examples/OneSignalDemo/app/src/test/java/com.onesignal.sdktest/application/DemoAppTests.kt
```diff
@@ -0,0 +1,138 @@
...
+        shouldThrow<IllegalStateException> {
+            OneSignal.Session
+        }
+        shouldThrow<IllegalStateException> {
+            OneSignal.Notifications
+        }
+        shouldThrow<IllegalStateException> {
+            OneSignal.Location
+        }
+    }
```
jkasten2: Blocker before merging: We discuss this before but making a comment here so it is tracked, we need to move all SDK related test out of this example app.

jkasten2 (changes requested):

Examples/OneSignalDemo/app/build.gradle
```diff
@@ -101,4 +118,20 @@ dependencies {
...
+    testImplementation(project(':OneSignal:testhelpers'))
+
+    testImplementation("io.kotest:kotest-runner-junit5:$kotestVersion")
+    testImplementation("io.kotest:kotest-runner-junit5-jvm:$kotestVersion")
+    testImplementation("io.kotest:kotest-assertions-core:$kotestVersion")
+    testImplementation("io.kotest:kotest-property:$kotestVersion")
+    testImplementation("org.robolectric:robolectric:4.8.1")
+    // kotest-extensions-android allows Robolectric to work with Kotest via @RobolectricTest
+    testImplementation("br.com.colman:kotest-extensions-android:0.1.1")
+    testImplementation("androidx.test:core-ktx:1.4.0")
```
abdulraqeeb33: can we create variables for these version numbers? atleast until we move to kotlin dsl

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Good point, it is confusing to have multiple variables tracking the initialization state. A new enum class is created to better reflect the initialization state.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Yes this is part of the slow path that will look into config value and can potentially block if SharedPreferences isn't available. I have included a test case for this scenario. To prevent ANRs, I have modified all receivers so they call the init in background, and potentially switch back to main when ready for processes involving the app lifecycle or UI operations that cannot be called from background.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Good point, the code is now refactored into two straightforward paths.

Examples/OneSignalDemo/app/src/test/java/com.onesignal.sdktest/application/DemoAppTests.kt
jinliu9508: All unit tests are now moved to core.internal.application. I have also added a few more since this review.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
```diff
@@ -119,3 +126,15 @@ fun suspendifyOnThread(
         }
     }
 }
+
+/**
+ * Create an awaiter for OneSignal SDK specifically.
+ */
+fun createOneSignalAwaiter() = LatchAwaiter("OneSignal SDK")
```
abdulraqeeb33: maybe we dont need these two methods?

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
```diff
@@ -0,0 +1,252 @@
...
+class BlockingPrefsContext(
+    context: Context,
+    private val unblockTrigger: LatchAwaiter,
+) : ContextWrapper(context) {
+    override fun getSharedPreferences(
+        name: String,
+        mode: Int,
+    ): SharedPreferences {
+        try {
+            unblockTrigger.waitForCompletion(2000)
```
abdulraqeeb33: could you create a constant out of it and then use it here 
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2368/files#diff-fb369a3cc6f7fc9583ca8b51130067cefc80216705d418adc2989898581dfcc6R107 - 500
and https://github.com/OneSignal/OneSignal-Android-SDK/pull/2368/files#diff-fb369a3cc6f7fc9583ca8b51130067cefc80216705d418adc2989898581dfcc6R198 here + 200

and other places as well?

And please make sure to pass the constant value into the BlockingPrefsContext rather than just using a public constant

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
```diff
@@ -46,25 +46,24 @@ abstract class NotificationOpenedActivityBase : Activity() {
...
-            onComplete = {
-                // Safely finish the activity on the main thread after processing is complete.
-                // This gives the system enough time to complete rendering before closing the Trampoline activity.
-                runOnUiThread {
+        OneSignal.initWithContext(applicationContext) { success ->
+            if (!success) {
+                return@initWithContext
+            }
+
+            suspendifyOnThread(
```
abdulraqeeb33: i am not sure how much of an overhead it is to switch from main thread to a background here and other places. But its probably ok.

https://github.com/OneSignal/OneSignal-Android-SDK/pull/2368/files#diff-e4d1f47b8f15feda823352af3f78a94f58f319e6b89b3790e570cb53ba06e625R217

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/bridges/OneSignalHmsEventBridge.kt
```diff
@@ -63,42 +63,43 @@ object OneSignalHmsEventBridge {
...
-            return
-        }
+        OneSignal.initWithContext(context) { success ->
+            if (!success) {
+                return@initWithContext
+            }
 
-        var time = OneSignal.getService<ITime>()
-        val bundleProcessor = OneSignal.getService<INotificationBundleProcessor>()
+            val time = OneSignal.getService<ITime>()
```
abdulraqeeb33: 💯  thank you for making the minor updates

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +169,228 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-            // schedule service starts out of main thread
-            startupService.scheduleStart()
-
-            isInitialized = true
-            return true
+                updateConfig()
+                val startupService = bootstrapServices()
+                initUser(forceCreateUser)
+                startupService.scheduleStart()
+                latchAwaiter.completeSuccess()
```
abdulraqeeb33: did have to set the initState to InitState.Success here?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
abdulraqeeb33, line 1: I think in general this class needs to be refactored. We should have a couple of smaller classes here.
1. for init
2. for createUser
3. login
That way each of these classes are testable on their own without disturbing the parent class.
This could be something we could tackle later as well?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
abdulraqeeb33, line 1: We should probably write a few tests for this class as well.
Since this is the one helping wait and throw out exceptions rather than having the app result in an ANR.

abdulraqeeb33 (commented):
a few nits and some must haves

Examples/OneSignalDemo/app/build.gradle
jinliu9508: These testImplementations are no longer there as we removed the app level tests out.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
jinliu9508: There is overhead for thread switching but it is negligible as long as we don't repeat it and only run it one time like in the NotificationOpenActivity or Receivers. And UI updates after background work is a typical flow to prevent lengthy work from clogging the main thread.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
```diff
@@ -204,8 +206,30 @@ object OneSignal {
...
+    /**
+     * Ensures OneSignal is initialized with the given context.
+     * Returns true if initialization succeeds, false otherwise.
+     */
+    suspend fun ensureOneSignalInitialized(context: Context): Boolean {
+        return kotlinx.coroutines.suspendCancellableCoroutine { cont ->
+            initWithContext(context) { initSuccess ->
+                cont.resume(initSuccess)
+            }
+        }
```
jkasten2: Let's not introduce logic to this file, it is designed to only be an interface to map method calls.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
```diff
@@ -204,8 +206,30 @@ object OneSignal {
      * THIS IS AN INTERNAL INTERFACE AND SHOULD NOT BE USED DIRECTLY.
      */
     @JvmStatic
-    fun initWithContext(context: Context): Boolean {
-        return oneSignal.initWithContext(context, null)
+    fun initWithContext(
+        context: Context,
+        onCompleted: ((Boolean) -> Unit),
```
jkasten2: Recommend this also be named `onCompleteOnMain`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
```diff
@@ -204,8 +206,30 @@ object OneSignal {
...
+    ) {
+        // init in background and signal the return value in callback
+        suspendifyOnThread(
+            // TODO: timeout?
```
jkasten2: Address TODO

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
```diff
@@ -79,21 +79,28 @@ fun suspendifyOnThread(
...
+    block: suspend () -> T,
+    onCompleteOnMain: ((Result<T>) -> Unit)? = null,
 ) {
     thread(priority = priority) {
+        val result =
+            try {
+                val value = runBlocking { block() }
+                Result.success(value)
+            } catch (e: Exception) {
+                Result.failure(e)
```
jkasten2: We used to call `Logging.error` if something throw from running `block()`. I recommend we keep this behavior so we know something went wrong when someone sends us a log.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
jinliu9508: Removed in the follow up commit

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +169,228 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+        return result
+    }
 
-            // bootstrap services
-            startupService.bootstrap()
+    private suspend fun suspendInitInternal(
+        context: Context,
+        appId: String?,
+    ): Boolean =
+        withContext(kotlinx.coroutines.Dispatchers.Default) {
```
jkasten2: Why not use `OSPrimaryCoroutineScope` here? `Dispatchers.Default` is global, anything else using it in the app could block this.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
jinliu9508: Good idea, now each test specify its own timeout for different testing purposes

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Good catch, I have added them in the caller instead and updated a few tests to also test the isInitialized state.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
jinliu9508: This is important, done in the follow up commit.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +169,229 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
             return true
         }
+
+        // app == null
+        // Slow path: wait for result synchronously; internal receivers and workers will need to init
+        // in background
+        val result =
+            runBlocking {
+                suspendInitInternal(context, appId)
+            }
```
jkasten2: Running on the background thread vs the main thread based on the `appId` is a very unexpected side effect. I would expect the caller to decide that.

I would recommend either we always run on the background thread and have the caller wait on it if it needs to, or make a new method that explicitly has the threading model in it's name. The former is probably a bit better, but the later could work if it's designed as a helper method instead if being more DRY is important.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
```diff
@@ -204,8 +206,30 @@ object OneSignal {
...
+            },
+        )
+    }
+
+    /**
+     * Ensures OneSignal is initialized with the given context.
+     * Returns true if initialization succeeds, false otherwise.
+     */
+    suspend fun ensureOneSignalInitialized(context: Context): Boolean {
+        return kotlinx.coroutines.suspendCancellableCoroutine { cont ->
```
jkasten2: The cancel feature is not used, can we use `suspendCoroutine` instead then?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
```diff
@@ -204,8 +206,30 @@ object OneSignal {
...
+            initWithContext(context) { initSuccess ->
+                cont.resume(initSuccess)
```
jkasten2: nit: Name could be simplified to `successful`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
jinliu9508: Addressed in the fixup commit. No more logic in this file and fun initWithContext(context: Context): Boolean is now suspend.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
jinliu9508: The signature is now changed to suspend and the onCompleted block is now removed.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
jinliu9508: The logic is now removed from this file in the fixup.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
jinliu9508: This file will remain unchanged as no longer needed in the fixup.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: I observed a significant performance drop by switching OSPrimariyCoroutineScope that uses a single thread context. I think Dispatchers.Default has optimal resource utilization here as it uses a shared thread pool and no thread creation overhead for each call. It is good for short lived operations like initialization. OSPrimaryCoroutineScope is good for long lived scope like OperationRepo that is constantly being used.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Good point. I have addressed this in the followup by making the internal initWithContext suspend

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
abdulraqeeb33, line 1: In the description you have mentioned that Public API changes, is that accurate?
image

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
```diff
@@ -35,21 +35,21 @@ import com.onesignal.debug.internal.logging.Logging
...
-        if (!OneSignal.initWithContext(this)) {
-            return false
-        }
-
-        var backgroundService = OneSignal.getService<IBackgroundManager>()
-
         suspendifyOnThread {
+            // init OneSignal in background
+            if (!OneSignal.initWithContext(this)) {
+                jobFinished(jobParameters, false)
```
abdulraqeeb33, line 41: did we need this here @jinliu9508 ? looks like jobFinished was not called earlier when returning false

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/UpgradeReceiver.kt
```diff
@@ -31,26 +31,32 @@ import android.content.Context
...
             return
         }
 
-        if (!OneSignal.initWithContext(context.applicationContext)) {
-            return
-        }
+        // init OneSignal and enqueue restore work in background
+        suspendifyOnThread {
+            if (!OneSignal.initWithContext(context.applicationContext)) {
+                Logging.warn("NotificationRestoreReceiver skipped due to failed OneSignal init")
```
abdulraqeeb33: typo - upgradeReceiver

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
```diff
@@ -72,13 +72,14 @@ class NotificationOpenedActivityHMS : Activity() {
     }
 
     private fun processOpen(intent: Intent?) {
-        if (!OneSignal.initWithContext(applicationContext)) {
-            return
-        }
+        // run init in background
+        suspendifyOnThread {
+            if (!OneSignal.initWithContext(applicationContext)) {
```
abdulraqeeb33, line 76: add a log statement here?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
+package com.onesignal.common.threading
+
+import com.onesignal.common.AndroidUtils
+import com.onesignal.debug.internal.logging.Logging
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+/**
+ * A generic latch that allows waiting for asynchronous initialization or completion
+ * with timeout support and detailed logging.
```
jkasten2: Can we remove the word "generic" and explain in more detail the purpose of this class? Another sentence or two worth of specifics around the main / UI thread and ANR use-cases.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
...
+ */
+class LatchAwaiter(
+    private val componentName: String = "Component",
+) {
+    companion object {
+        const val DEFAULT_TIMEOUT_MS = 30_000L // 30 seconds
+        const val ANDROID_ANR_TIMEOUT_MS = 4_800L // Conservative ANR threshold
+    }
+
+    private val latch = CountDownLatch(1)
```
jkasten2, line 25: nit: No action needed, just observations. Since we always use `1` here there might be something a bit simpler than a `CountDownLatch`. `Object.wait` and `notify` or a `Channel` are options but `CountDownLatch` is already pretty light and has a easy to use interface. Longer term would be good to have some guideline where and when to use these differ options.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    private fun getDefaultTimeout(): Long {
+        val isMainThread =
+            try {
+                AndroidUtils.isRunningOnMainThread()
+            } catch (_: Throwable) {
+                false
+            }
```
jkasten2: Can we remove the try-catch here? I don't see how this could ever throw. Also pre-existing usages in the code base also don't have a try-catch.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
...
+        return if (isMainThread) ANDROID_ANR_TIMEOUT_MS else DEFAULT_TIMEOUT_MS
+    }
+
+    private fun createTimeoutMessage(timeoutMs: Long): String {
+        val isMainThread =
+            try {
+                AndroidUtils.isRunningOnMainThread()
+            } catch (_: Throwable) {
+                false
+            }
```
jkasten2: Can we remove the try-catch here? I don't see how this could ever throw. Also pre-existing usages in the code base also don't have a try-catch.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        return if (isMainThread) {
+            "Timeout waiting for $componentName after ${timeoutMs}ms on the main thread. " +
+                "This can cause ANRs. Consider calling from a background thread."
+        } else {
+            "Timeout waiting for $componentName after ${timeoutMs}ms."
+        }
+    }
+
+    private fun logAllThreadsInfo(): String {
```
jkasten2: nit, naming, I don't think `Info` in the name adds anything. Maybe just call this `logAllThreads`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Wait for the latch and throw if not completed in time.
+     */
+    fun awaitOrThrow(timeoutMs: Long = getDefaultTimeout()) {
+        if (!await(timeoutMs)) {
+            throw IllegalStateException("$componentName initialization timeout after ${timeoutMs}ms")
+        }
+    }
```
jkasten2: Unused code (only see it tests) please remove.
* Also make sure to remove from comments too.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,106 @@
...
+        for ((thread, stack) in allThreads) {
+            sb.append("ThreadDump Thread: ${thread.name} [${thread.state}]\n")
+            for (element in stack) {
+                sb.append("\tat $element\n")
+            }
+        }
+
+        val output = sb.toString()
+        Logging.debug("Thread dump:\n$output")
+        return output
```
jkasten2: I see this method does two things:
1. Return a string
2. Print to the log

I see no reason why it should do both

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
```diff
@@ -22,21 +25,29 @@ class PermissionsActivity : Activity() {
...
-        handleBundleParams(intent.extras)
+            // finishActivity() and handleBundleParams must be called from main
+            withContext(Dispatchers.Main) {
+                if (!initialized) {
+                    finishActivity()
+                    return@withContext
+                }
+
+                requestPermissionService = OneSignal.getService()
+                preferenceService = OneSignal.getService()
```
jkasten2, line 46: Shouldn't these two lines run from `suspendifyOnThread` instead of `withContext(Dispatchers.Main)`? Otherwise we are not lower the main thread load here.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -504,4 +566,29 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+                Logging.debug("Waiting for init to complete...")
+                waitForInit()
+            }
+            InitState.FAILED -> {
+                throw IllegalStateException("Initialization failed. Cannot proceed.")
+            }
+            else -> {
+                // SUCCESS
+                waitForInit()
+            }
```
jkasten2, line 568: Do we need to call `waitForInit()` here?

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
```diff
@@ -0,0 +1,109 @@
...
+        test("waits for delayed completion") {
+            val completionDelay = 300L
+            val timeoutMs = 2000L
+
+            val startTime = System.currentTimeMillis()
+
+            // Simulate delayed success from another thread
+            Executors.newSingleThreadScheduledExecutor().schedule({
+                awaiter.release()
+            }, completionDelay, TimeUnit.MILLISECONDS)
```
jkasten2: nit: using `suspendifyOnThread` and `Thread.sleep` might be a bit easier to read here.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
```diff
@@ -0,0 +1,109 @@
...
+    context("timeout scenarios") {
+
+        test("await returns false when timeout expires") {
+            val timeoutMs = 200L
+            val startTime = System.currentTimeMillis()
+
+            val completed = awaiter.await(timeoutMs)
+            val duration = System.currentTimeMillis() - startTime
+
+            completed.shouldBeFalse()
```
jkasten2: nit: Let's keep asserts in a consistent format, use `completed shouldBe false` over `completed.shouldBeFalse()`.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
```diff
@@ -0,0 +1,109 @@
...
+            duration shouldBeGreaterThan (timeoutMs - 30)
+            duration shouldBeLessThan (timeoutMs + 100)
+        }
+
+        test("timeout of 0 returns false immediately") {
+            val startTime = System.currentTimeMillis()
+            val completed = awaiter.await(0)
+            val duration = System.currentTimeMillis() - startTime
+
+            completed.shouldBeFalse()
```
jkasten2: nit: same as above, let's keep asserts in a consistent format

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
```diff
@@ -0,0 +1,109 @@
...
+            duration shouldBeLessThan (20L)
+        }
+
+        test("timeout of 0 throws immediately in awaitOrThrow") {
+            val startTime = System.currentTimeMillis()
+            shouldThrow<IllegalStateException> {
+                awaiter.awaitOrThrow(0)
+            }
+            val duration = System.currentTimeMillis() - startTime
+            duration shouldBeLessThan (20L)
```
jkasten2: nit: Same `infix` comment as above.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
```diff
@@ -0,0 +1,109 @@
...
+            duration shouldBeLessThan (timeoutMs + 100)
+        }
+
+        test("timeout of 0 returns false immediately") {
+            val startTime = System.currentTimeMillis()
+            val completed = awaiter.await(0)
+            val duration = System.currentTimeMillis() - startTime
+
+            completed.shouldBeFalse()
+            duration shouldBeLessThan (20L)
```
jkasten2: nit: drop parentheses. `should*` methods are `infix`, so they don't need parentheses to call them.
* I would say parentheses are only helpful with a infix if there is some operation being done, such as "string" + myStr.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
```diff
@@ -0,0 +1,278 @@
...
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.runBlocking
+
+@RobolectricTest
+class SDKInitTests : FunSpec({
+
+    beforeAny {
+        val context = getApplicationContext<Context>()
+        val prefs = context.getSharedPreferences("OneSignal", Context.MODE_PRIVATE)
+        prefs.edit().clear().commit()
```
jkasten2, line 30: Are these 3 lines needed? Tests pass locally for me without them.

If they are needed, I would expect this should be `afterAny`, otherwise the last test that runs may not clean up sharedPreferences.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
```diff
@@ -46,25 +46,17 @@ abstract class NotificationOpenedActivityBase : Activity() {
...
-                val openedProcessor = OneSignal.getService<INotificationOpenedProcessor>()
-                openedProcessor.processFromContext(this, intent)
-                // KEEP: Xiaomi Compatibility:
-                // Must keep this Activity alive while trampolining, that is
-                // startActivity() must be called BEFORE finish(), otherwise
-                // the app is never foregrounded.
-            },
-            onComplete = {
-                // Safely finish the activity on the main thread after processing is complete.
-                // This gives the system enough time to complete rendering before closing the Trampoline activity.
```
jkasten2, line 63: Please keep these comments, it's important so authors of future changes don't break Xiaomi compatibility, it is very easy to do so and very hard to debug if it happens.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/BootUpReceiver.kt
```diff
@@ -30,19 +30,24 @@ import android.content.BroadcastReceiver
...
         context: Context,
         intent: Intent,
     ) {
-        if (!OneSignal.initWithContext(context.applicationContext)) {
-            return
-        }
-
-        val restoreWorkManager = OneSignal.getService<INotificationRestoreWorkManager>()
+        // in background, init onesignal and begin enqueueing restore work
+        suspendifyOnThread {
```
jkasten2, line 44: This was moved to a background thread but it does not follow these `BroadcastReceiver` AOSP rules:

> Because a receiver's [onReceive(Context, Intent)](https://developer.android.com/reference/android/content/BroadcastReceiver#onReceive(android.content.Context,%20android.content.Intent)) method runs on the main thread, it should execute and return quickly. If you need to perform long-running work, be careful about spawning threads or starting background services because the system can kill the entire process after onReceive() returns. For more information, see [Effect on process state](https://developer.android.com/develop/background-work/background-tasks/broadcasts#effects-process-state) To perform long running work, we recommend:
> 
> Calling [goAsync()](https://developer.android.com/reference/android/content/BroadcastReceiver#goAsync()) in your receiver's onReceive() method and passing the [BroadcastReceiver.PendingResult](https://developer.android.com/reference/android/content/BroadcastReceiver.PendingResult) to a background thread. This keeps the broadcast active after returning from onReceive(). However, even with this approach the system expects you to finish with the broadcast very quickly (under 10 seconds). It does allow you to move work to another thread to avoid glitching the main thread.

Source: https://developer.android.com/develop/background-work/background-tasks/broadcasts#security-considerations

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/FCMBroadcastReceiver.kt
```diff
@@ -23,48 +27,56 @@ class FCMBroadcastReceiver : BroadcastReceiver() {
             return
         }
 
-        if (!OneSignal.initWithContext(context.applicationContext)) {
-            return
-        }
+        // process in background
+        suspendifyOnThread {
```
jkasten2, line 30: See my `BroadcastReceiver` AOSP background rules comment above.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/FCMBroadcastReceiver.kt
```diff
@@ -23,48 +27,56 @@ class FCMBroadcastReceiver : BroadcastReceiver() {
...
-        setSuccessfulResultCode()
+            setSuccessfulResultCode()
+        }
     }
 
-    private fun setSuccessfulResultCode() {
-        if (isOrderedBroadcast) {
-            resultCode = Activity.RESULT_OK
+    private suspend fun setSuccessfulResultCode() {
+        withContext(Dispatchers.Main) {
```
jkasten2: This doesn't need to run on the main thread, better if it doesn't

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/FCMBroadcastReceiver.kt
```diff
@@ -23,48 +27,56 @@ class FCMBroadcastReceiver : BroadcastReceiver() {
...
+            }
         }
     }
 
-    private fun setAbort() {
+    private suspend fun setAbort() {
         if (isOrderedBroadcast) {
-            // Prevents other BroadcastReceivers from firing
-            abortBroadcast()
+            withContext(Dispatchers.Main) {
```
jkasten2: This doesn't need to run on the main thread, better if it doesn't

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/NotificationDismissReceiver.kt
```diff
@@ -28,22 +28,29 @@ import android.content.BroadcastReceiver
...
 
 class NotificationDismissReceiver : BroadcastReceiver() {
     override fun onReceive(
         context: Context,
         intent: Intent,
     ) {
-        if (!OneSignal.initWithContext(context.applicationContext)) {
-            return
-        }
+        suspendifyOnThread {
```
jkasten2, line 44: See my `BroadcastReceiver` AOSP background rules comment above.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/UpgradeReceiver.kt
```diff
@@ -31,26 +31,32 @@ import android.content.Context
...
         // Return early if using Android 7.0 due to upgrade restore crash (#263)
         if (Build.VERSION.SDK_INT == Build.VERSION_CODES.N) {
             return
         }
 
-        if (!OneSignal.initWithContext(context.applicationContext)) {
-            return
-        }
+        // init OneSignal and enqueue restore work in background
+        suspendifyOnThread {
```
jkasten2, line 54: See my `BroadcastReceiver` AOSP background rules comment above.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
```diff
@@ -72,13 +72,14 @@ class NotificationOpenedActivityHMS : Activity() {
     }
 
     private fun processOpen(intent: Intent?) {
-        if (!OneSignal.initWithContext(applicationContext)) {
-            return
-        }
+        // run init in background
+        suspendifyOnThread {
```
jkasten2: I have concerns this is going to have side-effects. Mainly that `notificationPayloadProcessorHMS.handleHMSNotificationOpenIntent` will now run AFTER `finish()`. I recommend use the same logic from `NotificationOpenedActivityBase` to avoid changing order of execution.

jkasten2 (changes requested):
This PR is based on `main`, however we discussed we want to release it with JWT so it should be based on top of that.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +172,248 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-                        null,
-                    )
+        initState = InitState.IN_PROGRESS
+        val result =
+            try {
+                withTimeout(LatchAwaiter.ANDROID_ANR_TIMEOUT_MS) {
+                    runBlocking {
+                        suspendInitInternal(context, null)
+                    }
                 }
```
jkasten2: `withTimeout` will cause `suspendInitInternal` to stop executing. This could lead the SDK to being in an unpredictable state so I would rather we let it run.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
abdulraqeeb33: yeah since it was something super light, I didnt bother to optimize, perhaps in future we can extend this

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +172,248 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
-            } else {
-                Logging.debug("initWithContext: using cached user ${identityModelStore!!.model.onesignalId}")
+            } catch (e: TimeoutCancellationException) {
+                Logging.log(LogLevel.ERROR, "initWithContext: Initialization timed out")
+                false
+            } catch (e: Exception) {
+                Logging.log(LogLevel.ERROR, "initWithContext: Initialization failed with exception", e)
+                false
             }
```
jkasten2: Don't catch here at all, it is going to hide root causes, as all we are going to get is a very generic `throw IllegalStateException("Initialization failed. Cannot proceed.")` later.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +172,248 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+        initState = if (result) InitState.SUCCESS else InitState.FAILED
+        return result
+    }
+
+    private suspend fun suspendInitInternal(
+        context: Context,
+        appId: String?,
+    ): Boolean =
+        withContext(Dispatchers.Default) {
```
jkasten2: I see two issues with this `withContext` line
1. I don't think we should switch threads in this method, let the caller pick the thread / context to simply this.
2. This could cause delays in init based on what else the app might be putting on this generic context. We also don't use `Dispatchers.Default` anywhere else in the code base either.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
jinliu9508: fun initWithContext(context: Context): Boolean is now suspend is somewhat a API change, even though it is recommended to be used internally only but it can still be called by consumers.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
jinliu9508: This method is now no longer needed and removed

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +172,248 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+                initUser(forceCreateUser)
+                startupService.scheduleStart()
+                latchAwaiter.release()
+                return@withContext true
+            } catch (e: Throwable) {
+                Logging.error("suspendInitInternal failed!", e)
+                initState = InitState.FAILED
+                latchAwaiter.release()
+                return@withContext false
+            }
```
jkasten2: Catching `Throwable` is going to hide root cause from crash reports, since we are not rethrowing here. If this was pre-existing before this PR it is ok to keep for now, but I don't think that was the case.

I do think the logic in the catch make sense to do if there is a failure, so let's keep doing that.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignal.kt
jinliu9508: The init with a callback is no longer needed as it is now suspend.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,181 +172,248 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
 
-    override fun initWithContext(
-        context: Context,
-        appId: String?,
-    ): Boolean {
-        Logging.log(LogLevel.DEBUG, "initWithContext(context: $context, appId: $appId)")
+    private fun initEssentials(context: Context) {
+        PreferenceStoreFix.ensureNoObfuscatedPrefStore(context)
 
-        synchronized(initLock) {
```
jkasten2, line 188: I don't see a replacement for `initLock` in this PR.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
jinliu9508: initWithContext will be in background, but calls like finishActivity and handleBundleParams will need to be called from main.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
jinliu9508: I think the log statement from initWithContext in case of a failed initialization is enough here.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
jinliu9508: Good point, the logic now runs in a suspendifyBlocking. This might cause ANR as we can no longer move it to the background.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
jinliu9508: Added back

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Removing this line actually causes reflect.InvocationTargetException when returning the getter so I have to keep it

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
jinliu9508: The initWithContext will be done in background, but for calls like finishActivity and handleBundleParams they must be called from main

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
jinliu9508: Yes, since now we always return true, jobFinished needs to be called to cancel the job if initialization is unsuccessful.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: added back around isSDKAccessible so no two processes can pass at the same time

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: addressed in the followup; the method is now no long in a thread

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: addressed in the followup; no longer in try-catch

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/LatchAwaiterTests.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
jinliu9508: Yes they are needed to clean up SharedPreferences as the tests are accessing it. Good point, I have added them in the afterAny

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
jinliu9508: The logic is now unchanged. However, ANR may still occur if initialization took too long

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
jinliu9508: I think the log from a failed initWithContext is enough here.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/activities/NotificationOpenedActivityBase.kt
jinliu9508: addressed in the followup

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/BootUpReceiver.kt
jinliu9508: Good point, I have used goAsync for all the receivers for more time initializing in background

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/FCMBroadcastReceiver.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/FCMBroadcastReceiver.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/FCMBroadcastReceiver.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/NotificationDismissReceiver.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/UpgradeReceiver.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/receivers/UpgradeReceiver.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -179,169 +167,218 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+        synchronized(initLock) {
+            if (initState.isSDKAccessible()) {
+                Logging.log(LogLevel.DEBUG, "initWithContext: SDK already initialized or in progress")
+                return true
+            }
+        }
 
-            // bootstrap services
-            startupService.bootstrap()
+        initState = InitState.IN_PROGRESS
```
jkasten2: `synchronized(initLock) { }` block needs to be around `initState = InitState.IN_PROGRESS` too, otherwise it doesn't solve anything.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * This class allows blocking execution until asynchronous initialization or completion is signaled, with support for configurable timeouts and detailed logging for troubleshooting.
+ * It is designed for scenarios where certain tasks, such as SDK initialization, must finish before continuing.
+ * When used on the main/UI thread, it applies a shorter timeout and logs a thread stack trace to warn developers, helping to prevent Application Not Responding (ANR) errors caused by blocking the UI thread.
+ *
+ * Usage:
+ *   val awaiter = LatchAwaiter("OneSignal SDK Init")
+ *   awaiter.release() // when done
+ *   awaiter.awaitOrThrow() // or await() to just check
```
jkasten2: awaitOrThrow no longer exists, remove this in the comment

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jkasten2: `synchronized(initLock) { }` block has to be around ` initState = InitState.IN_PROGRESS`, as only around this read doesn't do anything on it's own.

jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/LatchAwaiter.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jinliu9508: Addressed in the fixup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
jkasten2: This code was refactor so there isn't a context / thread switch here anymore.

jkasten2 (approved):

PR #2371, opened 2025-09-25
Fix: ANR on POST to Outcomes endpoint

jinliu9508:

# Description
## One Line Summary

### Motivation
**REQUIRED -** Why is this code change being made? Or what is the goal of this PR? Examples: Fixes a specific bug, provides additional logging to debug future issues, feature to allow X.

### Scope

### OPTIONAL - Other

# Testing
## Unit testing

## Manual testing

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/outcomes/impl/OutcomeEventsController.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/outcomes/impl/OutcomeEventsController.kt
@@ -18,6 +18,7 @@ import com.onesignal.session.internal.session.ISessionService
 import com.onesignal.user.internal.backend.SubscriptionObjectType
 import com.onesignal.user.internal.identity.IdentityModelStore
 import com.onesignal.user.internal.subscriptions.ISubscriptionManager
+import java.net.HttpURLConnection
 
 internal class OutcomeEventsController(
     private val _session: ISessionService,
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/outcomes/impl/OutcomeEventsController.kt
```diff
@@ -75,10 +76,15 @@ internal class OutcomeEventsController(
 
             _outcomeEventsCache.deleteOldOutcomeEvent(event)
         } catch (ex: BackendException) {
-            Logging.warn(
-                """OutcomeEventsController.sendSavedOutcomeEvent: Sending outcome with name: ${event.outcomeId} failed with status code: ${ex.statusCode} and response: ${ex.response}
+            if (ex.statusCode == HttpURLConnection.HTTP_BAD_REQUEST) {
```
nan-li: What do you think about using our existing NetworkUtils infrastructure? To drop this for any non-RETRYABLE response

abdulraqeeb33: It seems like the backend is adding this to the header `OneSignal-Retry-Limit`

https://github.com/OneSignal/OneSignal-Android-SDK/blob/8116c601b53dd134dba2486add75e8d9a9461921/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/HttpClient.kt#L313

I am wondering if its better to add some logic on the backend that if its a 400 (bad request - which is a client error) then set the retry limit to 0? 
I can't think of any reason why a 400 should have a retry.

jinliu9508: > What do you think about using our existing NetworkUtils infrastructure? To drop this for any non-RETRYABLE response

Good point, retrying is now only for retryable response. Omitting outcome event for any other error.

abdulraqeeb33 (approved):
looks fine to me. given that we have to clean up the DB locally if its a 400

PR #2374, opened 2025-10-01
feat: Add Kotlin MainApplication and suspend initialization support

abdulraqeeb33:
- Add MainApplicationKT.kt as Kotlin version of MainApplication.java
- Add initWithContextSuspend() method for async initialization
- Refactor OneSignalImp to use IO dispatcher internally for initialization
- Add comprehensive unit tests for suspend initialization
- Rename LatchAwaiter to CompletionAwaiter for better semantics
- Add helper classes for user management (AppIdHelper, LoginHelper, LogoutHelper, UserSwitcher)
- Update build.gradle to include Kotlin coroutines dependency
- Ensure ANR prevention by using background threads for initialization

# Description
## One Line Summary

### Motivation
**REQUIRED -** Why is this code change being made? Or what is the goal of this PR? Examples: Fixes a specific bug, provides additional logging to debug future issues, feature to allow X.

### Scope

### OPTIONAL - Other

# Testing
## Unit testing

## Manual testing

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
```diff
@@ -10,9 +14,10 @@ internal class StartupService(
     }
 
     // schedule to start all startable services in a separate thread
+    @OptIn(DelicateCoroutinesApi::class)
     fun scheduleStart() {
-        Thread {
+        GlobalScope.launch(Dispatchers.Default) {
```
abdulraqeeb33: Creating a thread is very expensive. Google promotes usages of Dispatcher

https://kotlinlang.org/docs/coroutines-basics.html#coroutines-are-light-weight

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -159,12 +161,53 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
         }
-
         for (module in modules) {
-            module.register(serviceBuilder)
+            module.register(this)
         }
+    }.build()
+
+    // get the current config model, if there is one
+    private val configModel: ConfigModel by lazy { services.getService<ConfigModelStore>().model }
```
abdulraqeeb33, line 168: earlier it was a var and we could probably run into race conditions.
lazy is thread safe and handles concurrent access automatically.
and most importantly its null safe and we dont need to use `!! ` anymore.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -159,12 +161,53 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+    private var _consentRequired: Boolean? = null
+    private var _consentGiven: Boolean? = null
+    private var _disableGMSMissingPrompt: Boolean? = null
+    private val initLock: Any = Any()
+    private val loginLock: Any = Any()
+    private val logoutLock: Any = Any()
+    private val userSwitcher by lazy {
+        val appContext = services.getService<IApplicationService>().appContext
+        UserSwitcher(
```
abdulraqeeb33, line 176: the class is broken down significantly into all these subclasses to address seperation of concern and improve testability

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -351,203 +285,94 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
+    @Deprecated(
+        "Use suspend version",
+        replaceWith = ReplaceWith("login(externalId, jwtBearerToken)")
+    )
     override fun login(
         externalId: String,
         jwtBearerToken: String?,
     ) {
-        Logging.log(LogLevel.DEBUG, "login(externalId: $externalId, jwtBearerToken: $jwtBearerToken)")
+        Logging.log(LogLevel.DEBUG, "Calling deprecated login(externalId: $externalId, jwtBearerToken: $jwtBearerToken)")
```
abdulraqeeb33, line 307: Logging deprecate calls so that someday we can get an estimate its usage and plan on its removal.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -570,4 +395,128 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
         return getter()
     }
+
+
+    private fun <T> blockingGet(getter: () -> T): T {
+        try {
+            if (AndroidUtils.isRunningOnMainThread()) {
+                Logging.warn("This is called on main thread. This is not recommended.")
+            }
+        } catch (e: RuntimeException) {
```
abdulraqeeb33, line 405: The exception catch here is for mostly testing

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
```diff
@@ -129,5 +140,113 @@ interface IOneSignal {
...
+     * @param context The Android context the SDK should use.
+     * @param appId The application ID the OneSignal SDK is bound to.
+     * @param externalId The external ID of the user that is to be logged in.
+     * @param jwtBearerToken The optional JWT bearer token generated by your backend to establish
+     * trust for the login operation.  Required when identity verification has been enabled.
+     * See [Identity Verification | OneSignal](https://documentation.onesignal.com/docs/identity-verification)
+     */
+    suspend fun login(
+        context: Context,
+        appId: String,
```
abdulraqeeb33: I see a bunch of crashes where login is called before initializing the SDK.

Is it ok to just accept the appId and context here and initialize the SDK (if its not done already) and continuing with the login process? It shouldn't be very hard for the consumer to pass the AppId IMHO.

Does it impact anything else?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
```diff
@@ -129,5 +140,113 @@ interface IOneSignal {
...
+    /**
+     * Logout the current user (suspend version).
+     * Handles initialization automatically.
+     *
+     * @param context The Android context the SDK should use.
+     * @param appId The application ID the OneSignal SDK is bound to.
+     */
+    suspend fun logout(
+        context: Context,
+        appId: String,
```
abdulraqeeb33: same with logout.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
```diff
@@ -129,5 +140,99 @@ interface IOneSignal {
      * be retrieved, except through this device as long as the app remains installed and the app
      * data is not cleared.
      */
+    @Deprecated(message = "Use suspend version", ReplaceWith("suspend fun logout()"))
     fun logout()
```
jkasten2: There isn't a return value, so it is valid to call it and not wait. What do you think about keeping both for these cases?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
jkasten2: Thinking about this as a consumer of this SDK, having both `initialize` and `login` with `appId` raises questions;
1. Does it matter if which one I pass appId to?
2. Is one better than the other?, which one is preferred?
3. Does this means login is designed to move the User to a different OneSignal app?

As far as internals of the SDK goes, it doesn't matter, but you see the confusion this would add to the public API.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
abdulraqeeb33: i kept both.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
abdulraqeeb33: updated this and removed Context and AppId from login and logout methods

Examples/OneSignalDemo/app/build.gradle
```diff
@@ -61,6 +62,20 @@ android {
...
+                exclude '**/MainApplication.java'
+            }
+        }
+    }
+
+    // Alternative approach: exclude from all source sets
+    android.sourceSets.all { sourceSet ->
+        sourceSet.java.exclude '**/MainApplication.java'
+    }
+
```
jkasten2: Would rather it still compile, otherwise it will silently degrade when future code changes are made. It would also be easier to swap out for testing if we can just update the `AndroidManifest.xml`.

What compile errors is this causing that isn't easy to resolve?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/PreferencesExtensionV4.kt
```diff
@@ -0,0 +1,11 @@
+package com.onesignal.core.internal.preferences
```
jkasten2, line 1: The file name `PreferencesExtension.kt` is very generic, can we put v4 in the name so it is clearer this is for a backwards compatibility only?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
jkasten2: Might be on your list already but this probably use `OneSignalDispatchers`, either `launchOnIO` or `launchOnDefault`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/AppIdResolution.kt
```diff
@@ -0,0 +1,39 @@
...
+    preferencesService: IPreferencesService,
+): AppIdResolution {
+    var forceCreateUser = false
+    var resolvedAppId: String? = inputAppId
+
+    if (inputAppId != null) {
+        if (!configModel.hasProperty(ConfigModel::appId.name) || configModel.appId != inputAppId) {
+            forceCreateUser = true
+        }
+    } else {
```
jkasten2: nit, can avoid nested `if` statements with early returns.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/AppIdResolution.kt
```diff
@@ -0,0 +1,39 @@
...
+    val failed: Boolean,
+)
+
+fun resolveAppId(
+    inputAppId: String?,
+    configModel: ConfigModel,
+    preferencesService: IPreferencesService,
+): AppIdResolution {
+    var forceCreateUser = false
+    var resolvedAppId: String? = inputAppId
```
jkasten2: nit, would be a bit cleaner to not use these, and always return `AppIdResolution` inside each if-else block.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+) {
+    suspend fun login(
+        externalId: String,
+        jwtBearerToken: String? = null,
+    ) {
+        var currentIdentityExternalId: String? = null
+        var currentIdentityOneSignalId: String? = null
+        var newIdentityOneSignalId: String = ""
+
+        synchronized(loginLock) {
```
jkasten2: login and logout need to share the same lock.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LogoutHelper.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class LogoutHelper(
+    private val logoutLock: Any,
+    private val identityModelStore: IdentityModelStore,
+    private val userSwitcher: UserSwitcher,
+    private val operationRepo: IOperationRepo,
+    private val configModel: ConfigModel,
+) {
+    fun logout() {
+        synchronized(logoutLock) {
```
jkasten2: login and logout need to share the same lock.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/CompletionAwaiterTests.kt
```diff
@@ -0,0 +1,364 @@
...
+
+            // Complete the awaiter
+            awaiter.complete()
+
+            // Wait for all to complete
+            blockingThreads.forEach { it.join(1000) }
+
+            // All should have completed
+            blockingResults.size shouldBe 2
+            blockingResults.all { it } shouldBe true
```
jkasten2: This can be simplified to one expects, I believe it can be something like `shouldBe arrayOf(true, true)`

Examples/OneSignalDemo/app/build.gradle
abdulraqeeb33: sure, it wasn't throwing any error per say. 
I reverted the changes to ignore the exclusion of MainApplication.java

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/AppIdResolution.kt
abdulraqeeb33: yeah i have been used to have a single return. lemme update it

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/AppIdResolution.kt
```diff
@@ -0,0 +1,36 @@
...
+        return AppIdResolution(appId = configModel.appId, forceCreateUser = false, failed = false)
+    }
+
+    // Case 3: No appId provided, no configModel appId - try legacy
+    val legacyAppId = preferencesService.getLegacyAppId()
+    if (legacyAppId == null) {
+        return AppIdResolution(appId = null, forceCreateUser = false, failed = true)
+    }
+
+    return AppIdResolution(appId = legacyAppId, forceCreateUser = true, failed = false)
```
jkasten2: This early return logic is flipped from the other two cases (to be clear, there isn't a logic bug), I would expect this to be checking for `legacyAppId != null` instead.
* Also don't forget to update the return values, so the result stays the same.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -115,56 +133,78 @@ internal class OneSignalImp : IOneSignal, IServiceProvider {
...
 
-        services = serviceBuilder.build()
+    private val logoutHelper by lazy {
+        LogoutHelper(
+            lock = loginLogoutLock,
+            identityModelStore = identityModelStore,
+            userSwitcher = userSwitcher,
+            operationRepo = operationRepo,
+            configModel = configModel,
+        )
```
jkasten2, line 207: nit, parameter order is different between `logoutHelper` and `loginHelper`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
```diff
@@ -0,0 +1,182 @@
...
+    private val idManager: IDManager = IDManager,
+    private val identityModelStore: IdentityModelStore,
+    private val propertiesModelStore: PropertiesModelStore,
+    private val subscriptionModelStore: SubscriptionModelStore,
+    private val configModel: ConfigModel,
+    private val oneSignalUtils: OneSignalUtils = OneSignalUtils,
+    private val carrierName: String? = null,
+    private val deviceOS: String? = null,
+    private val androidUtils: AndroidUtils = AndroidUtils,
+    private val appContextProvider: () -> Context,
```
jkasten2, line 45: Why a provider? Why not the `Context` directly?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
```diff
@@ -0,0 +1,182 @@
...
+    fun initUser(forceCreateUser: Boolean) {
+        // create a new local user
+        if (forceCreateUser ||
+            !identityModelStore.model.hasProperty(IdentityConstants.ONESIGNAL_ID)
+        ) {
+            val legacyPlayerId =
+                preferencesService.getString(
+                    PreferenceStores.ONESIGNAL,
+                    PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
+                )
```
jkasten2: Could also move this into the new `PreferencesExtensionV4.kt` file you created.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
```diff
@@ -0,0 +1,182 @@
...
+                Logging.debug("initWithContext: creating user linked to subscription $legacyPlayerId")
+
+                // Converting a 4.x SDK to the 5.x SDK.  We pull the legacy user sync values to create the subscription model, then enqueue
+                // a specialized `LoginUserFromSubscriptionOperation`, which will drive fetching/refreshing of the local user
+                // based on the subscription ID we do have.
+                val legacyUserSyncString =
+                    preferencesService.getString(
+                        PreferenceStores.ONESIGNAL,
+                        PreferenceOneSignalKeys.PREFS_LEGACY_USER_SYNCVALUES,
+                    )
```
jkasten2: Could also move this into the new PreferencesExtensionV4.kt file you created.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
```diff
@@ -0,0 +1,182 @@
...
+                        configModel.appId,
+                        identityModelStore.model.onesignalId,
+                        legacyPlayerId,
+                    ),
+                )
+                preferencesService.saveString(
+                    PreferenceStores.ONESIGNAL,
+                    PreferenceOneSignalKeys.PREFS_LEGACY_PLAYER_ID,
+                    null,
+                )
```
jkasten2: Could also move this into the new PreferencesExtensionV4.kt file you created.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
```diff
@@ -0,0 +1,182 @@
...
+                carrier = carrierName ?: ""
+                appVersion = AndroidUtils.getAppVersion(appContext) ?: ""
+            }
+
+        configModel.pushSubscriptionId = legacyPlayerId
+        subscriptionModelStore.add(pushSubscriptionModel, ModelChangeTags.NO_PROPOGATE)
+        return true
+    }
+
+    fun initUser(forceCreateUser: Boolean) {
```
jkasten2, line 120: This could be refactored more, this method still contains mostly migration logic from v4 code.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
abdulraqeeb33: It's for ease of testing. Originally, carrierName was extracted from the context inside the class, which made unit testing difficult as will need real Android components in Junit tests.
The context provider lambda is only invoked when needed so i could write unit tests easily.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
abdulraqeeb33: good catch

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
abdulraqeeb33: you are right. good call out there. i refactored it further, its much readable now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/HttpClient.kt
```diff
@@ -111,7 +107,7 @@ internal class HttpClient(
         var retVal: HttpResponse? = null
 
         val job =
-            GlobalScope.launch(Dispatchers.IO) {
+            OneSignalDispatchers.launchOnIO {
```
abdulraqeeb33, line 110: Please note I have used our Dispatcher IO thread here now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/impl/PreferencesService.kt
```diff
@@ -175,7 +173,7 @@ internal class PreferencesService(
     }
 
     private fun doWorkAsync() =
-        GlobalScope.async(Dispatchers.IO) {
+        OneSignalDispatchers.launchOnIO {
```
abdulraqeeb33, line 173: Please note I have used our Dispatcher IO thread here now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/impl/PreferencesService.kt
```diff
@@ -25,7 +23,7 @@ internal class PreferencesService(
             PreferenceStores.ONESIGNAL to mutableMapOf(),
             PreferenceStores.PLAYER_PURCHASES to mutableMapOf(),
         )
-    private var queueJob: Deferred<Unit>? = null
+    private var queueJob: Job? = null
```
abdulraqeeb33: The Deffered<Unit> was more expensive, now I am using a Job thats lighter and we are using the threads within our Dispatcher.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/migrations/RecoverFromDroppedLoginBug.kt
```diff
@@ -35,7 +33,7 @@ class RecoverFromDroppedLoginBug(
     private val _configModelStore: ConfigModelStore,
 ) : IStartableService {
     override fun start() {
-        GlobalScope.launch(Dispatchers.IO) {
+        suspendifyOnIO {
```
abdulraqeeb33, line 36: switched to using our Dispatcher

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -155,13 +154,13 @@ internal class InAppMessagesManager(
 
             // If paused is true and an In-App Message is showing, dismiss it
             if (value && _state.inAppMessageIdShowing != null) {
-                GlobalScope.launch(Dispatchers.Main) {
+                suspendifyOnMain {
```
abdulraqeeb33, line 157: this is also a bit controlled now.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/generation/impl/NotificationGenerationProcessor.kt
```diff
@@ -70,7 +68,7 @@ internal class NotificationGenerationProcessor(
         try {
             val notificationReceivedEvent = NotificationReceivedEvent(context, notification)
             withTimeout(30000L) {
-                GlobalScope.launch(Dispatchers.IO) {
+                launchOnIO {
```
abdulraqeeb33, line 71: changed this as well

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/generation/impl/NotificationGenerationProcessor.kt
```diff
@@ -103,7 +101,7 @@ internal class NotificationGenerationProcessor(
                 try {
                     val notificationWillDisplayEvent = NotificationWillDisplayEvent(notificationJob.notification)
                     withTimeout(30000L) {
-                        GlobalScope.launch(Dispatchers.IO) {
+                        launchOnIO {
```
abdulraqeeb33, line 104: here as well

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
```diff
@@ -79,7 +76,7 @@ internal class NotificationPermissionController(
         _requestPermission.registerAsCallback(PERMISSION_TYPE, this)
         pollingWaitInterval = _configModelStore.model.backgroundFetchNotificationPermissionInterval
         registerPollingLifecycleListener()
-        coroutineScope.launch {
+        launchOnIO {
```
abdulraqeeb33, line 79: this polling is also running on our dispatcher, I wonder if we should have another thread that runs this code?

abdulraqeeb33 (commented):
@jkasten2  please check the latest comments as well.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
```diff
@@ -50,15 +55,28 @@ class SDKInitTests : FunSpec({
...
         val context = getApplicationContext<Context>()
         val blockingPrefContext = BlockingPrefsContext(context, trigger, 2000)
         val os = OneSignalImp()
         var initSuccess = true
 
+        // Clear any existing appId from previous tests by clearing SharedPreferences
+        val prefs = context.getSharedPreferences("OneSignal", Context.MODE_PRIVATE)
+        prefs.edit()
+            .clear()
+            .remove("MODEL_STORE_config") // Specifically clear the config model store
```
abdulraqeeb33: somehow the tests were failing even though I had a full clear after any test.
so i had to specifically clear this field and then add the GT_APP_ID

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/CompletionAwaiterTests.kt
jkasten2: `blockingResults.size shouldBe 2` should be removed, it is redundant with `blockingResults shouldBe arrayOf(true, true)`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserSwitcher.kt
jkasten2: I see, we can leave this then.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/impl/PreferencesService.kt
jkasten2: Hmm, not sure why we kept a reference at all. Could probably just delete this `queueJob` as we write the value, but never read it.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
jkasten2: Seems the test are passing without it now, can we remove the commented out line?

jkasten2: Actually we don't need to clear at all since it is being done with the `afterAny` in this test file, so we can remove all these lines:
```kt
  prefs.edit()
            .clear()
//            .remove("MODEL_STORE_config") // Specifically clear the config model store
            .commit()
```

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/permissions/impl/NotificationPermissionController.kt
jkasten2: I think it is fine to run this on the IO dispatcher, the work done is very light.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
abdulraqeeb33: the clear all above is not really worked effectively, when this starts, its not starting with a clean state, maybe because tests are running in parallel or maybe not. So i had explicitly added to clear the preferences.

jkasten2: I see, we can look into cleaning this up later then.

PR #2375, opened 2025-10-06
chore: Dispatcher Threads

abdulraqeeb33:
# Description
## One Line Summary
Using a dispatcher instead of spawning new threads for different operations.

### Motivation
Optimizing the app and following standard implementations.

### Scope
Internal only how the execution happens.

### OPTIONAL - Other

# Testing
## Unit testing
Added Unit tests

## Manual testing

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/CompletionAwaiter.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/CompletionAwaiter.kt
@@ -101,12 +101,29 @@ class CompletionAwaiter(
     }
 
     private fun logAllThreads(): String {
-        val allThreads = Thread.getAllStackTraces()
         val sb = StringBuilder()
-        for ((thread, stack) in allThreads) {
-            sb.append("ThreadDump Thread: ${thread.name} [${thread.state}]\n")
-            for (element in stack) {
-                sb.append("\tat $element\n")
+
+        // Add OneSignal dispatcher status first (fast)
+        sb.append("=== OneSignal Dispatchers Status ===\n")
+        sb.append(OneSignalDispatchers.getStatus())
+        sb.append("\n\n")
+
+        // Add lightweight thread info (fast)
+        sb.append("=== Thread Summary ===\n")
+        val threads = Thread.getAllStackTraces().keys
+        for (thread in threads) {
+            sb.append("Thread: ${thread.name} [${thread.state}] ${if (thread.isDaemon) "(daemon)" else ""}\n")
+        }
+
+        // Only add full stack traces for OneSignal threads (much faster)
+        sb.append("\n=== OneSignal Thread Details ===\n")
+        for ((thread, stack) in Thread.getAllStackTraces()) {
+            if (thread.name.startsWith("OneSignal")) {
+                sb.append("Thread: ${thread.name} [${thread.state}]\n")
+                for (element in stack.take(10)) { // Limit to first 10 frames
+                    sb.append("\tat $element\n")
+                }
+                sb.append("\n")
             }
         }
 
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
@@ -1,0 +1,100 @@
+package com.onesignal.common.threading
+
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import java.util.concurrent.Executors
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Manages all threading for the OneSignal SDK.
+ *
+ * We use custom thread pools instead of Android's default dispatchers
+ * to have better control over thread naming and resource usage.
+ */
+object OneSignalDispatchers {
+    private const val CORE_POOL_SIZE = 4
+
+    private class OneSignalThreadFactory(private val namePrefix: String) : ThreadFactory {
+        private val threadNumber = AtomicInteger(1)
+
+        override fun newThread(r: Runnable): Thread {
+            val thread = Thread(r, "$namePrefix-${threadNumber.getAndIncrement()}")
+            thread.isDaemon = true
+            thread.priority = Thread.NORM_PRIORITY
+            return thread
+        }
+    }
+
+    // Thread pools for different types of work
+    private val ioExecutor =
+        Executors.newFixedThreadPool(
+            CORE_POOL_SIZE,
+            OneSignalThreadFactory("OneSignal-IO"),
+        )
+
+    private val computationExecutor =
+        Executors.newFixedThreadPool(
+            CORE_POOL_SIZE,
+            OneSignalThreadFactory("OneSignal-Computation"),
+        )
+
+    // Dispatchers that wrap our thread pools
+    val IO: CoroutineDispatcher = ioExecutor.asCoroutineDispatcher()
+    val Computation: CoroutineDispatcher = computationExecutor.asCoroutineDispatcher()
+
+    // Scopes for launching coroutines
+    val IOScope = CoroutineScope(SupervisorJob() + IO)
+    val DefaultScope = CoroutineScope(SupervisorJob() + Computation)
+
+    // Utility functions for common operations
+    suspend fun <T> withIO(block: suspend () -> T): T = withContext(IO) { block() }
+
+    suspend fun <T> withComputation(block: suspend () -> T): T = withContext(Computation) { block() }
+
+    fun launchOnIO(block: suspend () -> Unit) {
+        IOScope.launch { block() }
+    }
+
+    fun launchOnDefault(block: suspend () -> Unit) {
+        DefaultScope.launch { block() }
+    }
+
+    fun <T> runBlockingOnIO(block: suspend () -> T): T = runBlocking(IO) { block() }
+
+    fun <T> runBlockingOnComputation(block: suspend () -> T): T = runBlocking(Computation) { block() }
+
+    @VisibleForTesting
+    fun shutdown() {
+        try {
+            ioExecutor.shutdown()
+            computationExecutor.shutdown()
+            IOScope.cancel()
+            DefaultScope.cancel()
+        } catch (e: Exception) {
+            println("Error during OneSignalDispatchers shutdown: ${e.message}")
+        }
+    }
+
+    fun isInitialized(): Boolean {
+        return !ioExecutor.isShutdown && !computationExecutor.isShutdown
+    }
+
+    fun getStatus(): String {
+        return """
+            OneSignalDispatchers Status:
+            - IO Executor: ${if (ioExecutor.isShutdown) "Shutdown" else "Active"}
+            - Computation Executor: ${if (computationExecutor.isShutdown) "Shutdown" else "Active"}
+            - IO Scope: ${if (IOScope.isActive) "Active" else "Cancelled"}
+            - Default Scope: ${if (DefaultScope.isActive) "Active" else "Cancelled"}
+            """.trimIndent()
+    }
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
@@ -2,11 +2,17 @@ package com.onesignal.common.threading
 
 import com.onesignal.debug.internal.logging.Logging
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
-import kotlin.concurrent.thread
 
 /**
+ * Modernized ThreadUtils that leverages OneSignalDispatchers for better thread management.
+ *
+ * This file provides utilities for bridging non-suspending code with suspending functions,
+ * now using the centralized OneSignal dispatcher system for improved resource management
+ * and consistent threading behavior across the SDK.
+ *
+ *
+ * Deprecated functions are retained for backward compatibility but redirect to new implementations.
  * Allows a non-suspending function to create a scope that can
  * call suspending functions.  This is a blocking call, which
  * means it will not return until the suspending scope has been
@@ -27,11 +33,16 @@ import kotlin.concurrent.thread
  * The `withContext` will suspend until the main thread is available, but
  * the main thread is parked via this `suspendifyBlocking`. This will
  * never recover.
+ *
+ * @deprecated Use OneSignalDispatchers.runBlockingOnIO() instead
  */
+@Deprecated(
+    message = "Use OneSignalDispatchers.runBlockingOnIO() instead",
+    replaceWith = ReplaceWith("OneSignalDispatchers.runBlockingOnIO { block() }"),
+    level = DeprecationLevel.WARNING,
+)
 fun suspendifyBlocking(block: suspend () -> Unit) {
-    runBlocking {
-        block()
-    }
+    OneSignalDispatchers.runBlockingOnIO { block() }
 }
 
 /**
@@ -39,19 +50,16 @@ fun suspendifyBlocking(block: suspend () -> Unit) {
  * call suspending functions while on the main thread.  This is a nonblocking call,
  * the scope will start on a background thread and block as it switches
  * over to the main thread context.  This will return immediately!!!
+ *
+ * @deprecated Use OneSignalDispatchers.launchOnIO() instead
  */
+@Deprecated(
+    message = "Use OneSignalDispatchers.launchOnIO() instead",
+    replaceWith = ReplaceWith("OneSignalDispatchers.launchOnIO { block() }"),
+    level = DeprecationLevel.WARNING,
+)
 fun suspendifyOnMain(block: suspend () -> Unit) {
-    thread {
-        try {
-            runBlocking {
-                withContext(Dispatchers.Main) {
-                    block()
-                }
-            }
-        } catch (e: Exception) {
-            Logging.error("Exception on thread with switch to main", e)
-        }
-    }
+    suspendifyOnMainModern(block)
 }
 
 /**
@@ -59,12 +67,19 @@ fun suspendifyOnMain(block: suspend () -> Unit) {
  * call suspending functions.  This is a nonblocking call, which
  * means the scope will run on a background thread.  This will
  * return immediately!!!
+ *
+ * @deprecated Use OneSignalDispatchers.launchOnIO() instead
  */
+@Deprecated(
+    message = "Use OneSignalDispatchers.launchOnIO() instead",
+    replaceWith = ReplaceWith("OneSignalDispatchers.launchOnIO { block() }"),
+    level = DeprecationLevel.WARNING,
+)
 fun suspendifyOnThread(
     priority: Int = -1,
     block: suspend () -> Unit,
 ) {
-    suspendifyOnThread(priority, block, null)
+    suspendifyOnIO(block)
 }
 
 /**
@@ -82,20 +97,20 @@ fun suspendifyOnThread(
  * @param onComplete An optional lambda that will be invoked on the same
  *                   background thread after [block] has finished executing.
  *                   Useful for cleanup or follow-up logic.
+ *
+ * @deprecated Use OneSignalDispatchers.launchOnIO() instead
  **/
+@Deprecated(
+    message = "Use OneSignalDispatchers.launchOnIO() instead",
+    replaceWith = ReplaceWith("OneSignalDispatchers.launchOnIO { block(); onComplete?.invoke() }"),
+    level = DeprecationLevel.WARNING,
+)
 fun suspendifyOnThread(
     priority: Int = -1,
     block: suspend () -> Unit,
     onComplete: (() -> Unit)? = null,
 ) {
-    thread(priority = priority) {
-        try {
-            runBlocking { block() }
-            onComplete?.invoke()
-        } catch (e: Exception) {
-            Logging.error("Exception on thread", e)
-        }
-    }
+    suspendifyWithCompletion(useIO = true, block = block, onComplete = onComplete)
 }
 
 /**
@@ -103,19 +118,126 @@ fun suspendifyOnThread(
  * call suspending functions.  This is a nonblocking call, which
  * means the scope will run on a background thread.  This will
  * return immediately!!!
+ *
+ * @deprecated Use OneSignalDispatchers.launchOnIO() instead
  */
+@Deprecated(
+    message = "Use OneSignalDispatchers.launchOnIO() instead",
+    replaceWith = ReplaceWith("OneSignalDispatchers.launchOnIO { block() }"),
+    level = DeprecationLevel.WARNING,
+)
 fun suspendifyOnThread(
     name: String,
     priority: Int = -1,
     block: suspend () -> Unit,
 ) {
-    thread(name = name, priority = priority) {
-        try {
-            runBlocking {
+    suspendifyOnIO(block)
+}
+
+// ===============================
+// Modern OneSignal Dispatcher Functions
+// ===============================
+
+/**
+ * Modern utility for executing suspending code on the I/O dispatcher.
+ * Uses OneSignal's centralized thread management for better resource control.
+ *
+ * @param block The suspending code to execute
+ */
+fun suspendifyOnIO(block: suspend () -> Unit) {
+    OneSignalDispatchers.launchOnIO { block() }
+}
+
+/**
+ * Modern utility for executing suspending code on the default dispatcher.
+ * Uses OneSignal's centralized thread management for CPU-intensive operations.
+ *
+ * @param block The suspending code to execute
+ */
+fun suspendifyOnDefault(block: suspend () -> Unit) {
+    OneSignalDispatchers.launchOnDefault { block() }
+}
+
+/**
+ * Modern utility for executing suspending code on the main thread.
+ * Uses OneSignal's centralized thread management with proper main thread switching.
+ *
+ * @param block The suspending code to execute
+ */
+fun suspendifyOnMainModern(block: suspend () -> Unit) {
+    OneSignalDispatchers.launchOnIO {
+        withContext(Dispatchers.Main) { block() }
+    }
+}
+
+/**
+ * Modern utility for executing suspending code with completion callback.
+ * Uses OneSignal's centralized thread management for better resource control.
+ *
+ * @param useIO Whether to use IO scope (true) or Default scope (false)
+ * @param block The suspending code to execute
+ * @param onComplete Optional callback to execute after completion
+ */
+fun suspendifyWithCompletion(
+    useIO: Boolean = true,
+    block: suspend () -> Unit,
+    onComplete: (() -> Unit)? = null,
+) {
+    if (useIO) {
+        OneSignalDispatchers.launchOnIO {
+            try {
                 block()
+                onComplete?.invoke()
+            } catch (e: Exception) {
+                Logging.error("Exception in suspendifyWithCompletion", e)
             }
-        } catch (e: Exception) {
-            Logging.error("Exception on thread '$name'", e)
+        }
+    } else {
+        OneSignalDispatchers.launchOnDefault {
+            try {
+                block()
+                onComplete?.invoke()
+            } catch (e: Exception) {
+                Logging.error("Exception in suspendifyWithCompletion", e)
+            }
+        }
+    }
+}
+
+/**
+ * Modern utility for executing suspending code with error handling.
+ * Uses OneSignal's centralized thread management with comprehensive error handling.
+ *
+ * @param useIO Whether to use IO scope (true) or Default scope (false)
+ * @param block The suspending code to execute
+ * @param onError Optional error handler
+ * @param onComplete Optional completion handler
+ */
+fun suspendifyWithErrorHandling(
+    useIO: Boolean = true,
+    block: suspend () -> Unit,
+    onError: ((Exception) -> Unit)? = null,
+    onComplete: (() -> Unit)? = null,
+) {
+    if (useIO) {
+        OneSignalDispatchers.launchOnIO {
+            try {
+                block()
+                onComplete?.invoke()
+            } catch (e: Exception) {
+                Logging.error("Exception in suspendifyWithErrorHandling", e)
+                onError?.invoke(e)
+            }
+        }
+    } else {
+        OneSignalDispatchers.launchOnDefault {
+            try {
+                block()
+                onComplete?.invoke()
+            } catch (e: Exception) {
+                Logging.error("Exception in suspendifyWithErrorHandling", e)
+                onError?.invoke(e)
+            }
         }
     }
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OSPrimaryCoroutineScope.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OSPrimaryCoroutineScope.kt
@@ -2,20 +2,20 @@ package com.onesignal.common.threading
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.newSingleThreadContext
 
 object OSPrimaryCoroutineScope {
-    // CoroutineScope tied to the main thread
-    private val mainScope = CoroutineScope(newSingleThreadContext(name = "OSPrimaryCoroutineScope"))
+    // Uses computation dispatcher for CPU-intensive operations
+    private val computationScope = CoroutineScope(OneSignalDispatchers.Computation)
 
     /**
-     * Executes the given [block] on the OS primary coroutine scope.
+     * Executes the given [block] on the computation scope.
+     * Uses OneSignal's computation dispatcher for CPU-intensive work.
      */
     fun execute(block: suspend () -> Unit) {
-        mainScope.launch {
+        computationScope.launch {
             block()
         }
     }
 
-    suspend fun waitForIdle() = mainScope.launch { }.join()
+    suspend fun waitForIdle() = computationScope.launch { }.join()
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt
@@ -1,10 +1,8 @@
 package com.onesignal.core.internal.startup
 
 import com.onesignal.common.services.ServiceProvider
+import com.onesignal.common.threading.OneSignalDispatchers
 import kotlinx.coroutines.DelicateCoroutinesApi
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.launch
 
 internal class StartupService(
     private val services: ServiceProvider,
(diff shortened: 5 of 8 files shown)


OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/ThreadingPerformanceComparisonTests.kt
```diff
@@ -0,0 +1,560 @@
...
+
+// Performance tests - run manually when needed
+// To run these tests, set the environment variable: RUN_PERFORMANCE_TESTS=true
+class ThreadingPerformanceComparisonTests : FunSpec({
+
+    beforeAny {
+        Logging.logLevel = LogLevel.NONE
+    }
+
+    test("simple performance test") {
```
abdulraqeeb33: i have added these tests to test the performance, but have them skipped. It takes really long to run them.
we could run them 

```RUN_PERFORMANCE_TESTS=true ./gradlew :onesignal:core:test --tests "com.onesignal.common.threading.*"```

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * Performance optimizations:
+ * - Lazy initialization to reduce startup overhead
+ * - Custom thread pools for both IO and Default operations
+ * - Optimized thread pool configuration (smaller pools)
+ * - Work-stealing for better load balancing
+ * - Reduced context switching overhead
+ * - Efficient thread management with controlled resource usage
+ */
+object OneSignalDispatchers {
```
jkasten2: Should probably have `internal`, so customers don't use it. Also I don't think we need it for the wrapper SDKs either.

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplicationKT.kt
```diff
@@ -64,20 +65,23 @@ class MainApplicationKT : MultiDexApplication() {
         OneSignalNotificationSender.setAppId(appId)
 
         // Initialize OneSignal asynchronously on background thread to avoid ANR
-        CoroutineScope(Dispatchers.IO).launch {
-            val success = OneSignal.initWithContextSuspend(this@MainApplicationKT, appId)
-            Log.d(Tag.LOG_TAG, "OneSignal async init success: $success")
-            
-            if (success) {
+        OneSignalDispatchers.launchOnIO {
```
jkasten2: This is an example app, so we should use OneSignal SDK APIs we don't intent to document.

Examples/OneSignalDemo/app/build.gradle
```diff
@@ -49,6 +49,21 @@ android {
...
+        }
+        // Profileable release build for performance testing
+        profileable {
+            initWith release
+            debuggable false
+            profileable true
+            minifyEnabled false
+            signingConfig signingConfigs.debug
+            // Disable proguard for easier profiling
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
```
jkasten2: Comment doesn't line up;
1. `minifyEnabled false` is a few lines above
2. If remove these profile lines if we are not going to use them in variant.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/CompletionAwaiter.kt
```diff
@@ -101,12 +101,29 @@ class CompletionAwaiter(
...
+        sb.append("\n\n")
+
+        // Add lightweight thread info (fast)
+        sb.append("=== Thread Summary ===\n")
+        val threads = Thread.getAllStackTraces().keys
+        for (thread in threads) {
+            sb.append("Thread: ${thread.name} [${thread.state}] ${if (thread.isDaemon) "(daemon)" else ""}\n")
+        }
+
+        // Only add full stack traces for OneSignal threads (much faster)
```
jkasten2, line 121: I don't understand the "(much faster)" in this comment.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/CompletionAwaiter.kt
```diff
@@ -101,12 +101,29 @@ class CompletionAwaiter(
...
-            for (element in stack) {
-                sb.append("\tat $element\n")
+
+        // Add OneSignal dispatcher status first (fast)
+        sb.append("=== OneSignal Dispatchers Status ===\n")
+        sb.append(OneSignalDispatchers.getStatus())
+        sb.append("\n\n")
+
+        // Add lightweight thread info (fast)
+        sb.append("=== Thread Summary ===\n")
```
jkasten2: Can we add the wording "All" here? This will make it a bit more clear

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Optimized threading manager for the OneSignal SDK.
+ *
+ * Performance optimizations:
+ * - Lazy initialization to reduce startup overhead
+ * - Custom thread pools for both IO and Default operations
+ * - Optimized thread pool configuration (smaller pools)
+ * - Work-stealing for better load balancing
```
jkasten2: Is work-stealing really being used here? I don't see any references to it in the code.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+                    }
+
+                // Initialize Default executor for CPU operations
+                defaultExecutor =
+                    ThreadPoolExecutor(
+                        DEFAULT_CORE_POOL_SIZE,
+                        DEFAULT_MAX_POOL_SIZE,
+                        KEEP_ALIVE_TIME_SECONDS,
+                        TimeUnit.SECONDS,
+                        LinkedBlockingQueue(10),
```
jkasten2: Should be `DEFAULT_MAX_POOL_SIZE` instead of hard coded `10`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+                        // Small queue to prevent memory bloat
+                        OptimizedThreadFactory("OneSignal-Default"),
+                    ).apply {
+                        allowCoreThreadTimeOut(false) // Keep core threads alive
+                    }
+
+                ioDispatcher = ioExecutor!!.asCoroutineDispatcher()
+                defaultDispatcher = defaultExecutor!!.asCoroutineDispatcher()
+                ioScope = CoroutineScope(SupervisorJob() + ioDispatcher!!)
+                defaultScope = CoroutineScope(SupervisorJob() + defaultDispatcher!!)
```
jkasten2: Avoid `!!` in new code

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+    private fun initializeIfNeeded() {
+        if (!isInitialized.get()) {
+            // Use double-checked locking pattern but with minimal synchronization
+            if (!isInitialized.compareAndSet(false, true)) {
+                return // Another thread already initialized
+            }
+
+            try {
+                // Initialize IO executor for I/O operations
+                ioExecutor =
```
jkasten2: `ioExecutor` and `defaultExecutor` are configured to be the same. Shouldn't these have different settings of some sort?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+                    ThreadPoolExecutor(
+                        IO_CORE_POOL_SIZE,
+                        IO_MAX_POOL_SIZE,
+                        KEEP_ALIVE_TIME_SECONDS,
+                        TimeUnit.SECONDS,
+                        LinkedBlockingQueue(10),
```
jkasten2: Should be `IO_MAX_POOL_SIZE` instead of hard coded `10`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+                        // Small queue to prevent memory bloat
```
jkasten2: Unnecessary comment

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+
+                // Initialize Default executor for CPU operations
+                defaultExecutor =
+                    ThreadPoolExecutor(
+                        DEFAULT_CORE_POOL_SIZE,
+                        DEFAULT_MAX_POOL_SIZE,
+                        KEEP_ALIVE_TIME_SECONDS,
+                        TimeUnit.SECONDS,
+                        LinkedBlockingQueue(10),
+                        // Small queue to prevent memory bloat
```
jkasten2: Unnecessary comment

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+                        allowCoreThreadTimeOut(false) // Keep core threads alive
+                    }
+
+                ioDispatcher = ioExecutor!!.asCoroutineDispatcher()
+                defaultDispatcher = defaultExecutor!!.asCoroutineDispatcher()
+                ioScope = CoroutineScope(SupervisorJob() + ioDispatcher!!)
+                defaultScope = CoroutineScope(SupervisorJob() + defaultDispatcher!!)
+            } catch (e: Exception) {
+                // Fallback to Android's default dispatchers if custom ones fail
+                println("OneSignalDispatchers: Falling back to default dispatchers due to: ${e.message}")
```
jkasten2: Use OneSignal logging instead of generic logging.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+    private var ioExecutor: ThreadPoolExecutor? = null
+    private var defaultExecutor: ThreadPoolExecutor? = null
+    private var ioDispatcher: CoroutineDispatcher? = null
+    private var defaultDispatcher: CoroutineDispatcher? = null
+    private var ioScope: CoroutineScope? = null
+    private var defaultScope: CoroutineScope? = null
+
+    // Non-blocking lazy initialization to prevent startup delays
+    private fun initializeIfNeeded() {
+        if (!isInitialized.get()) {
```
jkasten2: Use early return to remove nesting, to make the code easier to read.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    // Lazy-initialized thread pools
+    private var ioExecutor: ThreadPoolExecutor? = null
+    private var defaultExecutor: ThreadPoolExecutor? = null
+    private var ioDispatcher: CoroutineDispatcher? = null
+    private var defaultDispatcher: CoroutineDispatcher? = null
+    private var ioScope: CoroutineScope? = null
+    private var defaultScope: CoroutineScope? = null
+
+    // Non-blocking lazy initialization to prevent startup delays
```
jkasten2: Kotlin's `object` type already does lazy creation. I don't see a reason why `OneSignalDispatchers` would be accessed and `launch` isn't called.

> The initialization of an object declaration is thread-safe and done on first access.

https://kotlinlang.org/docs/object-declarations.html#object-declarations-overview


The only win for using lazy I can see here if we split up the two `ThreadPoolExecutor` instances we create, however that adds a lot of complexity now for little gain.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    val IOScope: CoroutineScope by lazy {
+        initializeIfNeeded()
+        ioScope ?: CoroutineScope(SupervisorJob() + Dispatchers.IO)
+    }
+
+    val DefaultScope: CoroutineScope by lazy {
+        initializeIfNeeded()
+        defaultScope ?: CoroutineScope(SupervisorJob() + Dispatchers.Default)
+    }
```
jkasten2, line 123: These properties seem unnecessary to be public, I would expect everything outside this class just to call `launchOnIO` and `launchOnDefault`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    // Optimized utility functions with reduced overhead
+    suspend fun <T> withIO(block: suspend () -> T): T = withContext(IO) { block() }
+
+    suspend fun <T> withDefault(block: suspend () -> T): T = withContext(Default) { block() }
```
jkasten2: Would we ever want to use this pattern of others offered in this class? I would rather have less patterns in our code base for consistency.
* If this was a library for other teams / public facing that would be a different story.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+        IOScope.launch { block() }
+    }
+
+    fun launchOnDefault(block: suspend () -> Unit) {
+        DefaultScope.launch { block() }
+    }
+
+    fun <T> runBlockingOnIO(block: suspend () -> T): T = runBlocking(IO) { block() }
+
+    fun <T> runBlockingOnDefault(block: suspend () -> T): T = runBlocking(Default) { block() }
```
jkasten2: Would we ever want to use this pattern of others offered in this class? I would rather have less patterns in our code base for consistency.

If this was a library for other teams / public facing that would be a different story.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,207 @@
...
+            - IO Queue: ${ioExec.queue.size} pending tasks
+            - Default Pool: ${defaultExec.activeCount}/${defaultExec.corePoolSize} active/core threads
+            - Default Queue: ${defaultExec.queue.size} pending tasks
+            - Total completed tasks: ${ioExec.completedTaskCount + defaultExec.completedTaskCount}
+            - Memory usage: ~${(ioExec.activeCount + defaultExec.activeCount) * 1024}KB (thread stacks, ~1MB each)
+            """.trimIndent()
+    }
+
+    @VisibleForTesting
+    fun shutdown() {
```
jkasten2: If we use `@VisibleForTesting` should this be `internal` or `private`?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: ah right, earlier i had that and then later changed the implementation to have smaller queues of 10 tasks. will cleanup

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: I was thinking we could promote the customers to use these dispatchers directly, instead of launching it on whatever thread they choose and then internally we switch to using our Dispatchers.

But then a caveat, if the customers use it and we change the internal operations of this dispatcher, we will run into issues. Better to make it internal and ignore the cost of thread switching which is anyway very minimal.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: this is the Queue capacity to hold the tasks when the threads are busy, so tasks will be added to the Q (max size 10) and then as the thread frees up, the next task in the Q is picked up. This Q will be used in extreme cases when the work load is really high and the threads are extremely busy.
but yes, for the sake of future readers, i have made it a constant and wrote a comment next to it.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: ah yes, detekt should catch all this in future

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: yes thats a good question, i created two scopes just so that we can divide the work between the two scopes easily. But yeah, how about changing the priority of the IO thread to Max Priority = 10?
what do you think?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: Updated, this is not the max pool size, its a Q to maintain tasks.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: yeah i usually try to have one return statement as sometimes its hard to read with multiple returns.
but yes, in this case, an early return should make it more readable.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: you are right, i think i was trying to over engineer it to create the dispatcher ONLY when launch is called, but its pretty obvious, the customer would obviously call `launch`. I can simply this implementation a lot.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: yeah, i had kept the scope open for the `OperationRepoTests`, but lemme make a internal method and just make it visibleForTesting

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: actually we dont need this method. i cleaned up

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
```diff
@@ -2,120 +2,144 @@ package com.onesignal.common.threading
...
-                withContext(Dispatchers.Main) {
-                    block()
-                }
-            }
-        } catch (e: Exception) {
-            Logging.error("Exception on thread with switch to main", e)
-        }
+    OneSignalDispatchers.launchOnIO {
+        withContext(Dispatchers.Main) { block() }
     }
```
jkasten2, line 28: Since we are changing this function, can we skip going to a different context and just run the block directly on the main thread? Example:
```kt
GlobalScope.launch(Dispatchers.Main) { block() }
```

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
```diff
@@ -32,7 +32,7 @@ class PermissionsActivity : Activity() {
         }
 
         // init in background
-        suspendifyOnThread {
+        suspendifyOnIO {
```
jkasten2: This is more UI related than IO

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/CompletionAwaiter.kt
abdulraqeeb33: Get stackTrace of a thread is expensive.
Earlier we were getting the stackTrace of all the threads - as we have fire and forget threads. But now its probably not necessary for us.
Now we are just getting the summary of all the threads which is much faster and then get the stack trace + summary for OneSignalDispatcher threads - which is probably all we need to fix the problem.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
```diff
@@ -263,7 +266,7 @@ class OperationRepoTests : FunSpec({
         // When
         opRepo.start()
         opRepo.enqueue(mockOperation())
-        OSPrimaryCoroutineScope.waitForIdle()
+        OneSignalDispatchers.DefaultScope.launch { }.join()
```
jkasten2: There now Default and IO, so we have to wait for both. Can we add a `OneSignalDispatchers.waitForIdle` as a test helper to wait for both?

Examples/OneSignalDemo/app/build.gradle
abdulraqeeb33: yeah cleaned up

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplicationKT.kt
abdulraqeeb33: yes, i cleaned this up, earlier thought we should have the customer use our Dispatcher, but now, lets have them start on whatever thread they want and internally we are switching to use our threads anyway.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
abdulraqeeb33: here we are running init in the background and then running the UI related stuff on the main thread.

```// init in background
        suspendifyOnIO {
            val initialized = OneSignal.initWithContext(this)

            // finishActivity() and handleBundleParams must be called from main
            withContext(Dispatchers.Main) {
                if (!initialized) {
                    finishActivity()
                    return@withContext
                }

                requestPermissionService = OneSignal.getService()
                preferenceService = OneSignal.getService()

                handleBundleParams(intent.extras)
            }
        }```

File: OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt (hunk 23 above)
Comment (abdulraqeeb33) [reply OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt]: yup done

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt (hunk 21 above)
Comment (abdulraqeeb33) [reply OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt:28]: As per documentation, we should try to avoid using GlobalScope as its not bound to any job and is like a daemon thread.  
```OneSignalDispatchers.launchOnIO {
        withContext(Dispatchers.Main) { block() }
    }``` 
    
    here its starts the coroutine on IO thread from our customer dispatcher and then switches to the main thread.

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt (hunk 18 above)
Comment (abdulraqeeb33) [reply OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt]: sure, its ultimately running the same dispatcher just a different pattern to use. Should be ok to remove it.

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt (hunk 19 above)
Comment (abdulraqeeb33) [reply OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt]: done

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt (hunk 21 above)
Comment (jkasten2) [reply OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt:28]: ah good point!

If we are keeping this `suspendifyOnMain` we can keep the implementation you have now. The best long term solution would be to replace usages of this with `ViewModelScope` or `LifecycleScope` (there might be more options), as the reason this SDK wants to run something on the main thread is to update UI. And UI has a more limited lifetime than the app as a whole.

File: OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt [hunk 24]
```diff
@@ -0,0 +1,198 @@
...
+        try {
+            ThreadPoolExecutor(
+                IO_CORE_POOL_SIZE,
+                IO_MAX_POOL_SIZE,
+                KEEP_ALIVE_TIME_SECONDS,
+                TimeUnit.SECONDS,
+                LinkedBlockingQueue(QUEUE_CAPACITY),
+                OptimizedThreadFactory(
+                    namePrefix = IO_THREAD_NAME_PREFIX,
+                    priority = Thread.MAX_PRIORITY,
```
jkasten2: I would expect IO tasks to be Default or lower priority.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,198 @@
...
+                ),
+            ).apply {
+                allowCoreThreadTimeOut(false) // Keep core threads alive
+            }
+        } catch (e: Exception) {
+            Logging.error("OneSignalDispatchers: Failed to create IO executor, using fallback: ${e.message}")
+            throw e // Let the dispatcher fallback handle this
```
jkasten2: Error message is inconsistent with what is happening in the code.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,198 @@
...
+                KEEP_ALIVE_TIME_SECONDS,
+                TimeUnit.SECONDS,
+                LinkedBlockingQueue(QUEUE_CAPACITY),
+                OptimizedThreadFactory(DEFAULT_THREAD_NAME_PREFIX),
+            ).apply {
+                allowCoreThreadTimeOut(false) // Keep core threads alive
+            }
+        } catch (e: Exception) {
+            Logging.error("OneSignalDispatchers: Failed to create Default executor, using fallback: ${e.message}")
+            throw e // Let the dispatcher fallback handle this
```
jkasten2: Error message is inconsistent with what is happening in the code.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -0,0 +1,198 @@
...
+
+    @VisibleForTesting
+    internal fun waitForDefaultScope() {
+        runBlocking {
+            // Wait for all active coroutines in DefaultScope to complete
+            DefaultScope.coroutineContext[Job]?.children?.forEach { child ->
+                child.join()
+            }
+        }
+    }
```
jkasten2: This only accounts for `DefaultScope`, won't the tests also cause task to run on `IOScope` too? If so we should have a function that waits for both.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -161,7 +161,7 @@ internal class InAppMessagesManager(
             }
 
             if (!value) {
-                suspendifyOnThread {
+                suspendifyOnIO {
                     evaluateInAppMessages()
                 }
```
jkasten2, line 167: Recommend `suspendifyOnDefault` over `suspendifyOnIO` for calls to `evaluateInAppMessages()`, as this doesn't do I/O and it results in UI updates, IAM maybe displayed.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -727,7 +727,7 @@ internal class InAppMessagesManager(
 
         makeRedisplayMessagesAvailableWithTriggers(listOf(triggerId), false)
 
-        suspendifyOnThread {
+        suspendifyOnIO {
```
jkasten2: Recommend `suspendifyOnDefault` over `suspendifyOnIO` for calls to `evaluateInAppMessages()`, as this doesn't do I/O and it results in UI updates, IAM maybe displayed.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt
```diff
@@ -739,7 +739,7 @@ internal class InAppMessagesManager(
 
         makeRedisplayMessagesAvailableWithTriggers(listOf(newTriggerKey), true)
 
-        suspendifyOnThread {
+        suspendifyOnIO {
```
jkasten2: Recommend `suspendifyOnDefault` over `suspendifyOnIO` for calls to `evaluateInAppMessages()`, as this doesn't do I/O and it results in UI updates, IAM maybe displayed.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
abdulraqeeb33: changed to default

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: yeah i got rid of this method as i could improve the tests

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
```diff
@@ -27,7 +27,10 @@ class SDKInitTests : FunSpec({
     afterAny {
         val context = getApplicationContext<Context>()
         val prefs = context.getSharedPreferences("OneSignal", Context.MODE_PRIVATE)
-        prefs.edit().clear().commit()
+        prefs.edit()
+            .clear()
+            .remove("MODEL_STORE_config") // Specifically clear the config model store
```
jkasten2: Is there a way to remove all so we don't have to maintain a list?

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
```diff
@@ -781,6 +768,102 @@ class OperationRepoTests : FunSpec({
...
+                operations.size == 1 && operations.contains(groupableOp2) -> {
+                    executionOrder.add("execute-single-groupable-2")
+                    ExecutionResponse(ExecutionResult.SUCCESS)
+                }
+                else -> {
+                    executionOrder.add("execute-other-${operations.size}")
+                    ExecutionResponse(ExecutionResult.SUCCESS)
+                }
+            }
+        }
```
jkasten2, line 824: This is very hard to read, can we clean it up to make it more maintainable?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: as discussed dropped the priority to Normal - 1

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
abdulraqeeb33: ah yes, cleared everything.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/operations/OperationRepoTests.kt
abdulraqeeb33: have tried to simplify that piece of code

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplicationKT.kt
```diff
@@ -64,20 +69,23 @@ class MainApplicationKT : MultiDexApplication() {
...
-                
+
                 // Request permission - this will internally switch to Main thread for UI operations
                 OneSignal.Notifications.requestPermission(true)
+
+                Log.d(Tag.LOG_TAG, Text.ONESIGNAL_SDK_INIT)
+
+            } catch (e: Exception) {
+                Log.e(Tag.LOG_TAG, "OneSignal initialization error", e)
             }
```
jkasten2, line 87: I don't think we should try-catch here:
1. Not something our customers need to do
2. In development I would rather the app crash so we know there is a major issue.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
jkasten2: I agree, I don't like the idea of mixing internal and external tasks with OneSignal's dispatchers or scopes:
1. It creates a lot of variability. Example, consumers could run a lot of task or even call Thread.sleep in them and stall OneSignal tasks.
2. It make it harder to change our internal implementation

Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/application/MainApplicationKT.kt
abdulraqeeb33: ok i will merge this into the refactor branch and fix it there. thanks

PR #2378, opened 2025-10-10
Feat: Detect for timezone changes and update the user

nan-li:
# Description
## One Line Summary
Start detecting timezone changes and sending it up when it changes.

## Details
- Unless users switch via logging in and out, we never update timezone (as it is only sent in the Create User request).
- Behavior aligns with https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1542

### Motivation
This was a request of clients so that when users enter new timezones, they can be properly segmented without needing a Create User call to happen.

### Scope
- No longer hydrate timezone from remote (this aligns with existing iOS behavior)
- When timezone changes on the device, the new one will be sent
- We don't need to address the migration and legacy player issue that had to be addressed by iOS: https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1542. These PR changes automatically address them because missing timezone cases will get the timezone sent on the next session (since locally it is `null` which will trigger a change detection).

# Testing
## Unit testing
Added

## Manual testing
Change timezone on emulator and see the next session, that timezone is sent with the track session start call.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2378)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
@@ -146,7 +146,7 @@ internal class LoginUserOperationExecutor(
         var identities = mapOf<String, String>()
         var subscriptions = mapOf<String, SubscriptionObject>()
         val properties = mutableMapOf<String, String>()
-        properties["timezone_id"] = TimeUtils.getTimeZoneId()!!
+        properties["timezone_id"] = TimeUtils.getTimeZoneId()
         properties["language"] = _languageContext.language
 
         if (createUserOperation.externalId != null) {


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
```diff
@@ -33,13 +34,16 @@ internal class SessionListener(
...
+    private val _propertiesModelStore: PropertiesModelStore,
     private val _outcomeEventsController: IOutcomeEventsController,
 ) : IStartableService, ISessionLifecycleHandler {
     override fun start() {
         _sessionService.subscribe(this)
     }
 
     override fun onSessionStarted() {
+        // Detect any user updates to send with the TrackSessionStartOperation
+        _propertiesModelStore.model.update()
```
abdulraqeeb33: this looks fine to me. can we write a test to make sure this method is called.
also should this be done on a background thread?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -89,9 +89,8 @@ internal class RefreshUserOperationExecutor(
                 }
             }
 
-            if (response.properties.timezoneId != null) {
-                propertiesModel.timezone = response.properties.timezoneId
-            }
+            // No longer hydrate timezone from remote, set locally
```
abdulraqeeb33, line 93: same with this.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/properties/PropertiesModel.kt
```diff
@@ -122,4 +123,8 @@ class PropertiesModel : Model() {
 
         return null
     }
+
+    fun update() {
+        timezone = TimeUtils.getTimeZoneId()
```
abdulraqeeb33: does this `TimeUtils.getTimeZoneId()` match with what the backend really expects?
Can we write a test to make sure that?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -89,9 +89,8 @@ internal class RefreshUserOperationExecutor(
...
+            propertiesModel.update()
```
abdulraqeeb33: Should this be done on a background thread?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
nan-li: Since this PR is based on `main`, we aren't making these kind of changes on a background thread already. The `update()` call will enqueue an operation, if a change is detected, which is similar to what the next line is doing with `_operationRepo.enqueue(TrackSessionStartOperation...`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
nan-li: Same comment as before, `propertiesModel.update()` just sets `propertiesModel.timezone = TimeUtils.getTimeZoneId()`. If we look at the lines above this, we are setting various properties on the `propertiesModel` like:
```
propertiesModel.country = response.properties.country
propertiesModel.tags[tagKVP.key] = tagKVP.value!!
propertiesModel.language = response.properties.language
```
Setting timezone shouldn't run separately while this object is being set up. Also, during this time while the model is being set up, no updates will fire an operation to enqueue.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/properties/PropertiesModel.kt
nan-li: Yeah, it does. Currently, the only place the SDK ever sends timezone is in a create user call [here](https://github.com/OneSignal/OneSignal-Android-SDK/blob/6018f4d987733aef11b1a9ca7c15ecfc1d0bb401/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt#L149):
```
// the non-null assert is actually extraneous
properties["timezone_id"] = TimeUtils.getTimeZoneId()!!
```
> does this TimeUtils.getTimeZoneId() match with what the backend really expects? Can we write a test to make sure that?

I don't know how we could write a test for what the server expects. We could write a test that ensures timezone is always sent as a certain format, but if the backend ever changed its expectation, we wouldn't know.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
abdulraqeeb33: ah you are right. We should be ok with the threading, but we need to write a test or two to make sure we calling the update method in the onSessionStarted()

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
abdulraqeeb33: sounds good around the background thread. but we definitely need to write a test if the properties.update() is called and if the timezone is actually set.
this helps in possible regression defects and helps us gain confidence when we are making changes.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/properties/PropertiesModel.kt
abdulraqeeb33: You are right, we cant test server expectations directly. But we should write tests to catch internal breaking changes.

jkasten2: I think the model classes should be very simple and all setting should done outside of it. So we shouldn't add a helper like this that is set a specific implementation of getting timezone.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
jkasten2: Thinking it would better if this is a different class, it can subscribe to `_applicationService.addApplicationLifecycleHandler` and in `onFocus()` update the timezone by calling   `propertiesModel.timezone = TimeUtils.getTimeZoneId()`.

I think `onFocus` is a better fit than session, as this will fire more often to reflect timezone changes. If there is an Android API that can listen to timezone changes even better.
* I believe batching will still happen if there is a new session and timezone updates within 5 seconds, as this is the way the OperationRepo batches.

nan-li: There is an API, you can use a BroadcastReceiver that listens for the [android.intent.action.TIMEZONE_CHANGED](https://developer.android.com/reference/android/content/Intent#ACTION_TIMEZONE_CHANGED) but that's a bit excessive.

> I believe batching will still happen if there is a new session and timezone updates within 5 seconds, as this is the way the OperationRepo batches.

When I tested this by putting `propertiesModel.timezone = TimeUtils.getTimeZoneId()` in a different class within `onSessionStarted` (after subscribing it to `SessionService`), then changed timezones, and restarted the app, the timezone update and track session start may or may not be sent together, depending on which class's `onSessionStarted` gets fired first. This is because `TrackSessionStartOperation` in `SessionListener` has `flush = true` to immediately send the update (after changes in the read-your-write IAM implementation). It doesn't wait 5 seconds on a new session.

However, testing now by calling `propertiesModel.timezone = TimeUtils.getTimeZoneId()` within `onFocus`, this always happens before `onSessionStarted` so the time update gets enqueued before TrackSessionStart 👍🏼

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
```diff
@@ -260,4 +265,11 @@ internal open class UserManager(
             }
         }
     }
+
+    override fun onFocus(firedOnSubscribe: Boolean) {
+        // Detect any user properties updates that changed
+        _propertiesModel.timezone = TimeUtils.getTimeZoneId()
```
abdulraqeeb33, line 271: do you want to pass TimeUtils into this class? so that you can mock it?

abdulraqeeb33: also please make sure to write tests

nan-li:
> As the tests are in the other PR

Let's not merge this until then, we can pick those commits over when done

nan-li:
Hi @abdulraqeeb33 I used the changes you made in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2382 and modified to add the tests here. Let me know if [a4a6d5e](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2378/commits/a4a6d5e6cc245188a6d3e5f138aefea5ffe6462c) looks good to you.

abdulraqeeb33 (approved):

Issue #2203, opened 2024-10-10
[Bug]: PermissionsActivity does not follow app theme


michael-winkler:
### What happened?

If we request any permission with the default android implementation the dialog looks like this:
![image](https://github.com/user-attachments/assets/5d1bbdd3-3a6d-42df-82d7-5bcf63a8eef7)

Statusbar and navigationbar is here transparent.

If we request the notification permission with the OneSignal implementation the dialog looks like this:
![image](https://github.com/user-attachments/assets/6f4f545a-6c76-4845-838a-6fe58c522c6f)

Why the difference? Can this please be improved?
The lines of the OneSignal Activity are here:

```
<activity android:name="com.onesignal.core.activities.PermissionsActivity"
                  android:theme="@android:style/Theme.Translucent.NoTitleBar"
                  android:exported="false" />
```

![image](https://github.com/user-attachments/assets/0fd0f1e3-fb93-4368-8773-99f4a8b55537)

### Steps to reproduce?

```Markdown
Just use the OneSignal code "OneSignal.Notifications.requestPermission(false)"
```

### What did you expect to happen?

A better style for the theme like the default android permission dialogs!

### OneSignal Android SDK version

5.1.23

### Android version

14

### Specific Android models

```Markdown
All
```

### Relevant log output

```Shell
-
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

michael-winkler:
This can be fixed is we do following:

```
class PermissionsActivity : Activity() {
    private var requestPermissionService: RequestPermissionService? = null
    private var preferenceService: IPreferencesService? = null
    private var permissionRequestType: String? = null

    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        if (!OneSignal.initWithContext(this)) {
            return
        }

        // Needed for the theme fix
        actionBar?.hide()

        requestPermissionService = OneSignal.getService()
        preferenceService = OneSignal.getService()

        handleBundleParams(intent.extras)
    }
...
}
```

Android Manifest:

```
<manifest xmlns:android="http://schemas.android.com/apk/res/android">

    <!-- Required so the device can access the internet. -->
    <uses-permission android:name="android.permission.INTERNET" />
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />

    <application>
        <service
          android:name="com.onesignal.core.services.SyncJobService"
[...]

PR #2388, opened 2025-10-23
Fix permissions activity

abdulraqeeb33:
# Description
## One Line Summary
Fixed the permissions activity by breaking it down into a VM and activity.

### Motivation
All the business logic is happening in the viewmodel now.

### Scope
Restricted only to PermissionsActivity.
And several testing across the apps.

### OPTIONAL - Other

# Testing
## Unit testing
Added a bunch of tests

## Manual testing
Tested it manually.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2388)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
@@ -289,7 +289,7 @@ class SDKInitTests : FunSpec({
 
         // Clean up after ourselves to avoid polluting subsequent tests
         os.logout()
-        
+
         // Wait for logout to complete with polling
         var logoutAttempts = 0
         while (os.user.externalId.isNotEmpty() && logoutAttempts < 50) {


OneSignalSDK/onesignal/core/build.gradle
```diff
@@ -68,9 +69,14 @@ ext {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
```
abdulraqeeb33, line 71: changed `stdlib-jdk7` to `stdlib`.

context: I had to add these libraries to get the standard setup going to use Android View Models and lifescope methods.
`androidx.lifecycle:lifecycle-viewmodel-ktx`
`androidx.lifecycle:lifecycle-runtime-ktx`
`androidx.activity:activity-ktx` 

And now to test this code, I had to upgrade Roboelectric test library which then began complaining about duplicate class.

`Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules 
kotlin-stdlib-1.8.10.jar -> kotlin-stdlib-1.8.10 (org.jetbrains.kotlin:kotlin-stdlib:1.8.10) 
and kotlin-stdlib-jdk8-1.7.20.jar -> kotlin-stdlib-jdk8-1.7.20 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.20)`

So I moved from `stdlib-jdk7` to `stdlib`. The stdlib is a unified class that includes all functionality, avoids multiple stdlib variants with different versions.
https://mbonnin.medium.com/the-different-kotlin-stdlibs-explained-83d7c6bf293 this guy explains in detail.

But if you folks know something where we have to be dependent for jdk 7 explicitly, please call it out

OneSignalSDK/onesignal/core/src/main/AndroidManifest.xml
```diff
@@ -11,7 +11,7 @@
           android:exported="false" />
 
         <activity android:name="com.onesignal.core.activities.PermissionsActivity"
-                  android:theme="@android:style/Theme.Translucent.NoTitleBar"
+                  android:theme="@style/NoToolBarTheme"
```
abdulraqeeb33: this should fix this issue
https://github.com/OneSignal/OneSignal-Android-SDK/issues/2203#issue-2578661712

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/ApplicationServiceTests.kt
```diff
@@ -38,12 +38,9 @@ class ApplicationServiceTests : FunSpec({
 
     test("start application service with activity shows entry state as closed") {
         // Given
-        val activity: Activity
-
-        Robolectric.buildActivity(Activity::class.java).use { controller ->
-            controller.setup() // Moves Activity to RESUMED state
-            activity = controller.get()
-        }
+        val controller1 = Robolectric.buildActivity(Activity::class.java)
```
abdulraqeeb33, line 41: had to do this since we updated the Roboelectric library

OneSignalSDK/gradle.properties
```diff
@@ -23,7 +23,13 @@
 # Remove when creating an .aar build.
 #android.enableAapt2=false
 
-org.gradle.jvmargs=-Xmx1536m
+org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError
```
abdulraqeeb33, line 26: had to increase the memory for running tests

OneSignalSDK/onesignal/core/src/main/AndroidManifest.xml
michael-winkler: Have you tested it? In my own example app (I showed the code inside the issue) I needed to call also `actionBar?.hide()` inside the code. Not sure why this step was needed.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/impl/RequestPermissionService.kt
```diff
@@ -7,6 +7,7 @@ import com.onesignal.core.activities.PermissionsActivity
 import com.onesignal.core.internal.application.IActivityLifecycleHandler
 import com.onesignal.core.internal.application.IApplicationService
 import com.onesignal.core.internal.permissions.IRequestPermissionService
+import com.onesignal.core.internal.permissions.PermissionsViewModel
```
jinliu9508, line 10: I got a compile error from this line. This should be now com.onesignal.core.internal.viewmodel.PermissionsViewModel after the removal of the duplicate class.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/permissions/PermissionsViewModel.kt
```diff
@@ -0,0 +1,224 @@
+package com.onesignal.core.internal.viewmodel
```
jinliu9508: Need a cleanup to get rid of warning of "Package directive does not match the file location"

OneSignalSDK/build.gradle
```diff
@@ -55,4 +59,22 @@ allprojects {
...
+            eachDependency { details ->
+                if (details.requested.group == 'org.jetbrains.kotlin') {
+                    if (details.requested.name == 'kotlin-stdlib-jdk7' || 
+                        details.requested.name == 'kotlin-stdlib-jdk8') {
+                        details.useTarget "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
+                    }
+                }
+            }
+        }
+    }
```
jkasten2, line 79: Recommend we make sure this substitution works before publishing the 5.4.0 release. I have concerns if this really works correctly when it comes to the dependencies in the .aar files.

OneSignalSDK/onesignal/core/build.gradle
jkasten2: We still want to support Android 5 (API level 21). This seems to still work correctly from what I have read and also tested to confirm.

OneSignalSDK/onesignal/core/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="NoToolBarTheme">
+        <item name="android:windowAnimationStyle">@android:style/Animation</item>
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:windowIsTranslucent">true</item>
+        <item name="windowNoTitle">true</item>
+        <item name="windowActionBar">false</item>
+    </style>
+</resources>
```
jkasten2, line 4: This doesn't seem to fix the problem, in the example app it changes the background color of the status bar, confirmed on an Android 14 emulator.

If this is not easy to fix we can do this in a different PR, so not block this one or a release.

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/AndroidManifest.xml
abdulraqeeb33: I will add a fix for this in the next release.

PR #2405, opened 2025-11-06
WIP - Feat: OpenTelemetry crash reporting

jkasten2:
# Description
## One Line Summary
Add OpenTelemetry libraries to report crashes to improve SDK reliability.

### Motivation
Allows us to quickly identify and fix issues caused by this SDK.

### Scope
Only adds OpenTelemetry to track crashes.

# Testing
## Unit testing
TODO: Need to add unit testing yet.

## Manual testing
Tested on an Android 14 Emulator:
* Throw RuntimeException from example app and ensure SDK can catch it
* Ensured our crash handler saves it to a file, even when it is offline
* Ensured the last crash is sent to a remote server when the app is opened again.

Tested has only been done with a known working Otel backend, HoneyComb, until we
have our own backend to receive the payload.

### Remaning Tasks
* Add unit tests
* Turn on only if in remote parameters
* uncaughtException
   - Add try-catch, so any of our logic can't cause side-effects if something goes wrong
   - Shutdown any in progress uploading of crashes
* Look into if we can use older libraries so we don't have to use Kotlin 2.2
* Crash handling only works with Android 8, due to Otel libraries, look into if it easy to support olde version of Android.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2405)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/IParamsBackendService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/IParamsBackendService.kt
@@ -36,6 +36,7 @@ class ParamsObject(
     var opRepoExecutionInterval: Long? = null,
     var influenceParams: InfluenceParamsObject,
     var fcmParams: FCMParamsObject,
+    val remoteLoggingParams: RemoteLoggingParamsObject,
 )
 
 class InfluenceParamsObject(
@@ -52,4 +53,8 @@ class FCMParamsObject(
     val projectId: String? = null,
     val appId: String? = null,
     val apiKey: String? = null,
+)
+
+class RemoteLoggingParamsObject(
+    val enable: Boolean? = null,
 )
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/crash/OneSignalCrashUploader.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/crash/OneSignalCrashUploader.kt
@@ -1,6 +1,8 @@
 package com.onesignal.debug.internal.logging.otel.crash
 
+import com.onesignal.core.internal.config.ConfigModelStore
 import com.onesignal.core.internal.startup.IStartableService
+import com.onesignal.debug.internal.logging.Logging
 import com.onesignal.debug.internal.logging.otel.IOneSignalOpenTelemetryRemote
 import com.onesignal.debug.internal.logging.otel.config.OtelConfigCrashFile
 import io.opentelemetry.sdk.logs.data.LogRecordData
@@ -15,6 +17,7 @@ import java.util.concurrent.TimeUnit
 internal class OneSignalCrashUploader(
     private val _openTelemetryRemote: IOneSignalOpenTelemetryRemote,
     private val _crashPathProvider: IOneSignalCrashConfigProvider,
+    _configModelStore: ConfigModelStore,
 ) : IStartableService {
     companion object {
         const val SEND_TIMEOUT_SECONDS = 30L
@@ -27,7 +30,14 @@ internal class OneSignalCrashUploader(
                 _crashPathProvider.minFileAgeForReadMillis
             ).iterator()
 
+    private val enable =
+        _configModelStore.model.remoteLoggingParams.enable ?: false
+
     override fun start() {
+        Logging.info("OneSignalCrashUploader.enable: $enable")
+        if (!enable) {
+            return
+        }
         runBlocking { internalStart() }
     }
 
@@ -53,8 +63,10 @@ internal class OneSignalCrashUploader(
         // another one if there isn't an issue making network calls.
         while (reports.hasNext() && !failed) {
             val future = networkExporter.export(reports.next())
+            Logging.debug("Sending OneSignal crash report")
             val result = future.join(SEND_TIMEOUT_SECONDS, TimeUnit.SECONDS)
             failed = !result.isSuccess
+            Logging.debug("Done OneSignal crash report, failed: $failed")
         }
     }
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/impl/ParamsBackendService.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/impl/ParamsBackendService.kt
@@ -11,6 +11,7 @@ import com.onesignal.core.internal.backend.FCMParamsObject
 import com.onesignal.core.internal.backend.IParamsBackendService
 import com.onesignal.core.internal.backend.InfluenceParamsObject
 import com.onesignal.core.internal.backend.ParamsObject
+import com.onesignal.core.internal.backend.RemoteLoggingParamsObject
 import com.onesignal.core.internal.http.CacheKeys
 import com.onesignal.core.internal.http.IHttpClient
 import com.onesignal.core.internal.http.impl.OptionalHeaders
@@ -57,6 +58,15 @@ internal class ParamsBackendService(
                 )
         }
 
+        // Process Remote Logging params
+        var remoteLoggingParams: RemoteLoggingParamsObject? = null
+        responseJson.expandJSONObject("remote_logging") {
+            remoteLoggingParams =
+                RemoteLoggingParamsObject(
+                    enable = it.safeBool("enable"),
+                )
+        }
+
         return ParamsObject(
             googleProjectNumber = responseJson.safeString("android_sender_id"),
             enterprise = responseJson.safeBool("enterp"),
@@ -75,6 +85,7 @@ internal class ParamsBackendService(
             opRepoExecutionInterval = responseJson.safeLong("oprepo_execution_interval"),
             influenceParams = influenceParams ?: InfluenceParamsObject(),
             fcmParams = fcmParams ?: FCMParamsObject(),
+            remoteLoggingParams = remoteLoggingParams ?: RemoteLoggingParamsObject(),
         )
     }
 
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
@@ -301,6 +301,9 @@ class ConfigModel : Model() {
     val fcmParams: FCMConfigModel
         get() = getAnyProperty(::fcmParams.name) { FCMConfigModel(this, ::fcmParams.name) } as FCMConfigModel
 
+    val remoteLoggingParams: RemoteLoggingConfigModel
+        get() = getAnyProperty(::remoteLoggingParams.name) { RemoteLoggingConfigModel(this, ::remoteLoggingParams.name) } as RemoteLoggingConfigModel
+
     override fun createModelForProperty(
         property: String,
         jsonObject: JSONObject,
@@ -313,6 +316,12 @@ class ConfigModel : Model() {
 
         if (property == ::fcmParams.name) {
             val model = FCMConfigModel(this, ::influenceParams.name)
+            model.initializeFromJson(jsonObject)
+            return model
+        }
+
+        if (property == ::remoteLoggingParams.name) {
+            val model = RemoteLoggingConfigModel(this, ::remoteLoggingParams.name)
             model.initializeFromJson(jsonObject)
             return model
         }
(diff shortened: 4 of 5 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/IOneSignalCrashReporter.kt
```diff
@@ -0,0 +1,5 @@
+package com.onesignal.debug.internal.crash
```
abdulraqeeb33, line 1: can we place all the logging related code in its own module, just in case we go in the direction of KMP, it will be extremely easy to build an artifact out this.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/OneSignalCrashHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+         *
+         * NOTE: addShutdownHook() isn't a workaround as it doesn't fire for
+         * Process.killProcess, which KillApplicationHandler calls.
+         */
+        runBlocking { _crashReporter.saveCrash(thread, throwable) }
+        existingHandler?.uncaughtException(thread, throwable)
+    }
+}
+
+internal fun isOneSignalAtFault(throwable: Throwable): Boolean =
```
abdulraqeeb33, line 63: would this take too long to execute? Wondering if we would potentially miss the window of capturing the crash log in time before the OS kills the app?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/config/OtelConfigRemoteOneSignal.kt
```diff
@@ -0,0 +1,57 @@
...
+                .addLogRecordProcessor(
+                    OtelConfigShared.LogRecordProcessorConfig.batchLogRecordProcessor(
+                        HttpRecordBatchExporter.create(extraHttpHeaders)
+                    )
+                ).setLogLimits(LogLimitsConfig::logLimits)
+                .build()
+    }
+
+    object HttpRecordBatchExporter {
+        @RequiresApi(Build.VERSION_CODES.O)
```
abdulraqeeb33, line 50: The minSdk supported as of now is 21. What would happen when this method gets called on a App that is between 21 and 26 (version_code.0)?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/crash/OneSignalCrashUploader.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private val enable =
+        _configModelStore.model.remoteLoggingParams.enable ?: false
+
+    override fun start() {
+        Logging.info("OneSignalCrashUploader.enable: $enable")
+        if (!enable) {
+            return
+        }
+        runBlocking { internalStart() }
```
abdulraqeeb33, line 41: should we suspend here instead of runBlocking?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/OneSignalOpenTelemetry.kt
```diff
@@ -0,0 +1,124 @@
...
+        } catch (_: NullPointerException) {
+            Logging.error("Auth missing for crash log reporting!")
+            ""
+        }
+
+    val extraHttpHeaders: Map<String, String> by lazy {
+        mapOf(
+            "X-OneSignal-App-Id" to appId,
+            HTTP_SDK_VERSION_HEADER_KEY to HTTP_SDK_VERSION_HEADER_VALUE,
+            "x-honeycomb-team" to "", // TODO: REMOVE
```
abdulraqeeb33, line 87: should we remove this?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -214,6 +215,11 @@ internal class OneSignalImp(
 
         // Give the logging singleton access to the application service to support visual logging.
         Logging.applicationService = applicationService
+
+        // Crash handler needs to be one of the first things we setup,
+        // otherwise we'll not report some crashes, resulting in a false sense
+        // of stability.
+        services.getService<OneSignalCrashHandler>()
```
abdulraqeeb33, line 222: Should we depend on our internal architecture to handle crashes?
IMHO, it should be its own thing and as simple as possible to handle a crash.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/restoration/impl/NotificationRestoreWorkManager.kt
```diff
@@ -17,13 +17,14 @@ internal class NotificationRestoreWorkManager : INotificationRestoreWorkManager
     // Notifications will never be force removed when the app's process is running,
     //   so we only need to restore at most once per cold start of the app.
     private var restored = false
+    private val lock = Any()
```
abdulraqeeb33, line 20: should this be part of this PR?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/IParamsBackendService.kt
```diff
@@ -53,3 +54,7 @@ class FCMParamsObject(
     val appId: String? = null,
     val apiKey: String? = null,
 )
+
+class RemoteLoggingParamsObject(
+    val enable: Boolean? = null,
```
abdulraqeeb33, line 59: should we default this to false instead of null?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/attributes/OneSignalOtelFieldsTopLevel.kt
```diff
@@ -0,0 +1,67 @@
...
+// their spec. We do this to make it clear where the source of this field is.
+internal const val OS_OTEL_NAMESPACE: String = "ossdk"
+
+/**
+ * Purpose: Fields to be included in every Otel request that goes out.
+ * Requirements: Only include fields that can NOT change during runtime,
+ * as these are only fetched once. (Calculated fields are ok)
+ */
+internal class OneSignalOtelFieldsTopLevel(
+    private val _applicationService: IApplicationService,
```
abdulraqeeb33, line 21: can we pass the bare minimum java models as opposed to internal classes? trying to see if we can decouple the code from the SDK's main code base.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/crash/OneSignalCrashConfigProvider.kt
```diff
@@ -0,0 +1,17 @@
+package com.onesignal.debug.internal.logging.otel.crash
+
+import com.onesignal.core.internal.application.IApplicationService
+import java.io.File
+
+internal class OneSignalCrashConfigProvider(
+    private val _applicationService: IApplicationService
```
abdulraqeeb33, line 7: we should be able to pass the the path directly here instead of depending on our SDK class

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/crash/OneSignalCrashReporterOtel.kt
```diff
@@ -0,0 +1,38 @@
...
+                .builder()
+                .put(OTEL_EXCEPTION_MESSAGE, throwable.message)
+                .put(OTEL_EXCEPTION_STACKTRACE, throwable.stackTraceToString())
+                .put(OTEL_EXCEPTION_TYPE, throwable.javaClass.name)
+                // This matches the top level thread.name today, but it may not
+                // always if things are refactored to use a different thread.
+                .put("$OS_OTEL_NAMESPACE.exception.thread.name", thread.name)
+                .build()
+
+        _openTelemetry
```
abdulraqeeb33, line 30: How are we sure that this is ready with all the relevant fields populated into itself?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/OneSignalOpenTelemetry.kt
```diff
@@ -0,0 +1,124 @@
...
+            sdkCachedValue = localSdk
+            return localSdk
+        }
+    }
+
+    protected abstract fun getSdkInstance(attributes: Map<String, String>): OpenTelemetrySdk
+
+    override suspend fun forceFlush(): CompletableResultCode {
+        val sdkLoggerProvider = getSdk().sdkLoggerProvider
+        return suspendCoroutine {
```
abdulraqeeb33, line 52: does this work while the app is being closed down by the OS? Does it have to run on the main thread?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/OneSignalOpenTelemetry.kt
```diff
@@ -0,0 +1,124 @@
...
+}
+
+@RequiresApi(Build.VERSION_CODES.O)
+internal class OneSignalOpenTelemetryRemote(
+    private val _configModelStore: ConfigModelStore,
+    _osTopLevelFields: OneSignalOtelFieldsTopLevel,
+    _osPerEventFields: OneSignalOtelFieldsPerEvent,
+) : OneSignalOpenTelemetryBase(_osTopLevelFields, _osPerEventFields),
+    IOneSignalOpenTelemetryRemote {
+    private val appId: String get() =
```
abdulraqeeb33, line 75: can we make this lazy so that it fetches on time and then stores it in the memory?

OneSignalSDK/onesignal/core/build.gradle
```diff
@@ -88,6 +88,18 @@ dependencies {
         }
     }
 
+    implementation platform("io.opentelemetry:opentelemetry-bom:1.55.0")
```
abdulraqeeb33, line 91: can we create variables for the versions numbers used here instead of hardcoding them here?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/IOneSignalCrashReporter.kt
jkasten2: I'll make this a new module, that core depends on. We can switch the module to KMP later.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/OneSignalCrashHandler.kt
jkasten2: This line of code is pretty fast, 1 to 10ms

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/crash/OneSignalCrashUploader.kt
jkasten2: We would have to apply the change to all startup services. We decided to skip this for now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/config/OtelConfigRemoteOneSignal.kt
jkasten2: Need to test and most likely need to handle this.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/restoration/impl/NotificationRestoreWorkManager.kt
jkasten2: This is ok, since it was required for Kotlin version update.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/OneSignalOpenTelemetry.kt
jkasten2: Double check that `suspendCoroutine` isn't making a new thread.

PR #2406, opened 2025-11-10
fix: login with the same externalId fails to update the subscription

jinliu9508:
# Description
## One Line Summary
Fix issue where calling login with the same externalId as the previous app open fails to update the subscription in the backend.

### Motivation
Previously, when login was called using the same externalId as the previous app open, the SDK detected no change in the externalId and skipped creating a login request. Instead, it issued a refresh user request.

However, if the user’s external ID had been manually removed from the backend (for example, via the OneSignal Dashboard or calling Logout outside of SDK), this refresh request no longer linked the subscription with the intended externalId. The result was that the backend subscription remained unassociated with the external ID after login.

This change ensures that a login request is always created when login is called — even if the same externalId is passed again — so that the backend subscription is correctly updated and re-linked to that external ID. The subscription ID associated to the OneSignalID will still be reused. 

### Scope
- Affects user identity handling and external ID login flow.
- No changes to other SDK modules (notifications, outcomes, or in-app messaging).
- Maintains backward compatibility with existing login APIs.

# Testing
## Unit testing
Added a test unit to ensure a Login request is always created

## Manual testing
Step to reproduce:
  1. On a fresh install, call Login with an externalID right after initWithContext
  2. Observe that the subscription is created with the externalID
  3. Close app and remove the externalID through the dashboard
  4. Open the app again, Login with the same externalID is called but the externalID is still missing for the subscription in backend.
  
After fix: The subscription has successfully restored the externalID after the second Login. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2406)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
@@ -25,10 +25,7 @@ class LoginHelper(
             currentIdentityExternalId = identityModelStore.model.externalId
             currentIdentityOneSignalId = identityModelStore.model.onesignalId
 
-            if (currentIdentityExternalId == externalId) {
-                return
-            }
-
+            // always create a user on login in case of externalId being removed in backend
             // TODO: Set JWT Token for all future requests.
             userSwitcher.createAndSwitchToNewUser { identityModel, _ ->
                 identityModel.externalId = externalId
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/LoginHelperTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/LoginHelperTests.kt
@@ -37,7 +37,9 @@ class LoginHelperTests : FunSpec({
         Logging.logLevel = LogLevel.NONE
     }
 
-    test("login with same external id returns early without creating user") {
+    // in the event of externalId being removed in the backend, SDK needs to send a CreateUser request
+    // to restore the externalId for the subscription
+    test("login with same external id still creates a user") {
         // Given
         val mockIdentityModelStore =
             MockHelper.identityModelStore { model ->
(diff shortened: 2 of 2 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
```diff
@@ -25,10 +25,7 @@ class LoginHelper(
             currentIdentityExternalId = identityModelStore.model.externalId
             currentIdentityOneSignalId = identityModelStore.model.onesignalId
 
-            if (currentIdentityExternalId == externalId) {
```
abdulraqeeb33, line 28: should we check if externalId is empty and then create the user on login?

jinliu9508: Good question, externalId will actually never be empty because it is required for the Login call. It is possible to have an empty currentIdentityExternalId which indicates an anonymous user. They can never be both empty when Login is called.

The scenario when they can both be empty is when the app opens for the second time and Login is never called.

abdulraqeeb33: Ah yes, its part of the signature of the method.

nan-li (changes requested):
Blocking while reviewing

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
abdulraqeeb33, line 28: This same behaviour exists in iOS, so we need to figure out what is it thats different we are doing.

is it possible that the login method is called so quickly and identityModelStore is not setup yet?

Issue #2192, opened 2024-09-14
[Bug]: Crashes when OneSignal is used immediately after initiWithContext called


fanwgwg:
### What happened?

In my code, I've consistently encountered crashes when using OneSignal APIs after calling OneSignal.initWithContext, specifically receiving the error: java.lang.Exception: Must call 'initWithContext' before use.

I believe this is related to the recent changes in [PR #2151](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2151), which moved initialization to a background thread. However, there seems to be no clear guidance from OneSignal on how to determine when the SDK has completed initialization and when it's safe to call other APIs.

Relying on developers to wait for an arbitrary period, such as 2 seconds, before invoking OneSignal APIs is not a viable or professional coding practice—this is not something we see in other SDKs. If the initialization is now asynchronous, then the rest of the APIs should also be asynchronous. For instance, APIs could return a result only after initialization has completed.

Overall, I feel the quality of the SDK urgently needs improvement, especially given the frequent crashes introduced since the 5.0 release. Addressing these issues with more robust handling of the initialization process is crucial.

### Steps to reproduce?

```Markdown
OneSignal.initWithContext(xxx);
OneSignal.getUser();
```

### What did you expect to happen?

The code should not crash

### OneSignal Android SDK version

com.onesignal:OneSignal:5.1.21

### Android version

14

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jinliu9508:
Hello @fanwgwg, thank you for the report. Could you share more of the stack trace regarding the crash? I have tested your code but was unable to reproduce the exception. Part of the initialization has been moved to the background; however, you can still successfully call methods like getUser() immediately after initWithContext(). The system allows components like User to be initialized even if the background initialization is not yet complete. In this case, the system will simply use the component that was initialized by your call.

I do notice that this error can occur in certain multi-threading environments that are not specifically covered in our documentation. If you could provide more details, including how you are calling OneSignal API methods and the stack trace when the crash occurs, I can assist in identifying the root cause of the issue.

jinliu9508:
@fanwgwg Are you still able to reproduce the exception if you call 'OneSignal.getUser().getPushSubscription().getOptedIn();' immediately after 'OneSignal.initWithContext()'? It seems that your ViewModelProvider might be initialized earlier than AppSettingsActivity.onCreate(), depending on how your injection works.

[...]

PR #2408, opened 2025-11-12
fix: Remove throwing "initWithContext was not called or timed out", introduced in 5.4.0

abdulraqeeb33:
# Description

## One Line Summary

Fix race condition in initWithContext preventing concurrent initialization and premature SDK access.

## Details

This PR fixes a race condition that could occur when `initWithContext()` was called multiple times concurrently, or when SDK accessors were accessed before initialization completed. The fix ensures thread-safe initialization state management and prevents inconsistent SDK state.

### The Problem

The race condition could manifest in two scenarios:

1. **Concurrent Initialization**: If multiple threads called `initWithContext()` simultaneously, both could attempt to initialize the SDK, leading to:
   - Duplicate service initialization
   - Inconsistent `initState` values
   - Potential crashes or undefined behavior

2. **Premature Access**: If SDK accessors (like `OneSignal.User`, `OneSignal.Notifications`, etc.) were accessed before `initWithContext()` completed, they could:
   - Access uninitialized services
   - Return null or throw `NullPointerException`
   - Cause ANRs if blocking occurred on the main thread

### The Solution

The fix implements proper synchronization and state management:

1. **Synchronized State Checks**: Added `synchronized(initLock)` around `initState` checks to prevent concurrent initialization attempts
2. **State-Based Guarding**: Early return if SDK is already initialized or initialization is in progress
3. **Consistent State Management**: `initState` is properly set to `IN_PROGRESS` before starting initialization and `SUCCESS`/`FAILED` after completion
4. **Proper Waiting Mechanism**: Accessors that need initialization wait using `CompletableDeferred` to ensure consistent state

### Related Issues

- Fixes race condition reported in [#2192](https://github.com/OneSignal/OneSignal-Android-SDK/issues/2192#issuecomment-3520217567)

### Motivation

To address the race condition where concurrent initialization attempts or premature SDK access could lead to crashes or inconsistent behavior.

### Scope

- Internal race condition handling in `OneSignalImp.initWithContext()`
- Thread-safe state management using `synchronized` blocks
- Proper state transitions during initialization lifecycle
- No public API changes - purely internal implementation improvement

# Testing

## Unit testing

- Existing tests continue to pass, verifying backward compatibility
- Race condition scenarios are implicitly tested through existing initialization tests
- State management is verified through `InitState` enum transitions

## Manual testing

- Tested concurrent `initWithContext()` calls from multiple threads
- Verified SDK accessors work correctly when called immediately after `initWithContext()`
- Confirmed no crashes or inconsistent state when initialization is in progress
- Smoke tested on various devices and Android versions

# Affected code checklist

- [] Notifications
  - [] Display (protected by initialization state checks)
  - [] Open (protected by initialization state checks)
  - [] Push Processing (protected by initialization state checks)
  - [] Confirm Deliveries (protected by initialization state checks)
- [] Outcomes (protected by initialization state checks)
- [] Sessions (protected by initialization state checks)
- [] In-App Messaging (protected by initialization state checks)
- [] REST API requests (initialization must complete before API calls)
- [ ] Public API changes (no breaking changes)

# Checklist

## Overview

- [x] PR does one thing - fixes race condition in initialization

## Testing

## Final pass

  - Proper synchronization blocks with clear comments
  - State management follows clear lifecycle patterns
  - Thread-safety considerations are documented

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2408)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
@@ -25,8 +25,10 @@ import java.util.concurrent.atomic.AtomicInteger
  * - Small bounded queues (10 tasks) to prevent memory bloat
  * - Reduced context switching overhead
  * - Efficient thread management with controlled resource usage
+ *
+ * Made public to allow mocking in tests via IOMockHelper.
  */
-internal object OneSignalDispatchers {
+object OneSignalDispatchers {
     // Optimized pool sizes based on CPU cores and workload analysis
     private const val IO_CORE_POOL_SIZE = 2 // Increased for better concurrency
     private const val IO_MAX_POOL_SIZE = 3 // Increased for better concurrency
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/startup/StartupServiceTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/startup/StartupServiceTests.kt
@@ -5,6 +5,7 @@ import com.onesignal.common.services.ServiceProvider
 import com.onesignal.debug.LogLevel
 import com.onesignal.debug.internal.logging.Logging
 import com.onesignal.mocks.IOMockHelper
+import com.onesignal.mocks.IOMockHelper.awaitIO
 import io.kotest.assertions.throwables.shouldThrowUnit
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.comparables.shouldBeLessThan
@@ -13,7 +14,6 @@ import io.mockk.every
 import io.mockk.mockk
 import io.mockk.spyk
 import io.mockk.verify
-import kotlinx.coroutines.CompletableDeferred
 
 class StartupServiceTests : FunSpec({
     fun setupServiceProvider(
@@ -83,33 +83,27 @@ class StartupServiceTests : FunSpec({
 
     test("startup will call all IStartableService dependencies successfully after a short delay") {
         // Given
-        val mockStartupService1 = spyk<IStartableService>()
-        val mockStartupService2 = spyk<IStartableService>()
+        val mockStartupService1 = mockk<IStartableService>(relaxed = true)
+        val mockStartupService2 = mockk<IStartableService>(relaxed = true)
 
         val startupService = StartupService(setupServiceProvider(listOf(), listOf(mockStartupService1, mockStartupService2)))
 
         // When
         startupService.scheduleStart()
 
-        // Then
-        Thread.sleep(10)
+        // Then - wait deterministically for both services to start using IOMockHelper
+        awaitIO()
         verify(exactly = 1) { mockStartupService1.start() }
         verify(exactly = 1) { mockStartupService2.start() }
     }
 
     test("scheduleStart does not block main thread") {
         // Given
-        val mockStartableService1 = spyk<IStartableService>()
+        val mockStartableService1 = mockk<IStartableService>(relaxed = true)
         val mockStartableService2 = spyk<IStartableService>()
         val mockStartableService3 = spyk<IStartableService>()
         // Only service1 and service2 are scheduled - service3 is NOT scheduled
         val startupService = StartupService(setupServiceProvider(listOf(), listOf(mockStartableService1, mockStartableService2)))
-
-        // Block service1 to prove scheduleStart() doesn't wait for it
-        val blockTrigger = CompletableDeferred<Unit>()
-        every { mockStartableService1.start() } coAnswers {
-            blockTrigger.await() // Block until released
-        }
 
         // When - scheduleStart() is async, so it doesn't block
         val startTime = System.currentTimeMillis()
@@ -129,15 +123,12 @@ class StartupServiceTests : FunSpec({
         // Verify service3 was called immediately (proving main thread wasn't blocked)
         verify(exactly = 1) { mockStartableService3.start() }
 
-        // Wait a bit for async execution to start
-        Thread.sleep(50)
+        // Wait deterministically for async execution using IOMockHelper
+        awaitIO()
 
-        // Verify scheduled services were called (even though service1 is blocked)
+        // Verify scheduled services were called
         verify(exactly = 1) { mockStartableService1.start() }
         verify(exactly = 1) { mockStartableService2.start() }
-
-        // Unblock service1 to allow test cleanup
-        blockTrigger.complete(Unit)
 
         // The key assertion: scheduleStart() returned immediately without blocking,
         // allowing service3.start() to be called synchronously before scheduled services
--- a/OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
+++ b/OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
@@ -1,7 +1,7 @@
 package com.onesignal.mocks
 
+import com.onesignal.common.threading.OneSignalDispatchers
 import com.onesignal.common.threading.suspendifyOnIO
-import com.onesignal.common.threading.suspendifyWithCompletion
 import io.kotest.core.listeners.AfterSpecListener
 import io.kotest.core.listeners.BeforeSpecListener
 import io.kotest.core.listeners.BeforeTestListener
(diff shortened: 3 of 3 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -306,22 +305,48 @@ internal class OneSignalImp(
...
+        when (initState) {
+            InitState.FAILED -> {
+                throw IllegalStateException("Initialization failed. Cannot proceed.")
+            }
+            InitState.NOT_STARTED -> {
+                throw IllegalStateException("Must call 'initWithContext' before 'logout'")
+            }
+            InitState.IN_PROGRESS, InitState.SUCCESS -> {
+                // Continue - these states allow proceeding (will wait if needed)
+            }
```
jkasten2: Duplicate code, can we make a function for this?

jkasten2 (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -306,22 +305,48 @@ internal class OneSignalImp(
     ) {
         Logging.log(LogLevel.DEBUG, "Calling deprecated login(externalId: $externalId, jwtBearerToken: $jwtBearerToken)")
 
-        if (!initState.isSDKAccessible()) {
-            throw IllegalStateException("Must call 'initWithContext' before 'login'")
+        // Check state and provide appropriate error messages
```
jinliu9508: Can isSDKAccessible() be changed to something like checkSDKAccessible() so that it throws proper error messages based on initState, and do nothing if allow proceeding?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -306,22 +305,48 @@ internal class OneSignalImp(
...
+            }
+            InitState.NOT_STARTED -> {
+                throw IllegalStateException("Must call 'initWithContext' before 'login'")
+            }
+            InitState.IN_PROGRESS, InitState.SUCCESS -> {
+                // Continue - these states allow proceeding (will wait if needed)
+            }
         }
 
         waitForInit()
```
jinliu9508: We are essentially checking init state before and after waitForInit(). Should we let waitForInit() handle the throwing/proceeding based on the initState so we don't need to check for the state again in the usage?

abdulraqeeb33: good callout.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
abdulraqeeb33: good callout, i centralized all of this. including the code path for suspend and blocking.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
```diff
@@ -150,8 +150,19 @@ class SDKInitTests : FunSpec({
...
         // Then
-        // should complete even SharedPreferences is unavailable
+        // should complete even SharedPreferences is unavailable (non-blocking)
         accessorThread.isAlive shouldBe false
+
+        // Release the SharedPreferences lock so internalInit can complete
+        trigger.complete()
+
+        // Wait for initialization to complete (internalInit runs asynchronously)
+        var attempts = 0
```
jinliu9508: nit: the wait mechanism for the initialization seems repetitive in the test. Can we clean this up?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -333,33 +324,73 @@ internal class OneSignalImp(
...
+                throw IllegalStateException(message)
             }
             InitState.IN_PROGRESS -> {
-                Logging.debug("Suspend waiting for init to complete...")
+                Logging.debug("Waiting for init to complete...")
                 try {
-                    withTimeout(MAX_TIMEOUT_TO_INIT) {
+                    withTimeout(timeoutMs) {
                         initAwaiter.awaitSuspend()
                     }
```
jkasten2: If we timeout then have an inconsistent state where waitUntilInitInternal returns, but initState == InitState.IN_PROGRESS. I would rather we just wait here as long as it takes so code doesn't have to consider both states. I don't see what upsides this complexity gives us.

PR #2412 does this, and only logs if it is taking longer than expected.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -69,6 +70,7 @@ internal object OneSignalDispatchers {
                     priority = Thread.NORM_PRIORITY - 1,
                     // Slightly lower priority for I/O tasks
                 ),
+                ThreadPoolExecutor.CallerRunsPolicy(), // Execute on calling thread if queue is full (prevents rejection)
```
jkasten2: This can lead to calling things on the main thread, which in turn will lead to ANRs. Also this creates more inconsistencies around timing between thread, which could lead to very hard to reproduce race conditions.

I would rather we continue adding and growing the queue and not reject.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -87,6 +89,7 @@ internal object OneSignalDispatchers {
                 TimeUnit.SECONDS,
                 LinkedBlockingQueue(QUEUE_CAPACITY),
                 OptimizedThreadFactory(DEFAULT_THREAD_NAME_PREFIX),
+                ThreadPoolExecutor.CallerRunsPolicy(), // Execute on calling thread if queue is full (prevents rejection)
```
jkasten2: This can lead to calling things on the main thread, which in turn will lead to ANRs. Also this creates more inconsistencies around timing between thread, which could lead to very hard to reproduce race conditions.

I would rather we continue adding and growing the queue and not reject.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -27,15 +27,16 @@ import java.util.concurrent.atomic.AtomicInteger
  * - Efficient thread management with controlled resource usage
  */
 internal object OneSignalDispatchers {
-    // Optimized pool sizes based on CPU cores and workload analysis
-    private const val IO_CORE_POOL_SIZE = 2 // Increased for better concurrency
-    private const val IO_MAX_POOL_SIZE = 3 // Increased for better concurrency
+    // Optimized pool thread counts to handle more concurrent operations during init
+    // (especially important now that we wait indefinitely, which may cause more operations to queue)
+    private const val IO_CORE_POOL_SIZE = 4 // Increased for better concurrency during init
+    private const val IO_MAX_POOL_SIZE = 6 // Increased to handle bursts of operations
```
jkasten2: We have seen a number of bugs in 5.4.0, I have concerns that this might be due to race conditions between threads. Can we leave these thread counts unchanged for now?
* I was thinking we may even want to lower these, or even go down to 1 thread for both IO and Default if we can't find the root causes of 5.4.0 issues.

abdulraqeeb33: I will leave it unchanged for now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: yeah, thats true. I was wondering if this would help offload work back to the caller while the threads free up to do the work. but it will impact the main thread right away.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
abdulraqeeb33: answered above

PR #2415, opened 2025-11-18
DROPPED - Revert/5.4.0 changes

jkasten2:
`git revert` was used in this PR to rollback changes temporary made for 5.4.0, however is not a good approach in this case. 
* This PR breaks the build and we don't want to rollback the CI/CD improvements we also made on main (such as lint checks) nor make one off commits for a temporary rollback.

Instead of this PR, a [5.4.1 release](https://github.com/OneSignal/OneSignal-Android-SDK/commit/3ac0157db57ac74b2741bc6f1f83b0a58690d3bb) was made on top of the 5.1.38 tag to do the same rollback effect.
* This way we can focus on moving forward with main being latest, apply fixes to it, and releasing 5.4.2 once all fixes are ready.

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2415)


michael-winkler:
A description explaining why this is not consistent is missing..

PS:
It might be time to look into joining an initiative.

PR #2416, opened 2025-11-20
tests: more tests for InAppMessagesManager and LocationManager

jinliu9508:
# Description
## One Line Summary
Add more tests for InAppMessagesManager and LocationManager to improve test coverage.

## Production Code Change
All usages of Build.VERSION.SDK_INT in LocationManager.kt have been refactored to use AndroidUtils.androidSDKInt for better flexibility when mocking the Android API level in tests.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2416)

### DIFF
--- a/OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
+++ b/OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
@@ -87,14 +87,6 @@ object IOMockHelper : BeforeSpecListener, AfterSpecListener, BeforeTestListener,
         // ThreadUtilsKt = file that contains suspendifyOnIO
         mockkStatic(THREADUTILS_PATH)
 
-        every {
-            suspendifyWithCompletion(
-                useIO = any(),
-                block = any<suspend () -> Unit>(),
-                onComplete = any()
-            )
-        } answers { callOriginal() }
-
         every { suspendifyOnIO(any<suspend () -> Unit>()) } answers {
             val block = firstArg<suspend () -> Unit>()
 


OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
```diff
@@ -0,0 +1,625 @@
...
+    }
+})
+
+// Helper function to set SDK version using reflection
+private fun setSdkVersion(sdkInt: Int) {
+    try {
+        val buildVersionClass = Class.forName("android.os.Build\$VERSION")
+        val sdkIntField = buildVersionClass.getDeclaredField("SDK_INT")
+        sdkIntField.isAccessible = true
+        sdkIntField.setInt(null, sdkInt)
```
jkasten2: nit, couldn't we use mockk to simply this code?

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
```diff
@@ -0,0 +1,625 @@
...
+    } catch (e: Exception) {
+        // If reflection fails, the test will use the default SDK version
+        // This is acceptable for tests that don't strictly require a specific SDK version
+    }
```
jkasten2: I would rather throw so the test fails, rather than the test passing but not testing what it should be.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    private var ioWaiter: CompletableDeferred<Unit> = CompletableDeferred()
+
+    /**
+     * Wait for the current suspendifyOnIO work to finish.
+     * Can be called from tests instead of delay/Thread.sleep.
+     */
+    fun awaitIO() {
+        if (!ioWaiter.isCompleted) {
+            runBlocking {
```
jkasten2: We can omit this `runBlocking` and then make this function `suspend`, the tests support calling `suspend` functions.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+ *
+ *     test("xyz") {
+ *         iamManager.start()    // start() calls suspendOnIO
+ *         awaitIO()    // wait for background work deterministically
+ *         ...
+ *     }
+ */
+object IOMockHelper : BeforeSpecListener, AfterSpecListener, BeforeTestListener, TestListener {
+
+    private var ioWaiter: CompletableDeferred<Unit> = CompletableDeferred()
```
jkasten2, line 53: Couldn't there be multiple `suspendifyOnIO` and `suspendifyWithCompletion` calls that we could be waiting on?

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+        ioWaiter = CompletableDeferred()
+    }
+
+    override suspend fun beforeSpec(spec: Spec) {
+        // ThreadUtilsKt = file that contains suspendifyOnIO
+        mockkStatic("com.onesignal.common.threading.ThreadUtilsKt")
+
+        every { suspendifyOnIO(any<suspend () -> Unit>()) } answers {
```
jkasten2, line 90: Don't we need to also mock `suspendifyWithCompletion` as well?

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    override suspend fun beforeTest(testCase: TestCase) {
+        // fresh waiter for each test
+        ioWaiter = CompletableDeferred()
+    }
+
+    override suspend fun afterSpec(spec: Spec) {
+        unmockkStatic("com.onesignal.common.threading.ThreadUtilsKt")
```
jkasten2: If we really can't use a direct function reference and have to use a string let's make this a const we can reference in both places. (unless somehow Kotlin can compile time check this string)

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
```diff
@@ -0,0 +1,83 @@
...
+ *     listener(IOMockHelper)
+ *   ...
+ *
+ *     test("xyz") {
+ *         iamManager.start()    // start() calls suspendOnIO
+ *         awaitIO()    // wait for background work deterministically
+ *         ...
+ *     }
+ */
+object IOMockHelper : BeforeSpecListener, AfterSpecListener, BeforeTestListener, TestListener {
```
jkasten2, line 44: There are some limitations of this helper that either should addressed or noted in the comments of the limitations if we think they would rarely be needed: (Some of these are also noted on the lines below)
1. `suspendifyWithCompletion` isn't mocked here.
2. Doesn't support multiple things being waiting on.
3. Doesn't support calls to `defaultExecutor`

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
```diff
@@ -1,80 +1,1163 @@
...
+    val identityModelStore = MockHelper.identityModelStore()
+    val pushSubscription = mockk<IPushSubscription>(relaxed = true)
+    val outcomeEventsController = mockk<IOutcomeEventsController>(relaxed = true)
+    val state = mockk<InAppStateService>(relaxed = true)
+    val prefs = mockk<IInAppPreferencesController>(relaxed = true)
+    val repository = mockk<IInAppRepository>(relaxed = true)
+    val backend = mockk<IInAppBackendService>(relaxed = true)
+    val triggerController = mockk<ITriggerController>(relaxed = true)
+    val triggerModelStore = mockk<TriggerModelStore>(relaxed = true)
+    val inAppDisplay = mockk<IInAppDisplayer>(relaxed = true)
```
jkasten2: nit, I would call this `inAppDisplayer` instead of `inAppDisplay` to match the interface name a bit better.

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
```diff
@@ -1,80 +1,1163 @@
...
+    val lifecycle = mockk<IInAppLifecycleService>(relaxed = true)
+    val languageContext = MockHelper.languageContext()
+    val time = MockHelper.time(1000)
+    val inAppMessageLifecycleListener = spyk<IInAppMessageLifecycleListener>()
+    val inAppMessageClickListener = spyk<IInAppMessageClickListener>()
+    val rywData = RywData("token", 100L)
+
+    val deferred = mockk<CompletableDeferred<RywData?>>() {
```
jkasten2: `deferred` is a bit too generic in this context, would at least add `ryw` to this variable name. (`rywDeferred`)

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
```diff
@@ -1,80 +1,1163 @@
...
+    // mock default services needed for InAppMessagesManager
+    val applicationService = MockHelper.applicationService()
+    val sessionService = mockk<ISessionService>(relaxed = true)
+    val influenceManager = mockk<IInfluenceManager>(relaxed = true)
+    val configModelStore = MockHelper.configModelStore()
+    val userManager = mockk<IUserManager>(relaxed = true)
+    val identityModelStore = MockHelper.identityModelStore()
+    val pushSubscription = mockk<IPushSubscription>(relaxed = true)
+    val outcomeEventsController = mockk<IOutcomeEventsController>(relaxed = true)
+    val state = mockk<InAppStateService>(relaxed = true)
```
jkasten2: nit, `state` is a bit too generic in this context.

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
```diff
@@ -1,80 +1,1163 @@
...
+            every { mockTriggerModelStore.get("new-key") } returns null
+            every { mockTriggerModelStore.add(capture(triggerModelSlots)) } answers {}
+
+            // When
+            mocks.inAppMessagesManager.addTrigger("new-key", "new-value")
+
+            // Then
+            triggerModelSlots.size shouldBe 1
+            triggerModelSlots[0].key shouldBe "new-key"
+            triggerModelSlots[0].value shouldBe "new-value"
```
jkasten2: I believe this could be simplified into one `shouldBe` statement.

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
```diff
@@ -0,0 +1,608 @@
...
+
+    fun set_coarse_location_permission(granted: Boolean) {
+        every {
+            AndroidUtils.hasPermission(
+                LocationConstants.ANDROID_COARSE_LOCATION_PERMISSION_STRING,
+                true,
+                mockAppService,
+            )
+        } returns granted
+    }
```
jkasten2: We don't use snake case in our Kotlin code base.

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
```diff
@@ -0,0 +1,608 @@
...
+
+        test("requestPermission prompts for fine permission when not granted and in manifest") {
+            // Set SDK version to 23 using reflection
+            setSdkVersion(23)
+
+            // Verify SDK version was set (if reflection fails, skip this test)
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+                // SDK version couldn't be set, skip this test
+                return@test
+            }
```
jkasten2: Why skip the test? If it doesn't fail then what would get our attention to fix this tests?

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
```diff
@@ -0,0 +1,608 @@
...
+
+        test("requestPermission prompts for coarse permission when fine not in manifest") {
+            // Set SDK version to 23 using reflection
+            setSdkVersion(23)
+
+            // Verify SDK version was set (if reflection fails, skip this test)
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+                // SDK version couldn't be set, skip this test
+                return@test
+            }
```
jkasten2: Why skip the test? If it doesn't fail then what would get our attention to fix this tests?

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
```diff
@@ -0,0 +1,608 @@
...
+    context("requestPermission() Method - API >= 29 (Android 10+)") {
+        test("requestPermission prompts for background permission when fine granted but background not") {
+            // Set SDK version to 29 using reflection
+            setSdkVersion(29)
+
+            // Verify SDK version was set (if reflection fails, skip this test)
+            if (Build.VERSION.SDK_INT < 29) {
+                // SDK version couldn't be set, skip this test
+                return@test
+            }
```
jkasten2: Why skip the test? If it doesn't fail then what would get our attention to fix this tests?

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
```diff
@@ -1,80 +1,1163 @@
...
+import org.json.JSONArray
+import org.json.JSONObject
+
+private class Mocks {
+    // mock default services needed for InAppMessagesManager
+    val applicationService = MockHelper.applicationService()
+    val sessionService = mockk<ISessionService>(relaxed = true)
+    val influenceManager = mockk<IInfluenceManager>(relaxed = true)
+    val configModelStore = MockHelper.configModelStore()
+    val userManager = mockk<IUserManager>(relaxed = true)
```
abdulraqeeb33, line 64: are we ok with using relaxed = true for all these mocks?
i would go as far as to use the non relaxed instance.
run the tests and then mock the method that needs attention.

jkasten2: Using `relaxed` is fine, it just depends on the logic being tested. I think in this case most of the tests are fine with `relaxed`, since they are just ensuring something is called on those objects.

jinliu9508: @abdulraqeeb33 All we need from these services is to verify if a certain method is called from them. I think using relaxed = true is simpler than construct the whole instance that we don't really need.

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
jinliu9508: Mockk cannot mock final static fields like Build.VERSION.SDK_INT, so every { Build.VERSION.SDK_INT } returns sdkInt fails. In the follow up commit I added a helper function that returns SDK_Int in AndroidUtils so we can easily mock to test behaviors under different api level.

Ideally we can refactor all Build.VERSION usages to use this helper function, but that will be in other PRs when adding testing for other classes.

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
jinliu9508: changed to inAppStateService to match the type

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
jinliu9508: fixed

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
jinliu9508: I think it doesn't matter, but changed to a key-value style verification

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
jinliu9508: fixed

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
jinliu9508: fixed

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
jinliu9508: fixed

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
jinliu9508: fixed

OneSignalSDK/onesignal/location/src/test/java/com/onesignal/location/internal/LocationManagerTests.kt
jinliu9508: fixed

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jinliu9508: Yeah I think most of the time we only test one function at a time in each test, but to prevent confusion / test failure in the future, IOMockHelper now supports awaiting multiple IO tasks. 

1. If another IO task is added before the first task completes, awaitIO() will suspend until all tasks are completed.
2. If another IO task is added after the first task completes, awaitIO() can still be called to wait for the second task.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jinliu9508: addressed

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jinliu9508: Yes, that's correct, good catch. Addressed in the follow up commit.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jinliu9508: I think mockkStatic() doesn’t magically stub everything in that file – it just allows those functions to be mocked. Anything we don’t explicitly stub still calls the real implementation. That's probably the reason why it works without mocking suspendifyWithCompletion. But to prevent any future behavior change / be extra safe, I mocked with callOriginal() to keep the original implementation.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jinliu9508: Good point, the string is now save as a const.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jkasten2: I had a miss understand of how `suspendifyWithCompletion` is used at first. There are a few layers here with `ThreadUtils.kt` and `OneSignalDispatchers`, their parts are used directly to sometimes like `launchOnIO`. We should cover them in this helper or at least make a comment that this doesn't cover them all.

OneSignalSDK/onesignal/in-app-messages/src/test/java/com/onesignal/inAppMessages/internal/InAppMessagesManagerTests.kt
jkasten2: I was thinking something like this:
```kotlin
triggerModelSlots shouldBe arrayOf(mapOf("new-key" to "new-value"))
```

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
jinliu9508: Added a comment on awaitIO to address the coverage.

jinliu9508 (commented):
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2416#discussion_r2561895138
triggerModelSlots is a list of TriggerModel captured when add() is called by the triggerModelStore so we can't directly compare with a list of maps. I updated a new version for better readability.

OneSignalSDK/onesignal/testhelpers/src/main/java/com/onesignal/mocks/IOMockHelper.kt
```diff
@@ -0,0 +1,131 @@
...
+        // ThreadUtilsKt = file that contains suspendifyOnIO
+        mockkStatic(THREADUTILS_PATH)
+
+        every {
+            suspendifyWithCompletion(
+                useIO = any(),
+                block = any<suspend () -> Unit>(),
+                onComplete = any()
+            )
+        } answers { callOriginal() }
```
jkasten2: Remove this code as it doesn't do anything.

jinliu9508: removed

jinliu9508:
CI fails due to a flaky test in StartupService. Will address it in a separate PR.

PR #2419, opened 2025-11-21
fix: Rare User and Subscription creates and updates processing out of order (introduced in 5.4.0)

jkasten2:
# Description
## One Line Summary
Fix rare bugs with User and Subscription creates and updates processing out of order that was introduced in 5.4.0.

## Details
Made a new internal coroutine scope with a single thread to the OperationRepo to ensure operations are alway executed in the order they are queued in.

This mostly fixes the flaky OperationRepoTest class.

Release 5.4.0 may have introduced bugs related to this, however they would have been rare and very hard to reproduce.
* Pre-5.4.0 there was also a rarer chance that usages of `enqueueAndWait` could have been out of order, it wasn't being put on the same scope as `enqueue`.

### Motivation
The original motivation was to fix flaky OperationRepoTest tests, however the underlaying cause is that the OperationRepo sometimes processes operations out of order.

### Scope
Only effects operation repo logic.

# Testing
## Unit testing
No new tests added, existing coverage already.

## Manual testing
Tested on an Android 14 and Android 6 emulator, ensured a new User and Subscription gets created.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2419)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -1,7 +1,6 @@
 package com.onesignal.core.internal.operations.impl
 
 import com.onesignal.common.threading.WaiterWithValue
-import com.onesignal.common.threading.suspendifyOnIO
 import com.onesignal.core.internal.config.ConfigModelStore
 import com.onesignal.core.internal.operations.ExecutionResult
 import com.onesignal.core.internal.operations.GroupComparisonType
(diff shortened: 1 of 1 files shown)


abdulraqeeb33 (commented):
Instead of creating a new coroutine scope, should we instead create a new thread just for OperationRepo tasks? just as we have Default and IO, maybe another as Operations and use that here?

PR #2500, opened 2025-12-04
Fix: NPE in SyncJobService since 5.4

jinliu9508:
# Description
## One Line Summary
Fix NPE from SyncJobService introduced on 5.4.0 caused by incompleted / failed OneSignal initialization during onStartJob.

### Motivation
Address the new NPE introduced on 5.4.0. 

### Scope
After 5.4, onStartJob calls initWithContext in background. That introduced a small possibility when onStopJob is called before the initialization completes in the background. When that happens, OneSignal.getService will run into a NPE. In that case, we just need to omit the job and do not reschedule.

# Testing
## Unit testing
Add several test cases covering SyncJobService.kt

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2500)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/services/SyncJobServiceTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/services/SyncJobServiceTests.kt
@@ -71,6 +71,19 @@ class SyncJobServiceTests : FunSpec({
         verify { syncJobService.jobFinished(jobParameters, false) }
     }
 
+    test("onStartJob calls jobFinished with false when initWithContext failed") {
+        // Given
+        val syncJobService = mocks.syncJobService
+        val jobParameters = mocks.jobParameters
+        coEvery { OneSignal.initWithContext(any()) } returns false
+
+        // When
+        syncJobService.onStartJob(jobParameters)
+
+        // Then
+        verify { syncJobService.jobFinished(jobParameters, false) }
+    }
+
     test("onStartJob calls jobFinished with false when needsJobReschedule is false") {
         // Given
         val syncJobService = mocks.syncJobService


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
```diff
@@ -38,7 +38,6 @@ class SyncJobService : JobService() {
         suspendifyOnIO {
             // init OneSignal in background
             if (!OneSignal.initWithContext(this)) {
-                jobFinished(jobParameters, false)
```
jkasten2, line 41: We must always call `jobFinished`, otherwise the system will not release the wakelock on the app. See Android docs:
https://developer.android.com/reference/android/app/job/JobService#onStartJob(android.app.job.JobParameters)

This logic looks like it would benefit from a `finally` block, calling the `jobFinished` from there, so it is clear it must always be called.

jinliu9508: That's right. We didn't call jobFinished before 5.4 but now onStartJob always returning true tells the system to hold until onStopJob is called. Addressed in the follow up commit.

jkasten2: > ... tells the system to hold until onStopJob is called

Until `jobFinished` you mean.

Can we also add a new test for this change?

jkasten2 (changes requested):
To be clear, this PR address two 5.4.0 bugs then:
1. Throwing on `onStopJob`
2. Not closing the SyncJobService wakelock if the OneSignal initialization fails.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
jinliu9508: Yes, a new test case is now added

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/services/SyncJobService.kt
```diff
@@ -36,31 +36,51 @@ import com.onesignal.debug.internal.logging.Logging
...
+         * After 5.4, onStartJob calls initWithContext in background. That introduced a small possibility
+         * when onStopJob is called before the initialization completes in the background. When that happens,
+         * OneSignal.getService will run into a NPE. In that case, we just need to omit the job and do not
+         * reschedule.
+         */
+
+        // Additional hardening in the event of getService failure
+        try {
+            // We assume init has been called via onStartJob\
+            val backgroundService = OneSignal.getService<IBackgroundManager>()
```
abdulraqeeb33, line 77: whats the guarantee that this service would be instantiated and ready to consume?
its more a general question not particularly for this PR.

jinliu9508: OneSignal.getService will instantiate the service if the service isn't available. This requires initWithContext to complete successfully.

Issue #2413, opened 2025-11-18
[Bug]: Model.getStringProperty java.lang.NullPointerException - null cannot be cast to non-null type kotlin.String


ningkanggithub:
### What happened?

          Non-fatal Exception: java.lang.NullPointerException: null cannot be cast to non-null type kotlin.String
       at com.onesignal.common.modeling.Model.getStringProperty(Model.kt:487)
       at com.onesignal.common.modeling.Model.getStringProperty$default(Model.kt:484)
       at com.onesignal.user.internal.properties.PropertiesModel.getOnesignalId(PropertiesModel.kt:12)
       at com.onesignal.user.internal.operations.impl.listeners.PropertiesModelStoreListener.getUpdateOperation(PropertiesModelStoreListener.kt:47)
       at com.onesignal.user.internal.operations.impl.listeners.PropertiesModelStoreListener.getUpdateOperation(PropertiesModelStoreListener.kt:13)
       at com.onesignal.core.internal.operations.listeners.SingletonModelStoreListener.onModelUpdated(SingletonModelStoreListener.kt:54)
       at com.onesignal.common.modeling.SingletonModelStore$onModelUpdated$1.invoke(SingletonModelStore.kt:66)
       at com.onesignal.common.modeling.SingletonModelStore$onModelUpdated$1.invoke(SingletonModelStore.kt:66)
       at com.onesignal.common.events.EventProducer.fire(EventProducer.kt:50)
       at com.onesignal.common.modeling.SingletonModelStore.onModelUpdated(SingletonModelStore.kt:66)
       at com.onesignal.common.modeling.ModelStore$onChanged$1.invoke(ModelStore.kt:96)
       at com.onesignal.common.modeling.ModelStore$onChanged$1.invoke(ModelStore.kt:96)
       at com.onesignal.common.events.EventProducer.fire(EventProducer.kt:50)
       at com.onesignal.common.modeling.ModelStore.onChanged(ModelStore.kt:96)
       at com.onesignal.common.modeling.Model$notifyChanged$1.invoke(Model.kt:662)
       at com.onesignal.common.modeling.Model$notifyChanged$1.invoke(Model.kt:662)
       at com.onesignal.common.events.EventProducer.fire(EventProducer.kt:50)
       at com.onesignal.common.modeling.Model.notifyChanged(Model.kt:662)
       at com.onesignal.common.modeling.Model.setOptAnyProperty(Model.kt:456)
[...]

PR #2505, opened 2025-12-12
Fix: NPE when setting timezone on app focus

jinliu9508:
# Description
## One Line Summary
Fix NPE from setting timezone on app focus when onesignalId is missing from the property model.

### Motivation
Fix NPE from setting timezone on app focus when onesignalId is missing from the property model.

This change was added in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2378

### Scope
Update the timezone in onSessionStart and onSessionActive instead of in UserManager.onFocus, since the propertiesModel may not be fully initialized at that time.

### Related Issues
https://github.com/OneSignal/OneSignal-Android-SDK/issues/2413

# Testing
## Unit testing
Add a test case for updating the timezone when onSessionStart() and onSessionActive is called.

## Manual testing
Step to reproduce:
1. Call initWithContext without an app ID. The app crashes on startup.
2. Reopen the app, observe the error
image
After fix this error does not show up anymore.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2505)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
@@ -260,8 +260,4 @@ internal open class UserManager(
             }
         }
     }
-
-    override fun onFocus(firedOnSubscribe: Boolean) { }
-
-    override fun onUnfocused() { }
 }
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/UserManagerTests.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/UserManagerTests.kt
@@ -26,7 +26,7 @@ class UserManagerTests : FunSpec({
         every { languageContext.language = capture(languageSlot) } answers { }
 
         val userManager =
-            UserManager(mockSubscriptionManager, MockHelper.identityModelStore(), MockHelper.propertiesModelStore(), languageContext, MockHelper.applicationService())
+            UserManager(mockSubscriptionManager, MockHelper.identityModelStore(), MockHelper.propertiesModelStore(), languageContext)
 
         // When
         userManager.setLanguage("new-language")
@@ -44,7 +44,7 @@ class UserManagerTests : FunSpec({
             }
 
         val userManager =
-            UserManager(mockSubscriptionManager, identityModelStore, MockHelper.propertiesModelStore(), MockHelper.languageContext(), MockHelper.applicationService())
+            UserManager(mockSubscriptionManager, identityModelStore, MockHelper.propertiesModelStore(), MockHelper.languageContext())
 
         // When
         val externalId = userManager.externalId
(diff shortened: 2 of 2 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
```diff
@@ -268,7 +268,10 @@ internal open class UserManager(
 
     override fun onFocus(firedOnSubscribe: Boolean) {
         // Detect any user properties updates that changed
-        _propertiesModel.timezone = TimeUtils.getTimeZoneId()
+        // Skip updating timezone when onesignalId is not set in the property model
+        if (_propertiesModel.hasProperty(PropertiesModel::onesignalId.name)) {
```
abdulraqeeb33: This seems to be decent workaround to prevent the crash. And this method will be called frequently and the timezone could be updated later when the oneSignaId is actually available

abdulraqeeb33 (approved):

abdulraqeeb33:
> This guard works but there's a deeper issue. There's no clear signal in the code that setting timezone requires onesignal ID to exist first. This is the kind of implicit dependency that's impossible to catch without already knowing about it.

Good catch on this @nan-li . i hope the new fix that we discussed will address this issue.

abdulraqeeb33 (approved):
the newer approach looks fine to me, but the tests are broken @jinliu9508

jinliu9508:
@abdulraqeeb33 
> the newer approach looks fine to me, but the tests are broken @jinliu9508

The test is now removed because onFocus no longer updates timezone

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
```diff
@@ -266,10 +265,7 @@ internal open class UserManager(
         }
     }
 
-    override fun onFocus(firedOnSubscribe: Boolean) {
-        // Detect any user properties updates that changed
-        _propertiesModel.timezone = TimeUtils.getTimeZoneId()
-    }
+    override fun onFocus(firedOnSubscribe: Boolean) { }
```
nan-li: See https://github.com/OneSignal/OneSignal-Android-SDK/pull/2378/changes, we can get rid of the ApplicationService stuff

nan-li (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
abdulraqeeb33: @nan-li  please check, i removed the references.

PR #2506, opened 2025-12-15
Improvement: failure message shows that appID is missing

jinliu9508:
# Description
## One Line Summary
Update the failure message so it is clear when appID is previously not provided.

### Motivation
The original failure message "Initialization failed. Cannot proceed." is vague and can potentially hide the real cause when appID is not provided as expected. We want to show this message so developers can immediately figure out the wrong usage.

### Scope
OneSignalImp.kt will now save and throw a clearer error message when OneSignal is accessed before initWithContext is called with an appID. 

### Comparison with 5.1.x behavior
5.1.x:
* Calling initWithContext without an appId does not throw immediately.
  * The first time we access OneSignal afterward, we get:
`Caused by: java.lang.Exception: Must call 'initWithContext' before use`
  * This points to the accessor call site, but it hides the real root cause: the missing appId.

5.4.x:
* The internal initWithContext(context) is now suspend and must be called from a coroutine.
  1. If a consumer migrates to 5.4.x and still calls the internal initWithContext directly on the main thread, they’ll get a compile-time error, which makes the incorrect usage obvious.
  2. Initialization is generally performed in a “fire-and-forget” coroutine. When the IO coroutine fails inside the SDK, there is no longer a meaningful JVM call stack back to the original caller (e.g., which Activity or method triggered init). It’s a separate async execution.
    * In this case, the error message is:
`suspendInitInternal: no appId provided or found in local storage. Please pass a valid appId to initWithContext().`
    * This message clearly states the root cause (no appId supplied), but the consumer will need to locate where they call init in their own code to fix it.

# Testing
## Unit testing
Add a new test case to assert the error message. 

## Manual testing
Tested on Emulator Pixel 9 API 35:
 1. Remove the appID and call OneSignal.initWithContext(context) in MainApplicationKT.kt
 3. Fresh install and observe the error message asking to provide an appID.
image

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2506)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -282,7 +282,7 @@ internal class OneSignalImp(
         val startupService = bootstrapServices()
         val result = resolveAppId(appId, configModel, preferencesService)
         if (result.failed) {
-            val message = "No OneSignal appId provided or found in local storage. Please pass a valid appId to initWithContext()."
+            val message = "suspendInitInternal: no appId provided or found in local storage. Please pass a valid appId to initWithContext()."
             val exception = IllegalStateException(message)
             // attach the real crash cause to the init failure exception that will be throw shortly after
             initFailureException?.addSuppressed(exception)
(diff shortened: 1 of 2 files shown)


jkasten2 (changes requested):
Can we also include the original stacktrace? This will make it easier to find which call to `initWithContext` was the cause of their issue.

Did you test how this scenario behaved in 5.1.x? Can you make a note of that in the PR?

abdulraqeeb33 (approved):

PR #2512, opened 2026-01-02
fix: send receive receipt even when preventDefault is called

nan-li:
# Description
## One Line Summary
Always send receive receipt even when preventDefault is called.

## Details
- Previously, when preventDefault() was called in INotificationLifecycleListener, the receive receipt was not sent to the server because notificationReceived() was never called for non-displayed notifications. 
- This change ensures the receive receipt is always sent (so the server knows the device received the notification), while influence tracking and analytics are only recorded when the notification is actually displayed.

### Motivation
Client use case

### Scope
Always send receive receipts regardless of display, matches iOS behavior

# Testing
## Unit testing
Updated

## Manual testing
API 35 emulator

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [x] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [ ] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2512)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/generation/impl/NotificationGenerationProcessor.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/generation/impl/NotificationGenerationProcessor.kt
@@ -235,8 +235,8 @@ internal class NotificationGenerationProcessor(
             markNotificationAsDismissed(notificationJob)
         }
 
-        // Always call notificationReceived, but pass wasDisplayed to control influence/analytics tracking
-        _lifecycleService.notificationReceived(notificationJob, wasDisplayed)
+        // Always call notificationReceived, regardless of wasDisplayed
+        _lifecycleService.notificationReceived(notificationJob)
     }
 
     // Saving the notification provides the following:
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
@@ -96,14 +96,8 @@ internal class NotificationLifecycleService(
         return canReceive
     }
 
-    override suspend fun notificationReceived(notificationJob: NotificationGenerationJob, wasDisplayed: Boolean) {
-        // Always send receive receipt regardless of whether notification was displayed
+    override suspend fun notificationReceived(notificationJob: NotificationGenerationJob) {
         _receiveReceiptWorkManager.enqueueReceiveReceipt(notificationJob.apiNotificationId)
-
-        // Only track influence and analytics if the notification was actually displayed
-        if (!wasDisplayed) {
-            return
-        }
 
         _influenceManager.onNotificationReceived(notificationJob.apiNotificationId)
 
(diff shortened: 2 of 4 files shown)


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/generation/impl/NotificationGenerationProcessor.kt
```diff
@@ -233,10 +233,10 @@ internal class NotificationGenerationProcessor(
             // Notification channel disable or not displayed
             // save notification as dismissed to avoid user re-enabling channel and notification being displayed due to restore
             markNotificationAsDismissed(notificationJob)
-            return
         }
 
-        _lifecycleService.notificationReceived(notificationJob)
+        // Always call notificationReceived, but pass wasDisplayed to control influence/analytics tracking
+        _lifecycleService.notificationReceived(notificationJob, wasDisplayed)
```
abdulraqeeb33: 💯

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
```diff
@@ -96,9 +96,15 @@ internal class NotificationLifecycleService(
...
+    override suspend fun notificationReceived(notificationJob: NotificationGenerationJob, wasDisplayed: Boolean) {
+        // Always send receive receipt regardless of whether notification was displayed
         _receiveReceiptWorkManager.enqueueReceiveReceipt(notificationJob.apiNotificationId)
 
+        // Only track influence and analytics if the notification was actually displayed
+        if (!wasDisplayed) {
+            return
+        }
+
         _influenceManager.onNotificationReceived(notificationJob.apiNotificationId)
```
abdulraqeeb33, line 102: @nan-li  is the influence and analytics related to displaying or just receiving the notification?

abdulraqeeb33 (approved):

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
nan-li: Influence is related to notifications influencing outcomes such as this notification was in the tray when something happened such as a click event. Analytics is for the firebase analytics where we send events like notification received. It's debatable if either of these should be tracked in this case but I think influence shouldn't happen. However, I checked iOS and **we do both of these** for a notification that doesn't display.

Only sending the confirmed delivery would be the minimal change, but to align with iOS, we would track more for the notification, although the influence component doesn't seem correct.

abdulraqeeb33: i think we should keep both iOS and Android consistent in what ever we do.

abdulraqeeb33 (approved):

PR #2520, opened 2026-01-13
fix: end initialization early if device storage is locked

jinliu9508:
# Description
## One Line Summary
Add isDeviceStorageUnlocked check to avoid accessing user data before the device is unlocked on Android 7.0+, and end initialization early if device storage is locked.

### Motivation
This allows the SDK to avoid reading/writing user data too early (e.g., in boot receivers or early background services), and to avoid crashes with the inconsistent state.

### Scope
Terminate the initialization early depends on user storage availability.

# Testing
## Unit testing
Two test cases are added.

## Manual testing
Tested on Emulator Pixel 9 API 35
Step to reproduce:
1. Set a PIN for the device
2. Modify AndroidManifest.xml
`<receiver android:name="com.onesignal.notifications.receivers.BootUpReceiver" android:exported="true" android:directBootAware="true"> <intent-filter> <action android:name="android.intent.action.BOOT_COMPLETED" /> <action android:name="android.intent.action.QUICKBOOT_POWERON" /> <action android:name="android.intent.action.LOCKED_BOOT_COMPLETED"/> </intent-filter> </receiver>`
3. Run the app once
4. Restart the device
5. Before unlocking the phone, run the command to capture the log
  adb logcat | grep -E "(SharedPreferences|BootUpReceiver)"
  
After fix:
The exception is no longer thrown in the background. After the user has unlocked the device, notifications can be received and interacted as normal even the app is not opened. 

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2520)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -277,6 +277,16 @@ internal class OneSignalImp(
         context: Context,
         appId: String?,
     ): Boolean {
+        // Check whether current Android user is accessible.
+        // Return early if it is inaccessible, as we are unable to complete initialization without access
+        // to device storage like SharedPreferences.
+        if (!AndroidUtils.isAndroidUserUnlocked(context)) {
+            Logging.warn("initWithContext called when device storage is locked, no user data is accessible!")
+            initState = InitState.FAILED
+            notifyInitComplete()
+            return false
+        }
+
         initEssentials(context)
 
         val startupService = bootstrapServices()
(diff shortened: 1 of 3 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -277,6 +277,16 @@ internal class OneSignalImp(
...
     ): Boolean {
+        // Check whether current Android user is accessible.
+        // Return early if it is inaccessible, as we are unable to complete initialization without access
+        // to device storage like SharedPreferences.
+        if (!AndroidUtils.isAndroidUserUnlocked(context)) {
+            Logging.warn("initWithContext called when device storage is locked, no user data is accessible!")
+            initState = InitState.FAILED
+            notifyInitComplete()
+            return false
+        }
```
jinliu9508, line 288: This part can be in a separate function

PR #2523, opened 2026-01-20
fix: SDK stuck in bad state when calling login with restricted external IDs

jinliu9508:
# Description
## One Line Summary
Treat restricted/invalid External ID login responses as non-retryable failures and does not need further handling so the SDK doesn’t pause the operation repo to get stuck in a bad state.

### Motivation
We reproduced an issue where calling OneSignal.login() with a restricted / blocked External ID can leave the SDK in a “bad” state where subsequent operations stop working until reinstall. The root cause is that the network response is being mapped to FAIL_PAUSE_OPREPO, which pauses the operation repository and effectively prevents recovery.

This PR changes the execution mapping so invalid (restricted) responses are treated as a non-retryable failure (FAIL_INVALID_LOGIN) that don't retry rather than pausing the op repo. This allows the SDK to continue functioning and lets the app recover (e.g., by retrying with a different External ID or after logout).

* Why don't we map it to FAIL_NO_RETRY? Because when we get a FAIL_NO_RETRY, we still try to recover by creating the user. By mapping the response to FAIL_INVALID_LOGIN, we completely disregard the bad request that will surely be rejected.

### Scope
* Affected: Login-related operations that receive an INVALID response status (e.g., restricted/blocked External IDs). OperationRepo execution handling for this response type.
* Not affected: Retryable / unauthorized handling (RETRYABLE, UNAUTHORIZED) and normal successful login flows. No public API behavior changes besides avoiding an unrecoverable stuck state.

### OPTIONAL - Other
We are basically hitting this loop:
1. login(externalId) updates local identity state (externalId is set)
2. backend rejects it (400 “restricted”), but the op is treated as “pause and retry later”
3. user never gets created (onesignalId stays unset)
4. the “bad state recovery” sees externalId set + onesignalId unset and tries to “recover”
5. recovery replays the same login, which will always be rejected -> stuck

# Testing
## Unit testing
Updated unit coverage to verify ResponseStatusType.INVALID maps to ExecutionResult.FAIL_INVALID_LOGIN (and does not return FAIL_PAUSE_OPREPO).

## Manual testing
Tested on emulator Pixel 7 API 33
Step to reproduce:
1. Install app
2. Call Login with external ID "1"
3. Observe that the SDK is now stuck in a bad state and unable to perform any operation until app is reinstall.

After fix: Now able to call Logout or Login with another valid external ID.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2523)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationExecutor.kt
@@ -79,6 +79,13 @@ enum class ExecutionResult {
 
     /**
      * Used in special create user case.
+     * The operation failed due to invalid arguments (eg. restricted external ID is used.)
+     * We should not retry the operation as the external ID should not be used again.
+     */
+    FAIL_INVALID_LOGIN,
+
+    /**
+     * Used in special create user case.
      * The operation failed due to a non-retryable error. Pause the operation repo
      * and retry on a new session, giving the SDK a chance to recover from the failed user create.
      */
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationExecutor.kt
@@ -83,11 +83,4 @@ enum class ExecutionResult {
      * We should not retry the operation as the external ID should not be used again.
      */
     FAIL_INVALID_LOGIN,
-
-    /**
-     * Used in special create user case.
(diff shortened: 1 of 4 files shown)


abdulraqeeb33 (approved):
looks fine to me. thank you for adding the unit tests

nan-li (changes requested):
temp block

PR #2531, opened 2026-01-27
chore: Logging lower severity

jkasten2:
# Description
## One Line Summary
Better classify errors, warning, ect we send to OneSignal, which will lower volume as well.

### Motivation
The main motivation is to better classify errors, warning, ect we send in the otel format to OneSignal, which will lower volume as well.

## Considerations
There is only one severity per log entry, however the severity is different between the app developer and OneSignal. 
Example, the app developer may have something configured wrong in their project preventing the SDK from registering for push, this would be an error in the context of the app developer, but OneSignal would consider this info instead. An error in the context of the app developer since this means a major feature is not working for their app, but only an info level since the SDK is not at fault.

### Scope
Only effects logging

## Manual testing
Tested on an Android 14 emulator

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [X] PR does one thing

## Testing

## Final pass

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2531)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/background/impl/BackgroundManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/background/impl/BackgroundManager.kt
@@ -186,7 +186,7 @@ internal class BackgroundManager(
         } catch (e: NullPointerException) {
             // Catch for buggy Oppo devices
             // https://github.com/OneSignal/OneSignal-Android-SDK/issues/487
-            Logging.error(
+            Logging.info(
                 "scheduleSyncServiceAsJob called JobScheduler.jobScheduler which " +
                     "triggered an internal null Android error. Skipping job.",
                 e,
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
@@ -155,9 +155,9 @@ internal class NotificationLifecycleService(
                 },
                 onError = { ex ->
                     if (ex is BackendException) {
-                        Logging.error("Notification opened confirmation failed with statusCode: ${ex.statusCode} response: ${ex.response}")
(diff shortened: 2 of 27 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/background/impl/BackgroundManager.kt
```diff
@@ -186,7 +186,7 @@ internal class BackgroundManager(
         } catch (e: NullPointerException) {
             // Catch for buggy Oppo devices
             // https://github.com/OneSignal/OneSignal-Android-SDK/issues/487
-            Logging.error(
+            Logging.info(
```
abdulraqeeb33, line 189: should this be a warn?

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/lifecycle/impl/NotificationLifecycleService.kt
```diff
@@ -275,22 +275,19 @@ internal class NotificationLifecycleService(
...
                     activity.startActivity(intent)
                 }
             } else {
-                Logging.info("SDK not showing an Activity automatically due to it's settings.")
+                Logging.debug("SDK not showing an Activity automatically due to it's settings.")
             }
         } catch (e: JSONException) {
-            Logging.error("Could not parse JSON to open notification activity.")
-            e.printStackTrace()
+            Logging.error("Could not parse JSON to open notification activity.", e)
```
abdulraqeeb33, line 286: 💯

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/background/impl/BackgroundManager.kt
jkasten2: It's a known bug with some devices, per the comment above. Since it is being handled as expected by the code I would consider this info.

PR #2548, opened 2026-02-17
chore: Removes outdated example app

fadi-george:
Deletes the old example app and associated prompt, streamlining the repository by removing unnecessary files and focusing on the current implementation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2548)


abdulraqeeb33 (approved):
have you made sure the Huawei version works correctly?
also I see the spotless files deleted, would these be carried over to the new implementation?

PR #2565, opened 2026-03-09
fix: notification priority set

abdulraqeeb33:
## Summary
- Fixed Android notification priority mapping so that OneSignal payload priority 9–10 correctly maps to `Notification.PRIORITY_MAX` and `IMPORTANCE_MAX`, instead of `PRIORITY_HIGH` / `IMPORTANCE_HIGH`.
- Extracted all priority threshold logic into a shared `NotificationPriorityMapper` utility to prevent the three call sites from drifting out of sync.

## Context
Customers reported that notifications sent with the highest priority (10) were consistently ranked below competitor notifications in the Android notification shade. The `adb shell dumpsys notification` output showed `pri=1` (PRIORITY_HIGH) instead of the expected `pri=2` (PRIORITY_MAX), because the backend sends `pri=9` for the highest setting and the SDK used a strict `> 9` threshold that excluded it.

## Changes
- **`NotificationPriorityMapper.kt`** (new) — single source of truth for OneSignal-to-Android priority/importance mapping with a `HIGH_PRIORITY_THRESHOLD` constant.
- **`NotificationDisplayBuilder.kt`** — delegates `convertOSToAndroidPriority()` to the shared mapper.
- **`NotificationChannelManager.kt`** — delegates `priorityToImportance()` to the shared mapper.
- **`NotificationBundleProcessor.kt`** — uses `NotificationPriorityMapper.isHighPriority()` instead of an inline threshold check.

## Updated mapping (OneSignal pri → Android)
| OneSignal pri | Android Priority | Android Importance |
|:---|:---|:---|
| 9–10 | PRIORITY_MAX (2) | IMPORTANCE_MAX (5) |
| 7–8 | PRIORITY_HIGH (1) | IMPORTANCE_HIGH (4) |
| 5–6 | PRIORITY_DEFAULT (0) | IMPORTANCE_DEFAULT (3) |
| 3–4 | PRIORITY_LOW (-1) | IMPORTANCE_LOW (2) |
| 1–2 | PRIORITY_MIN (-2) | IMPORTANCE_MIN (1) |

## Test plan
- [x] Unit tests for full priority range mapping in `NotificationDisplayBuilderTests`
- [x] Unit tests for full importance range mapping in `NotificationChannelManagerTests`
- [x] Unit tests for `isHighPriority` flag in `NotificationBundleProcessorTests`
- [x] Regression tests verifying `pri=9` maps to MAX (not HIGH) in all three locations
- [x] Manual verification: sent priority-10 push via REST API, confirmed `pri=2` (PRIORITY_MAX) in `adb shell dumpsys notification`

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Android-SDK/2565)

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/bundle/impl/NotificationBundleProcessor.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/bundle/impl/NotificationBundleProcessor.kt
@@ -7,6 +7,7 @@ import com.onesignal.core.internal.time.ITime
 import com.onesignal.notifications.internal.bundle.INotificationBundleProcessor
 import com.onesignal.notifications.internal.common.NotificationConstants
 import com.onesignal.notifications.internal.common.NotificationFormatHelper
+import com.onesignal.notifications.internal.common.NotificationPriorityMapper
 import com.onesignal.notifications.internal.generation.INotificationGenerationWorkManager
 import org.json.JSONArray
 import org.json.JSONException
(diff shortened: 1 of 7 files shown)


fadi-george:
why do we have 10 levels of priority

PR #2569, opened 2026-03-10
fix: use versionName instead of versionCode for app version reporting

fadi-george:
# Description
## One Line Summary
Use `versionName` instead of `versionCode` in `AndroidUtils.getAppVersion()` to align with iOS behavior.

### Motivation
Android reports `versionCode` (integer build number) as the app version while iOS reports `CFBundleShortVersionString` (display version string). This causes inconsistent `app_version` values across platforms, especially for cross-platform frameworks like .NET MAUI where `ApplicationVersion` (build number) and `ApplicationDisplayVersion` (display version) are distinct.

Fixes https://github.com/OneSignal/OneSignal-DotNet-SDK/issues/82

### Scope
Only affects `AndroidUtils.getAppVersion()`. All existing call sites (`OneSignalImp`, `SubscriptionOperationExecutor`, `LoginUserOperationExecutor`, `SubscriptionManager`, `CustomEventOperationExecutor`, `OtelPlatformProvider`) will now receive the display version string instead of the build number. Falls back to `versionCode.toString()` if `versionName` is null.

# Testing
## Unit testing
Existing test in `OtelPlatformProviderTest` validates `getAppVersion` is non-null and non-empty, which still passes.

## Manual testing
Not yet tested on device.

# Affected code checklist
   - [x] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/AndroidUtils.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/AndroidUtils.kt
@@ -80,14 +80,12 @@ object AndroidUtils {
     }
 
     fun getAppVersion(context: Context): String? {
-        val appVersion: Int? =
-            try {
-                context.packageManager.getPackageInfo(context.packageName, 0).versionCode
-            } catch (e: PackageManager.NameNotFoundException) {
-                null
-            }
-
-        return appVersion?.toString()
+        return try {
+            val packageInfo = context.packageManager.getPackageInfo(context.packageName, 0)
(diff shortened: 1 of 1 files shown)


nan-li (changes requested):
Blocking while I review, we're technically making an API change and clients may already be using the previous versioning style

fadi-george:
> Blocking while I review, we're technically making an API change and clients may already be using the previous versioning style

Could mark it as a breaking change I suppose, but developers should at the release notes and make according changes if it.

PR #2571, opened 2026-03-12
fix: Unify OTEL exporter diagnostics toggle and align headers

abdulraqeeb33:
## Summary
- centralize OTEL exporter diagnostic logging behind a single provider-level toggle (OTEL_EXPORTER_LOGGING_ENABLED surfaced as isOtelExporterLoggingEnabled) so all OTEL export paths use one source of truth
- add exporter-level diagnostics that log OTEL export lifecycle (request sent, success, failure) and enrich failures with parsed HTTP details (status, status message, and response body when available)
- keep OTEL transport compatible with current edge behavior by using the SDK header expected by the HTTP stack and explicitly validating that legacy blocked header variants are not sent
- normalize OTEL endpoint construction to .../sdk/log?app_id={appId} and add coverage to prevent regressions in path/query formatting
- remove redundant setAllAttributes(Attributes) extension in favor of the platform API/built-in method to reduce duplicated behavior and avoid extension shadowing noise
- expand OTEL tests to cover header shape and endpoint formation, alongside existing exporter/provider wiring checks

## Test plan
- [x] Run `./gradlew :OneSignal:otel:test`
- [x] Run `./gradlew :OneSignal:otel:compileDebugKotlin`
- [x] Verify branch diff against `origin/main` only includes OTEL/header and platform-provider toggle wiring changes

### DIFF
--- a/OneSignalSDK/onesignal/otel/src/main/java/com/onesignal/otel/OneSignalOpenTelemetry.kt
+++ b/OneSignalSDK/onesignal/otel/src/main/java/com/onesignal/otel/OneSignalOpenTelemetry.kt
@@ -17,24 +17,6 @@ internal fun LogRecordBuilder.setAllAttributes(attributes: Map<String, String>):
     return this
 }
 
-/**
- * Extension function to set all attributes from an Attributes object.
- * Made public so it can be used from other modules (e.g., core module for logging).
- */
-fun LogRecordBuilder.setAllAttributes(attributes: io.opentelemetry.api.common.Attributes): LogRecordBuilder {
-    attributes.forEach { key, value ->
-        val keyString = key.key
-        when (value) {
-            is String -> this.setAttribute(keyString, value)
-            is Long -> this.setAttribute(keyString, value)
-            is Double -> this.setAttribute(keyString, value)
-            is Boolean -> this.setAttribute(keyString, value)
-            else -> this.setAttribute(keyString, value.toString())
-        }
-    }
-    return this
-}
-
 internal abstract class OneSignalOpenTelemetryBase(
     private val osTopLevelFields: OtelFieldsTopLevel,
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/android/OtelPlatformProvider.kt
```diff
@@ -135,6 +138,8 @@ internal class OtelPlatformProvider(
         }
     }
 
+    override val isOtelExporterLoggingEnabled: Boolean = OTEL_EXPORTER_LOGGING_ENABLED
```
fadi-george, line 141: how does isOtelExporterLoggingEnabled get set to true?

abdulraqeeb33: we set it manually whenever we want to debug in our local test env.

OneSignalSDK/onesignal/otel/src/main/java/com/onesignal/otel/OneSignalOpenTelemetry.kt
```diff
@@ -103,15 +103,19 @@ internal class OneSignalOpenTelemetryRemote(
 
     val extraHttpHeaders: Map<String, String> by lazy {
         mapOf(
-            "X-OneSignal-App-Id" to appId,
```
nan-li, line 106: we don't need app_id in header anymore because it's already captured elsewhere right

abdulraqeeb33: thats right

Issue #2575, opened 2026-03-12
[Bug]: OneSignal 5.7.1 causes OkHttp runtime incompatibility crash in React Native app (NoClassDefFoundError okhttp3/internal/Util)


azcion:
### What happened?

After upgrading `react-native-onesignal` from `5.3.2` to `5.3.3`, our Android app started crashing at startup with:

`java.lang.NoClassDefFoundError: Failed resolution of: Lokhttp3/internal/Util;`

Sentry points to:
- `okhttp3.JavaNetCookieJar.decodeHeaderAsJavaNetCookies`
- `com.facebook.react.modules.network.ReactCookieJarContainer.loadForRequest`

The crash happens on the OkHttp dispatcher thread shortly after launch when network/cookie handling runs.

Environment:
- React Native: `0.84.1`
- `react-native-onesignal`: `5.3.3` (problematic), `5.3.2` (works)
- OneSignal Android SDK transitively used by RN plugin:
  - `5.7.1` via RN OneSignal `5.3.3` (crashes)
  - `5.6.2` via RN OneSignal `5.3.2` (stable)

What we observed in dependency resolution:
- `./gradlew :app:dependencyInsight --configuration mojakrkaReleaseRuntimeClasspath --dependency com.squareup.okhttp3:okhttp`
- Selected by conflict resolution: `com.squareup.okhttp3:okhttp:5.2.1` (between `5.2.1`, `4.9.2`, `3.14.9`)
- React Native networking stack still includes `okhttp-urlconnection:4.9.2`

This appears to create a runtime-incompatible mix that triggers the crash.

### Steps to reproduce?

```Markdown
1. Use a React Native 0.84.x Android app with OneSignal integrated via react-native-onesignal.
2. Set react-native-onesignal to 5.3.3.
3. Build/install Android release app.
4. Launch app and let first network requests run (cookie handling path).
5. Observe crash: NoClassDefFoundError: okhttp3/internal/Util
6. Run dependency insight and observe resolved OkHttp 5.2.1.
7. Downgrade to react-native-onesignal@5.3.2 (OneSignal Android 5.6.2) and rebuild.
8. Crash disappears; OkHttp resolves to 4.9.2.
```

### What did you expect to happen?

Upgrading OneSignal should not introduce a transitive dependency combination that causes runtime crashes in a standard React Native Android networking setup. The app should start and handle network/cookies without crashing.

### OneSignal Android SDK version

[...]

PR #2577, opened 2026-03-17
fix: [SDK-4160] exclude OkHttp 5.x from otel module transitive dependencies

fadi-george:
# Description

## One Line Summary

Exclude OkHttp 5.x transitive dependency from the otel module to prevent runtime crashes in consumers using OkHttp 4.x.

### Motivation

The `otel` module (added in 5.7.0) depends on `opentelemetry-exporter-otlp`, which transitively pulls in `okhttp-jvm:5.2.1` via `opentelemetry-exporter-sender-okhttp`. OkHttp 5.x is binary-incompatible with OkHttp 4.x — specifically, `okhttp3.internal.Util` was removed in 5.x.

When a consumer (e.g. a React Native app) also depends on OkHttp 4.x, Gradle's conflict resolution upgrades everything to 5.2.1 (since OkHttp 5.x's module metadata declares `okhttp-jvm` replaces `okhttp`). This causes a `NoClassDefFoundError: okhttp3/internal/Util` crash at runtime when React Native's networking stack (`ReactCookieJarContainer` → `JavaNetCookieJar`) handles cookies.

**Dependency chain:**
```
com.onesignal:core:5.7.x
  → com.onesignal:otel:5.7.x
    → io.opentelemetry:opentelemetry-exporter-otlp:1.55.0
      → io.opentelemetry:opentelemetry-exporter-sender-okhttp:1.55.0
        → com.squareup.okhttp3:okhttp-jvm:5.2.1  ← breaks OkHttp 4.x consumers
```

### Scope

- Excludes `com.squareup.okhttp3` group from `opentelemetry-exporter-otlp` transitive dependencies
- Explicitly provides `com.squareup.okhttp3:okhttp:4.12.0` (latest stable 4.x)
- OpenTelemetry's sender-okhttp uses standard public OkHttp APIs that are compatible with 4.x
- No changes to public API or behavior

Fixes https://github.com/OneSignal/OneSignal-Android-SDK/issues/2575

# Testing

## Unit testing

All existing otel module unit tests pass (`./gradlew :OneSignal:otel:testReleaseUnitTest`).

## Manual testing

Verified via `./gradlew :OneSignal:otel:dependencies --configuration releaseRuntimeClasspath` that `okhttp-jvm:5.2.1` no longer appears in the dependency tree and `okhttp:4.12.0` is resolved instead.

Tested crash logs
Screenshot 2026-03-17 at 11 16 55 AM

# Affected code checklist

- [ ] Notifications
- [ ] Outcomes
- [ ] Sessions
- [ ] In-App Messaging
- [x] REST API requests
- [ ] Public API changes

# Checklist

## Overview

- [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
+++ b/examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
@@ -235,7 +235,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application), I
                 refreshTriggers()
                 loadExistingTags()
                 refreshPushSubscription()
-                // Loading stays on; onUserStateChange will call fetchUserDataFromApi() to dismiss it
+                fetchUserDataFromApi()
             }
         }
     }


OneSignalSDK/onesignal/otel/build.gradle
```diff
@@ -52,7 +52,16 @@ dependencies {
...
+    // opentelemetry-exporter-otlp transitively pulls in okhttp-jvm 5.x (via
+    // opentelemetry-exporter-sender-okhttp). OkHttp 5.x is binary-incompatible with consumers
+    // that depend on OkHttp 4.x (e.g. React Native's networking stack references
+    // okhttp3.internal.Util which was removed in 5.x). Exclude it so Gradle doesn't upgrade
+    // the consumer's OkHttp to an incompatible version.
+    // See: https://github.com/OneSignal/OneSignal-Android-SDK/issues/2575
+    implementation('io.opentelemetry:opentelemetry-exporter-otlp') {
+        exclude group: 'com.squareup.okhttp3'
+    }
+    implementation 'com.squareup.okhttp3:okhttp:4.12.0'
```
abdulraqeeb33, line 64: can you please move the version number out in the root project?

PR #2599, opened 2026-03-30
feat: identity verification 5.8

nan-li:
# Description
## One Line Summary
Add Identity Verification (JWT-based) support to the Android SDK, gating API operations behind server-issued JWTs when enabled.

### Motivation
Identity Verification allows app developers to authenticate users with JWTs before the SDK sends operations to the OneSignal backend. This prevents unauthorized API calls by requiring a valid JWT for all user-scoped operations when the feature is enabled server-side via the `jwt_required` remote param.

### Scope

**Public API additions:**
- `OneSignal.login(externalId, jwtBearerToken)` — accepts an optional JWT on login
- `OneSignal.updateUserJwt(externalId, token)` — supply a fresh JWT when the current one expires
- `OneSignal.addUserJwtInvalidatedListener(listener)` / `removeUserJwtInvalidatedListener(listener)` — get notified when a JWT is invalidated (e.g. 401 from backend) so the app can provide a new one
- `IUserJwtInvalidatedListener` / `UserJwtInvalidatedEvent` — listener interface and event class

**Internal changes:**
- `JwtTokenStore` — persistent store mapping externalId → JWT, backed by SharedPreferences
- `IdentityVerificationService` — reacts to the `jwt_required` remote param arriving via config HYDRATE; purges anonymous operations when IV is enabled; wires JWT invalidation callbacks
- `OperationRepo` — gates operation execution on valid JWT when IV is enabled (`hasValidJwtIfRequired`); handles `FAIL_UNAUTHORIZED` by invalidating the JWT, re-queuing operations, and notifying the developer; stamps `externalId` synchronously at enqueue time to avoid race conditions
- `LogoutHelper` — when IV is enabled, disables the push subscription internally and suppresses the backend login operation (the device-scoped user is local-only)
- `ConfigModelStoreListener` / `ParamsBackendService` — reads `jwt_required` from remote params, defaults to `false` when absent
- HTTP layer — attaches `Authorization: Bearer <jwt>` header to requests when IV is enabled
- In-App Messaging — uses alias-based fetch endpoint and includes JWT in IAM backend calls
- Operation executors — propagate `externalId` to result operations and attach JWT to backend requests

**What is NOT affected:**
- Apps without Identity Verification enabled behave exactly as before — `jwt_required` defaults to `false` and no JWT gating occurs
- Notification display, open handling, push processing, and confirm deliveries are unchanged

# Testing
## Unit testing
WIP Adding more testing
- `OperationRepoTests` — added tests for `FAIL_UNAUTHORIZED` handling (identified user fires handler + retries; anonymous user drops operations), and synchronous `externalId` stamping before async dispatch
- `LoginUserOperationExecutorTests` — updated for identity verification parameters
- `LogoutHelperTests` — covers IV-enabled logout flow
- `LoginHelperTests` — covers IV-enabled login flow with JWT

## Manual testing
Tested with the demo app on an Android emulator (Pixel 7, API 35):
- Login with valid JWT → operations execute normally
- Login with expired JWT → 401 triggers `IUserJwtInvalidatedListener`, call `updateUserJwt` → operations resume
- Logout with IV enabled → push subscription disabled, new anonymous user created locally
- App without IV enabled → SDK operates normally, no regression

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [x] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


nan-li:
@claude re-review

examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
```diff
@@ -262,6 +299,12 @@ class MainViewModel(application: Application) : AndroidViewModel(application), I
         }
     }
 
+    fun setUseIdentityVerification(enabled: Boolean) {
+        SharedPreferenceUtil.cacheIdentityVerification(getApplication(), enabled)
```
abdulraqeeb33, line 303: i think should be launched in a viewmodelscope, not on the main thread
something like this
https://github.com/OneSignal/OneSignal-Android-SDK/pull/2599/changes#diff-3887ba13da20ce4e4cc419c785a1ba18d3e796b701c7b7e2c9813b22db761eb5R271

examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
```diff
@@ -127,6 +134,7 @@ class MainViewModel(application: Application) : AndroidViewModel(application), I
         _privacyConsentGiven.value = repository.getPrivacyConsent()
         _inAppMessagesPaused.value = repository.isInAppMessagesPaused()
         _locationShared.value = repository.isLocationShared()
+        _useIdentityVerification.value = SharedPreferenceUtil.getCachedIdentityVerification(context)
```
abdulraqeeb33, line 137: this should run on a background thread

OneSignalSDK/detekt/detekt-baseline-core.xml
```diff
@@ -62,6 +74,8 @@
     <ID>ConstructorParameterNaming:NewRecordsState.kt$NewRecordsState$private val _time: ITime</ID>
     <ID>ConstructorParameterNaming:OSDatabase.kt$OSDatabase$private val _outcomeTableProvider: OutcomeTableProvider</ID>
     <ID>ConstructorParameterNaming:OperationRepo.kt$OperationRepo$private val _configModelStore: ConfigModelStore</ID>
+    <ID>ConstructorParameterNaming:OperationRepo.kt$OperationRepo$private val _identityModelStore: IdentityModelStore</ID>
```
abdulraqeeb33, line 77: @nan-li can we try to address as many detekt issues as possible instead of adding this to the baseline?

OneSignalSDK/detekt/detekt-baseline-in-app-messages.xml
```diff
@@ -1,9 +1,9 @@
-<?xml version="1.0" ?>
+<?xml version='1.0' encoding='UTF-8'?>
 <SmellBaseline>
-  <ManuallySuppressedIssues></ManuallySuppressedIssues>
+  <ManuallySuppressedIssues/>
   <CurrentIssues>
     <ID>ComplexCondition:InAppMessagesManager.kt$InAppMessagesManager$!message.isTriggerChanged &amp;&amp; isMessageDisplayed &amp;&amp; (isTriggerOnMessage || isNewTriggerAdded &amp;&amp; isOnlyDynamicTriggers)</ID>
-    <ID>ComplexMethod:TriggerController.kt$TriggerController$private fun evaluateTrigger(trigger: Trigger): Boolean</ID>
+    <ID>ComplexMethod:InAppMessagesManager.kt$InAppMessagesManager$private suspend fun fetchMessages(rywData: RywData)</ID>
```
abdulraqeeb33: we should break it down if its a complex method.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/IdentityVerificationService.kt
```diff
@@ -0,0 +1,66 @@
...
+ */
+internal class IdentityVerificationService(
+    private val _configModelStore: ConfigModelStore,
+    private val _operationRepo: IOperationRepo,
+    private val _identityModelStore: IdentityModelStore,
+    private val _jwtTokenStore: JwtTokenStore,
+    private val _userManager: UserManager,
+) : IStartableService, ISingletonModelStoreChangeHandler<ConfigModel> {
+    override fun start() {
+        _configModelStore.subscribe(this)
```
abdulraqeeb33, line 31: can this be in a background thread?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -148,6 +157,18 @@ internal class OperationRepo(
...
      * @returns true if the OperationQueueItem was added, false if not
      */
+    /**
+     * Capture the externalId from the current identity model onto the operation
+     * synchronously on the caller's thread, before the async enqueue coroutine runs.
+     * Operations that already set externalId in their constructor (e.g. LoginUserOperation)
+     * are left unchanged.
+     */
+    private fun stampExternalId(operation: Operation) {
+        if (operation.externalId == null) {
```
abdulraqeeb33: wouldn't operation.externalId contain the externalId when login happens?
who is supposed to populate this from the start?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -268,7 +290,21 @@ internal class OperationRepo(
...
                 }
-                ExecutionResult.FAIL_UNAUTHORIZED, // TODO: Need to provide callback for app to reset JWT. For now, fail with no retry.
+                ExecutionResult.FAIL_UNAUTHORIZED -> {
+                    val externalId = startingOp.operation.externalId
+                    if (externalId != null) {
+                        _jwtTokenStore.invalidateJwt(externalId)
+                        _jwtInvalidatedHandler?.invoke(externalId)
+                        Logging.warn("Operation execution failed with 401 Unauthorized, JWT invalidated for user: $externalId. Operations re-queued.")
+                        synchronized(queue) {
+                            ops.reversed().forEach { queue.add(0, it) }
```
abdulraqeeb33: should we remove if we are invalidating jwt?
and it an authorized hit to the backend?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -312,9 +348,13 @@ internal class OperationRepo(
             // if there are operations provided on the result, we need to enqueue them at the
             // beginning of the queue.
             if (response.operations != null) {
+                val parentExternalId = startingOp.operation.externalId
                 synchronized(queue) {
                     for (op in response.operations.reversed()) {
                         op.id = UUID.randomUUID().toString()
+                        if (op.externalId == null && parentExternalId != null) {
```
abdulraqeeb33: why are we attaching the parentExternalId to this transaction?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -389,6 +433,16 @@ internal class OperationRepo(
         }
     }
 
+    private fun hasValidJwtIfRequired(
```
abdulraqeeb33, line 469: can you please add some comments around what this method is doing?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -450,6 +504,32 @@ internal class OperationRepo(
...
         initialized.complete(Unit)
     }
+
+    override fun removeOperationsWithoutExternalId() {
+        synchronized(queue) {
+            val toRemove = queue.filter { it.operation.externalId == null }
+            toRemove.forEach {
+                queue.remove(it)
+                _operationModelStore.remove(it.operation.id)
+                it.waiter?.wake(false)
```
abdulraqeeb33, line 556: should we start a wait first?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/Operation.kt
```diff
@@ -16,6 +16,18 @@ abstract class Operation(name: String) : Model() {
             setStringProperty(::name.name, value)
         }
 
+    /**
+     * The external ID of the user this operation belongs to. Used by [IOperationRepo] to look up
+     * the correct JWT when identity verification is enabled, and to gate anonymous operations.
+     * Stamped automatically by [IOperationRepo] at enqueue time from the current identity model
+     * when not already set by the concrete operation's constructor.
+     */
+    var externalId: String?
```
abdulraqeeb33, line 25: how do we indicate this to be JWT, can we do a better naming?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -409,6 +415,23 @@ internal class OneSignalImp(
         }
     }
 
+    override fun updateUserJwt(
```
abdulraqeeb33, line 424: we should have suspend versions of the same just like other methods

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/identity/JwtTokenStore.kt
```diff
@@ -0,0 +1,105 @@
...
+ * [com.onesignal.core.internal.config.ConfigModel.useIdentityVerification].
+ */
+class JwtTokenStore(
+    private val _prefs: IPreferencesService,
+) {
+    private val tokens: MutableMap<String, String> = mutableMapOf()
+    private var isLoaded = false
+
+    /** Not thread-safe; callers must hold `synchronized(tokens)`. */
+    private fun ensureLoaded() {
```
abdulraqeeb33, line 25: can this be offloaded on a background thread? this a blocking call since its reading from disk

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/identity/JwtTokenStore.kt
```diff
@@ -0,0 +1,105 @@
...
+                }
+            } catch (e: Exception) {
+                Logging.warn("JwtTokenStore: failed to parse persisted tokens, starting fresh", e)
+            }
+        }
+        isLoaded = true
+    }
+
+    /** Not thread-safe; callers must hold `synchronized(tokens)`. */
+    private fun persist() {
```
abdulraqeeb33, line 46: all these writes should be offloaded to background threads

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/listeners/PropertiesModelStoreListener.kt
```diff
@@ -27,6 +33,8 @@ internal class PropertiesModelStoreListener(
         oldValue: Any?,
         newValue: Any?,
     ): Operation? {
+        if (shouldSuppressForAnonymousUser()) return null
```
abdulraqeeb33: in what other circumstances this returns a null. I would hope that for this we return an Operation but its a no-op.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/LoginUserOperation.kt
```diff
@@ -32,15 +32,6 @@ class LoginUserOperation() : Operation(LoginUserOperationExecutor.LOGIN_USER) {
             setStringProperty(::onesignalId.name, value)
         }
 
-    /**
-     * The optional external ID of this newly logged-in user. Must be unique for the [appId].
-     */
-    var externalId: String?
```
abdulraqeeb33, line 38: we have something similar in the Operation.kt now? did we move this from here?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LogoutHelper.kt
```diff
@@ -18,20 +20,24 @@ class LogoutHelper(
                 return
             }
 
-            // Create new device-scoped user (clears external ID)
-            userSwitcher.createAndSwitchToNewUser()
+            if (configModel.useIdentityVerification == true) {
```
abdulraqeeb33, line 23: why dont we move this login into the userSwitcher? that way we want write tests on it?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
```diff
@@ -43,6 +45,21 @@ internal open class UserManager(
...
+    fun addJwtInvalidatedListener(listener: IUserJwtInvalidatedListener) {
+        jwtInvalidatedNotifier.subscribe(listener)
+    }
+
+    fun removeJwtInvalidatedListener(listener: IUserJwtInvalidatedListener) {
+        jwtInvalidatedNotifier.unsubscribe(listener)
+    }
+
+    fun fireJwtInvalidated(externalId: String) {
+        jwtInvalidatedNotifier.fireOnMain {
```
abdulraqeeb33: why should it be fired on main?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
```diff
@@ -75,7 +75,7 @@ internal class UserModule : IModule {
         builder.register<LoginUserOperationExecutor>().provides<IOperationExecutor>()
         builder.register<LoginUserFromSubscriptionOperationExecutor>().provides<IOperationExecutor>()
         builder.register<RefreshUserOperationExecutor>().provides<IOperationExecutor>()
-        builder.register<UserManager>().provides<IUserManager>()
+        builder.register<UserManager>().provides<IUserManager>().provides<UserManager>()
```
abdulraqeeb33: is this a typo?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
```diff
@@ -226,4 +226,31 @@ interface IOneSignal {
...
     suspend fun logoutSuspend()
+
+    /**
+     * Update the JWT bearer token for a user identified by [externalId]. Call this when
+     * a token is about to expire or after receiving an [IUserJwtInvalidatedListener] callback.
+     *
+     * @param externalId The external ID of the user whose token is being updated.
+     * @param token The new JWT bearer token.
+     */
+    fun updateUserJwt(
```
abdulraqeeb33, line 237: like i mentioned elsewhere, we need suspend methods for these

abdulraqeeb33 (changes requested):

examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
nan-li: This is just the demo app and this is not using an SDK API. It's also consistent with the other methods in this file like `fun setConsentRequired(required: Boolean)`

examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
nan-li: This is just the demo app and this is not using an SDK API. A few lines above this in the same method we do `_appId.value = SharedPreferenceUtil.getOneSignalAppId(context)` so it follows existing behavior.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/IdentityVerificationService.kt
nan-li: it follows the pattern of other subscribing services like [InAppMessagesManager.kt start](https://github.com/OneSignal/OneSignal-Android-SDK/blob/1bc6dc4b39d5839fefa9532a74bc3ea332536cca/OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/InAppMessagesManager.kt#L187), and the `start()` method itself is managed by `useBackgroundThreading` or not (see [StartupService.kt](https://github.com/OneSignal/OneSignal-Android-SDK/blob/1bc6dc4b39d5839fefa9532a74bc3ea332536cca/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/startup/StartupService.kt#L32))

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: Typically the operation will not have an externalId, so we stamp it at enqueue time. However, I just updated this logic so externalId is set on operations at creation time, similar to how onesignalId is set on it.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: No, we give the client a chance to update the token and retry this operation. Since tokens are only 1 hour long max, it will be common for them to expire. Otherwise the client is not going to know, for example, the addTag operation they just did failed and they should call addtag again but the addtag they did 30 minutes ago is ok, does that make sense

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: I rethought this and we will set externalId at operation creation time, similar to how onesignalId is set on it.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: Done

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: Unsure of this question?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/Operation.kt
nan-li: This is the externalId of the user the operation belongs to, I'm not sure I understand your question. While it is relevant for Identity Verification, the term `externalId` is the most clear

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
nan-li: added

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/identity/JwtTokenStore.kt
nan-li: Good callout, I traced all code paths to `ensureLoaded()` (disk read) and `persist()` (disk write) and it's not an issue. This is also similar to the ModelStore's usage of **reading in `load()`**: https://github.com/OneSignal/OneSignal-Android-SDK/blob/1bc6dc4b39d5839fefa9532a74bc3ea332536cca/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ModelStore.kt#L170 and **writing in `persist()`**: https://github.com/OneSignal/OneSignal-Android-SDK/blob/1bc6dc4b39d5839fefa9532a74bc3ea332536cca/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/ModelStore.kt#L219
- All getJwt() callers (operation executors, OperationRepo.hasValidJwtIfRequired, IdentityVerificationService) run on background threads.
- invalidateJwt() and pruneToExternalIds() are only called from OperationRepo's coroutine — also background.
- putJwt() from LoginHelper.login() and updateUserJwtSuspend() — background. The non-suspend updateUserJwt() was the one path that could run on the caller's thread, but we've now added waitForInit to it, which means ensureLoaded() has already completed on background (via OperationRepo.start() → pruneToExternalIds()) before putJwt is reached. So ensureLoaded() is a no-op at that point.
- persist() uses SharedPreferences.Editor.apply() (async) — no blocking disk write.

**Result:** the only work that can happen on the caller's thread is an in-memory map write + apply(). No changes needed here.

OneSignalSDK/detekt/detekt-baseline-core.xml
nan-li: Yes, I undid the detekt changes and then added them back in the last few commits with reasoning. This particular one is because we have a convention already of prepending `_` to param names.

OneSignalSDK/detekt/detekt-baseline-in-app-messages.xml
nan-li: Yes, I undid the detekt changes and then added them back in the last few commits with reasoning. I actually didn't end up making changes to in-app-messages in detekt.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/identity/JwtTokenStore.kt
nan-li: responded in other comment about `persist()` and `ensureLoaded()`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/listeners/PropertiesModelStoreListener.kt
nan-li: With IV=ON, this can happen on new install before they call `login` yet and locally we have an anonymous user, or they called `logout`. We can either do this which doesn't enqueue any operations for these states, or we can enqueue them and just drop them later when we process. This keeps us from some extraneous calls like making the operation queue larger and having to keep persisting, etc.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/LoginUserOperation.kt
nan-li: Yeah we moved this to the parent `Operation.kt` now that all operations are going to keep an externalId property. We don't need LoginUserOperation to be unique.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LogoutHelper.kt
nan-li: Unsure about the tests part since we have LogoutHelperTests and UserSwitcherTests. Do you mean to maybe remove LogoutHelper since it doesn't do much but just delegate to userSwitcher?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
nan-li: I was following the example of some other observers we have, but taking another look, it is inconsistent. We fire things like `onPushSubscriptionChange` or `onNotificationPermissionChange` with fireOnMain but not `onUserStateChange`. The developer **could** respond with updating some sort of UI, but that is probably not typical, so I updated this to just `fire`

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/UserModule.kt
nan-li: This registers UserManager as providing both the interface and the concrete type, it's needed because IdentityVerificationService injects UserManager (concrete) while other consumers inject IUserManager (interface). We have a similar example from [RequestPermissionService](https://github.com/OneSignal/OneSignal-Android-SDK/blob/1bc6dc4b39d5839fefa9532a74bc3ea332536cca/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/CoreModule.kt#L73-L75)
```
builder.register<RequestPermissionService>()
     .provides<RequestPermissionService>()
     .provides<IRequestPermissionService>()
```

but I will update the formatting to be consistent (new lines).

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/IOneSignal.kt
nan-li: added

examples/demo/app/src/main/java/com/onesignal/sdktest/ui/main/MainViewModel.kt
abdulraqeeb33: i will create to make sure we update the code in the demo app that offloads work on the background thread. https://linear.app/onesignal/issue/SDK-4343/offload-io-tasks-etc-to-run-on-background-thread-in-the-android-demo

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LogoutHelper.kt
abdulraqeeb33: ah never mind, i think we are good here

abdulraqeeb33 (approved):
Lets try and do a bug bash on this today/tomorrow?

nan-li:
@claude review

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
```diff
@@ -4,46 +4,73 @@ import com.onesignal.core.internal.config.ConfigModel
...
 
-            newIdentityOneSignalId = identityModelStore.model.onesignalId
+            val newOneSignalId = identityModelStore.model.onesignalId
+
+            val existingOneSignalId =
+                if (configModel.useIdentityVerification == true) {
+                    null
+                } else {
+                    if (currentExternalId == null) currentOneSignalId else null
+                }
```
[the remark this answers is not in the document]
nan-li: However, if IV returns as IV=OFF, then this would not be correct state

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/Operation.kt
```diff
@@ -16,6 +16,18 @@ abstract class Operation(name: String) : Model() {
...
+        get() = getOptStringProperty(::externalId.name)
+        set(value) {
+            setOptStringProperty(::externalId.name, value)
+        }
```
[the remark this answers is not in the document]
nan-li: Addressed

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -268,7 +283,28 @@ internal class OperationRepo(
...
+                        dispatchJwtInvalidatedToApp(externalId)
+                    } else {
+                        Logging.warn("Operation execution failed with 401 Unauthorized for anonymous user. Operations dropped.")
+                        ops.forEach { _operationModelStore.remove(it.operation.id) }
+                        ops.forEach { it.waiter?.wake(false) }
+                    }
+                }
                 ExecutionResult.FAIL_NORETRY,
                 ExecutionResult.FAIL_CONFLICT,
                 -> {
```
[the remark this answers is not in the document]
nan-li: > 401, 403 -> ResponseStatusType.UNAUTHORIZED

Changing this is outside the scope of this PR

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -268,7 +285,29 @@ internal class OperationRepo(
...
+                            ops.reversed().forEach {
+                                queue.add(0, OperationQueueItem(it.operation, waiter = null, bucket = it.bucket, retries = it.retries))
+                            }
+                        }
+                        dispatchJwtInvalidatedToApp(externalId)
+                    } else {
+                        Logging.warn("Operation execution failed with 401 Unauthorized for anonymous user. Operations dropped.")
+                        ops.forEach { _operationModelStore.remove(it.operation.id) }
+                        ops.forEach { it.waiter?.wake(false) }
+                    }
```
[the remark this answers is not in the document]
nan-li: Not an issue. The ops were intentional user actions that failed only because of an expired JWT, not because they were invalid. When alice logs in again with a fresh JWT, those ops should execute (retry what was pending).

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
```diff
@@ -190,7 +201,8 @@ internal class SubscriptionOperationExecutor(
                     AndroidUtils.getAppVersion(_applicationService.appContext),
                 )
 
-            val rywData = _subscriptionBackend.updateSubscription(lastOperation.appId, lastOperation.subscriptionId, subscription)
+            val jwt = lastOperation.externalId?.let { _jwtTokenStore.getJwt(it) }
+            val rywData = _subscriptionBackend.updateSubscription(lastOperation.appId, lastOperation.subscriptionId, subscription, jwt)
 
             if (rywData != null) {
                 _consistencyManager.setRywData(startingOperation.onesignalId, IamFetchRywTokenKey.SUBSCRIPTION, rywData)
```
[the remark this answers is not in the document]
nan-li: > updateSubscription()'s catch block is still missing the UNAUTHORIZED -> FAIL_UNAUTHORIZED branch... A 401 from the backend still falls through... siently dropping the operation

UpdateSubscriptionOperation has requiresJwt = false — the subscription update endpoint does not require JWT authentication, so it won't return 401 for a missing JWT. The jwt passed to the backend call is opportunistic; the endpoint works without it. Updating the JWT token will not help it pass.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -378,14 +384,20 @@ internal class OneSignalImp(
...
+        if (isBackgroundThreadingEnabled) {
+            suspendifyOnIO { loginHelper.enqueueLogin(context) }
+        } else {
             Thread {
                 runBlocking(runtimeIoDispatcher) {
-                    loginHelper.login(externalId, jwtBearerToken)
+                    loginHelper.enqueueLogin(context)
                 }
             }.start()
         }
```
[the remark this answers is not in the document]
nan-li: There's no more "redacted" logs. This is already addressed, the review bot was reviewing the previous commit

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/SubscriptionOperationExecutor.kt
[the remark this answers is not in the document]
nan-li: `UpdateSubscriptionOperation.canStartExecute` requires `!IDManager.isLocalId(subscriptionId)`. If the `CreateSubscriptionOperation` hasn't executed yet, the `subscriptionId` is still a local ID, so the update can never be selected as a starting op by `getNextOps()`. This scenario is unreachable.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/IdentityVerificationService.kt
```diff
@@ -0,0 +1,69 @@
...
+            val externalId = _identityModelStore.model.externalId
+            if (externalId != null && _jwtTokenStore.getJwt(externalId) == null) {
+                Logging.debug("IdentityVerificationService: IV enabled but no JWT for $externalId, will fire invalidated event after queue wake")
+                jwtInvalidatedExternalId = externalId
+            }
+        }
+
+        _operationRepo.forceExecuteOperations()
+
+        jwtInvalidatedExternalId?.let { _userManager.fireJwtInvalidated(it) }
```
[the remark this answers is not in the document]
nan-li: The `fireJwtInvalidated` on HYDRATE is a migration convenience for the current user — if a previous user never had a JWT, the developer never provided one and should already know. Operations for non-current users will execute if/when that user logs in again with a valid JWT.

OneSignalSDK/onesignal/in-app-messages/src/main/java/com/onesignal/inAppMessages/internal/backend/impl/InAppBackendService.kt
```diff
@@ -209,6 +212,7 @@ internal class InAppBackendService(
         baseUrl: String,
         rywData: RywData,
         sessionDurationProvider: () -> Long,
+        jwt: String? = null,
     ): List<InAppMessage>? {
         var attempts = 0
         var retryLimit: Int = 0 // retry limit is remote defined & set dynamically below
```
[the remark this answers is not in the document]
nan-li: Addressed in #2613 — `InAppBackendService` now throws `BackendException` on 401, and `InAppMessagesManager` catches it, stores pending retry state, and retries after JWT refresh via `IJwtUpdateListener`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/session/internal/session/impl/SessionListener.kt
```diff
@@ -44,7 +44,7 @@ internal class SessionListener(
 
     override fun onSessionStarted() {
         _propertiesModelStore.model.timezone = TimeUtils.getTimeZoneId()
-        _operationRepo.enqueue(TrackSessionStartOperation(_configModelStore.model.appId, _identityModelStore.model.onesignalId), true)
+        _operationRepo.enqueue(TrackSessionStartOperation(_configModelStore.model.appId, _identityModelStore.model.onesignalId, _identityModelStore.model.externalId), true)
     }
 
     override fun onSessionActive() {
```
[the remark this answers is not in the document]
nan-li: minor but we can simplify it

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/user/internal/LogoutHelperTests.kt
```diff
@@ -168,4 +178,85 @@ class LogoutHelperTests : FunSpec({
...
+        verify {
+            mockOperationRepo.enqueue(
+                withArg<LoginUserOperation> { operation ->
+                    operation.appId shouldBe appId
+                    operation.externalId shouldBe null
+                },
+            )
+        }
+    }
 })
```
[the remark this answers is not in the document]
nan-li: Pre-existing test, not introduced by this PR. The onesignalId assertion passes due to mock behavior — noted, but out of scope for this change.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
```diff
@@ -237,12 +237,15 @@ class ConfigModel : Model() {
...
     /**
-     * Whether SMS auth hash should be used.
+     * Whether identity verification (JWT) is required for this application.
+     * - `null` = unknown (remote params haven't arrived yet; all operations are held)
+     * - `false` = explicitly disabled (SDK behaves as today, no JWT gating)
+     * - `true` = enabled (operations require a valid JWT, anonymous users are blocked)
      */
-    var useIdentityVerification: Boolean
-        get() = getBooleanProperty(::useIdentityVerification.name) { false }
+    var useIdentityVerification: Boolean?
```
abdulraqeeb33, line 245: should we consider maintaining an Enum state here as opposed to a boolean that signifies three states?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/http/impl/HttpClient.kt
```diff
@@ -195,6 +183,20 @@ internal class HttpClient(
                         con.setRequestProperty("OneSignal-Session-Duration", headers.sessionDuration.toString())
                     }
 
+                    if (headers?.jwt != null) {
```
abdulraqeeb33, line 186: can we try using the kotlin version of null check here?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -188,7 +200,8 @@ internal class OperationRepo(
             }
 
             val ops = getNextOps(executeBucket)
-            Logging.debug("processQueueForever:ops:\n$ops")
+            val queueSnapshot = synchronized(queue) { queue.toList() }
```
abdulraqeeb33, line 203: Can you please elaborate on what we are doing here?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -268,7 +288,29 @@ internal class OperationRepo(
                     ops.forEach { _operationModelStore.remove(it.operation.id) }
                     ops.forEach { it.waiter?.wake(true) }
                 }
-                ExecutionResult.FAIL_UNAUTHORIZED, // TODO: Need to provide callback for app to reset JWT. For now, fail with no retry.
+                ExecutionResult.FAIL_UNAUTHORIZED -> {
+                    val identityVerificationEnabled = _configModelStore.model.useIdentityVerification == true
```
abdulraqeeb33, line 292: should we move this as a convenience method in the configModelStore so that its easy enough to refer to it. you could event create a Kotlin extension for it. also it could be testable

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -239,6 +252,13 @@ internal class OperationRepo(
         }
     }
 
+    private fun dispatchJwtInvalidatedToApp(externalId: String) {
+        _jwtInvalidatedHandler?.let { handler ->
```
abdulraqeeb33, line 256: if these handlers are not setup in time, we could silently skip this. should we queue up the invocations and invoke them when the handler is actually set?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -389,6 +441,40 @@ internal class OperationRepo(
...
     }
 
+    /**
+     * Drop anonymous operations at enqueue time when IV is enabled.
+     * LoginUserOperation is exempt — it's enqueued intentionally during logout
+     * and purged later by [removeOperationsWithoutExternalId] if needed.
+     */
+    private fun shouldSuppressAnonymousOp(op: Operation): Boolean {
+        if (op is LoginUserOperation) return false
+        return _configModelStore.model.useIdentityVerification == true && op.externalId == null
```
abdulraqeeb33, line 451: is useIdentityVerification set to be volatile? that way has the accurate value in every read?

nan-li:
done in new prs

PR #2511, opened 2025-12-28
feat: tel crash reporting

abdulraqeeb33:
# Description
## One Line Summary
Logging, Crash handling using Otel

### Motivation
We want to be able to capture logs, crashes and ANRs  - report them to our servers so that we can inspect and fix them on devices.

### 1. Platform-Agnostic Design & Dependency Reduction

- **Created `OtelIdResolver`** - Reads OneSignal IDs directly from SharedPreferences (no service dependencies)
- **Refactored `OtelPlatformProvider`** - Uses `OtelIdResolver` internally, removing all service dependencies
- **Created factory function `createAndroidOtelPlatformProvider(context)`** - Single parameter, no services needed
- **Removed service dependencies from `initializeOtelLogging()`** - No longer needs `IInstallIdService`, `ConfigModelStore`, `IdentityModelStore`

### 2. Initialization Consolidation

- **Converted `OneSignalCrashLogInit` from `object` to `class`** - Manages platform provider lifecycle as instance variable
- **Platform provider created once and reused** - Single instance used for both crash handler and logging initialization
- **Early crash handler initialization** - Initialized before other services to capture all crashes

### 3. Component Simplification

- **`OneSignalCrashHandlerFactory`** - Simplified to only need `Context` and `IOtelLogger`
- **`OtelFactory.createCrashHandler`** - Refactored to compose other factory methods, reducing duplication

### 4. Constants & Utilities

- **Created `AnrConstants.kt`** - Centralized `DEFAULT_ANR_THRESHOLD_MS` and `DEFAULT_CHECK_INTERVAL_MS`
- **Centralized `isOneSignalAtFault()` logic** - Moved to `OtelCrashHandler.kt` (shared by crash handler and ANR detector)
- **Renamed constants** - `CONFIG_NAME_SPACE`, `IDENTITY_NAME_SPACE` to match naming conventions (updated 77 usages across 8 files)

### 5. Code Quality Improvements

- **Extracted helper methods in `OtelAnrDetector`** - Reduced complexity (LongMethod, NestedBlockDepth)
- **Fixed all detekt issues** - Added appropriate `@Suppress` annotations where needed
- **Improved exception handling** - More specific exception types where appropriate

### 6. Comprehensive Test Suite

- **Created `OtelIdResolverTest.kt`** - Comprehensive tests for all ID resolution methods with fallback scenarios
- **Created `OtelPlatformProviderTest.kt`** - Tests for platform provider properties and factory function
[...]

PR #2601, opened 2026-04-07
fix: [SDK-4192] downgrade Kotlin from 2.2.0 to 1.9.25

fadi-george:
# Description
## One Line Summary
Downgrade Kotlin from 2.2.0 to 1.9.25 to maximize compatibility with consuming projects (including wrapper SDKs on Kotlin 1.9.x).

### Motivation
The otel crash reporting PR (#2511) bumped Kotlin from 1.9.25 to 2.2.0. This forces all consumers of the SDK to use Kotlin 2.x, which is a messy change for wrapper SDKs (Flutter, React Native, Cordova, Capacitor, Unity, .NET) and any app still on Kotlin 1.9.x. Since no Kotlin 2.x language features are used in the SDK source code, downgrading to 1.9.25 restores the widest compatibility window — consumers on both 1.9.x and 2.x can use the SDK.

### Scope
- **SDK build files**: `kotlinVersion` reverted to `1.9.25` in `OneSignalSDK/build.gradle`
- **Otel module**: Added `-Xskip-metadata-version-check` compiler flag to handle transitive dependencies (okio via OkHttp via OpenTelemetry OTLP exporter) compiled with Kotlin 2.x metadata. This is a compile-time-only metadata check skip; binary compatibility is maintained.
- **Test fix**: `OtelIntegrationTest.kt` required an explicit `!!` non-null assertion that Kotlin 2.x's improved smart casts had made unnecessary.
- **Demo app**: Replaced the Kotlin 2.0+ `org.jetbrains.kotlin.plugin.compose` plugin with the pre-2.0 `composeOptions` approach using Compose compiler extension `1.5.15` (the version compatible with Kotlin 1.9.25). Also renamed user-facing "Activity" labels to "Screen".
- **No source code changes** in the SDK itself — only build config and one test fix.

# Testing
## Unit testing
- `assembleRelease` builds successfully (all modules including otel)
- `:OneSignal:otel:testReleaseUnitTest` passes
- `:OneSignal:core:testReleaseUnitTest` passes (2 pre-existing `SDKInitTests` failures unrelated to this change — also fail on `main`)

## Manual testing
Build verified locally via `./gradlew assembleRelease` — all 358 tasks succeeded.
Screenshot 2026-04-07 at 10 39 57 AM
Screenshot 2026-04-07 at 10 38 16 AM

Check crash still logs:

# Affected code checklist
   - [ ] Notifications
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/debug/internal/crash/OtelIntegrationTest.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/debug/internal/crash/OtelIntegrationTest.kt
@@ -76,7 +76,7 @@ class OtelIntegrationTest : FunSpec({
             put(identityModel)
         }
 
-        sharedPreferences.edit()
+        sharedPreferences!!.edit()
             .putString(PreferenceOneSignalKeys.MODEL_STORE_PREFIX + configNameSpace, configArray.toString())
             .putString(PreferenceOneSignalKeys.MODEL_STORE_PREFIX + identityNameSpace, identityArray.toString())
             .putString(PreferenceOneSignalKeys.PREFS_OS_INSTALL_ID, "test-install-id")
--- a/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/debug/internal/crash/OtelIntegrationTest.kt
+++ b/OneSignalSDK/onesignal/core/src/test/java/com/onesignal/debug/internal/crash/OtelIntegrationTest.kt
@@ -35,133 +35,128 @@ private infix fun <T> T.shouldBeOneOf(expected: List<T>) {
 
 @RobolectricTest
 @Config(sdk = [Build.VERSION_CODES.O])
-class OtelIntegrationTest : FunSpec({
-    var appContext: Context? = null
-    var sharedPreferences: SharedPreferences? = null
+class OtelIntegrationTest : FunSpec() {
+    private lateinit var appContext: Context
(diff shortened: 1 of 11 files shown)


OneSignalSDK/onesignal/core/src/test/java/com/onesignal/debug/internal/crash/OtelIntegrationTest.kt
```diff
@@ -76,7 +76,7 @@ class OtelIntegrationTest : FunSpec({
             put(identityModel)
         }
 
-        sharedPreferences.edit()
+        sharedPreferences!!.edit()
```
abdulraqeeb33: can we avoid the `!!` instead add .?

abdulraqeeb33 (changes requested):

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/debug/internal/crash/OtelIntegrationTest.kt
fadi-george: Updated to use lateinit var

abdulraqeeb33 (approved):

PR #2608, opened 2026-04-09
fix(ci): PR diff coverage uses base branch SHAs and aggregate 80% gate

abdulraqeeb33:
## Summary
Aligns diff coverage with the PR’s chosen **base branch** and clarifies what the check enforces.

## Changes
- **PR diff source:** On `pull_request`, use `github.event.pull_request.base.sha` and `head.sha` so the measured patch matches GitHub’s PR (any base branch, not `main` only and not the merge ref alone).
- **Gate:** Require **aggregate** coverage on touched executable lines ≥ `DIFF_COVERAGE_THRESHOLD` (80%). Per-file numbers remain in the report as context.
- **Safety:** If `src/main` Kotlin/Java files change in the diff but none map to the merged JaCoCo report, the job fails with a clear message.
- **UX:** Renamed the workflow step so it reads as coverage on changed lines, not a vague “diff tests” label.

## Testing
- `bash -n OneSignalSDK/coverage/checkCoverage.sh`
- CI will exercise the script on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,9 +63,13 @@ jobs:
           else
             echo "bypass=false" >> $GITHUB_OUTPUT
           fi
-      - name: "[Diff Coverage] Check coverage"
+      - name: "[PR] Coverage on changed lines (min ${{ env.DIFF_COVERAGE_THRESHOLD }}%)"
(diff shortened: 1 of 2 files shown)


fadi-george:
Could keep it simple with just 80% test threshold (overall) check in any situation like in other sdks

fadi-george (approved):

PR #2612, opened 2026-04-14
feat: SDK-4363: Turbine remote SDK feature flags and foreground refresh

abdulraqeeb33:
## Summary

Implements Turbine-aligned remote SDK feature flags for Android. The SDK now fetches a per-app feature-flag document from Turbine while in the foreground and applies the result through `FeatureManager`, with per-flag activation modes so threading-sensitive flags only flip at cold start.

## Linear

SDK-4363

### Backend client
- `IFeatureFlagsBackendService` + sealed `RemoteFeatureFlagsFetchOutcome` (`Success` / `Unavailable`) so callers don't have to interpret HTTP status codes.
- `FeatureFlagsBackendService` — `GET /apps/:app_id/sdk/features/android/:sdk_version`:
  - 4xx → `Unavailable`, logged at WARN with a 200-char single-line body snippet (Turbine errors like 403/404 are obvious in logcat without flipping log level).
  - 5xx / network (`statusCode=0`) → `Unavailable`, logged at DEBUG to keep transient failures quiet.
  - 200 with non-contract JSON → `Unavailable`, logged at WARN (surfaces real contract regressions).
- `TurbineSdkFeatureFlagsPath` — validates and encodes the SDK version path segment (allow-lists `[A-Za-z0-9._-]`, rejects empties / path-traversal).
- `HttpResponse.isClientError` — narrows `!isSuccess` to 4xx so the service can log 4xx at WARN and 5xx/network at DEBUG without reimplementing status-code ranges at each call site.

### JSON parsing
- `FeatureFlagsJsonParser` (kotlinx.serialization):
  - Parses `{ "features": ["flag", { "id": "flag", ...metadata }] }` — both string and object forms per flag.
  - Trims + deduplicates keys, preserves first-seen order.
  - Encodes per-flag metadata objects into a compact JSON string stored on `ConfigModel.sdkRemoteFeatureFlagMetadata`; exposes a decode helper that returns `Map<String, JsonObject>` for consumers.
  - Tolerates missing `features`, wrong root type, and malformed entries (drops them, keeps valid ones).

### Config + foreground refresh
- `ConfigModel` gains `sdkRemoteFeatureFlags: List<String>` and `sdkRemoteFeatureFlagMetadata: String?`.
- `ConfigModelChangeTags.REMOTE_FEATURE_FLAGS` scopes in-place writes so a stale full-model hydrate can't overwrite a fresh Turbine result.
- `FeatureFlagsRefreshService` (`IStartableService`, `IApplicationLifecycleHandler`, `ISingletonModelStoreChangeHandler<ConfigModel>`):
  - Kicks off a fetch immediately on `onFocus` (fires on subscribe), then every **8 minutes** while foregrounded.
  - Cancels the poll loop on `onUnfocused` and on app-id changes; re-arms when `ConfigModel` is rehydrated / replaced.
  - Coroutine-cancellation safe (re-throws `CancellationException`, swallows everything else with a WARN).

### FeatureManager integration
- `FeatureFlag` enum entries now carry a `FeatureActivationMode`:
  - `IMMEDIATE` — in-memory state + side effects applied as soon as remote config changes.
  - `APP_STARTUP` — value is persisted to `ConfigModel` the moment it arrives, but the in-memory `featureStates[feature]` and side effects are **not** flipped mid-run; the new value is latched on the next cold start. An `INFO` breadcrumb records the pending change.
- `SDK_BACKGROUND_THREADING` is wired as `APP_STARTUP` so `ThreadingMode` can't swap mid-session.
- `IFeatureManager.remoteFeatureFlagMetadata()` exposes the decoded `Map<String, JsonObject>` for downstream consumers (e.g. nested Turbine config blocks).
- `FeatureManager.init` reads the cached `ConfigModel` with `applyNextRunOnlyFeatures = true`, so any value written during the previous run is activated before the rest of the SDK starts consuming feature state.

### DI / wiring
- `CoreModule` registers `FeatureFlagsBackendService` and `FeatureFlagsRefreshService`.
- `build.gradle` picks up `kotlinx-serialization-json` for the parser.

## Test results

Run locally with `./gradlew :OneSignal:core:testOriginalUnitTest` — all green:

| Suite | Tests |
|---|---|
| `FeatureFlagsBackendServiceTest` | 4 / 4 |
| `FeatureFlagsBackendServiceTests` (logging + outcome behavior) | 8 / 8 |
| `FeatureFlagsJsonParserTests` | 13 / 13 |
| `TurbineSdkFeatureFlagsPathTest` | 3 / 3 |
| `FeatureManagerTests` (4 new covering remote flags + metadata, 4 preexisting) | 8 / 8 |
| `FeatureFlagTests` | 2 / 2 |

Notable coverage:
- 4xx / 5xx / `statusCode=0` / 200-with-bad-JSON all route to `Unavailable` with the right log level and body snippet; snippet capped at 200 chars; newlines/CRs flattened so logcat stays single-line.
- JSON parser handles mixed string + object entries, deduplicates, trims whitespace, drops malformed entries, round-trips metadata through the encode/decode pair.
- `FeatureManager` honors `APP_STARTUP` (latches current run, persists new value for next run) and `IMMEDIATE` (applies in place); `remoteFeatureFlagMetadata()` returns the decoded map or `null` when unset.
- Turbine path segment rejects empty / traversal / illegal characters; encoded identically for valid labels.

## Test plan

- [ ] CI passes (currently failing on `spotlessKotlinCheck` — needs `./gradlew :OneSignal:core:spotlessApply` for import ordering in `FeatureFlagsRefreshService.kt` / `FeatureManagerTests.kt` and brace indent in `FeatureManager.kt`).
- [x] Unit tests for backend service, JSON parser, refresh path, and `FeatureManager`.
- [ ] Manual verification: cold start with a Turbine-enabled app id returns flags and `FeatureManager.isEnabled(...)` reflects them on the next run; mid-session flips for `APP_STARTUP` flags are persisted but not applied until the next process start.


nan-li:
@claude review

OneSignalSDK/build.gradle
```diff
@@ -17,6 +17,7 @@ buildscript {
         kotlinVersion = '1.9.25'
         dokkaVersion = '1.9.10'
         coroutinesVersion = '1.7.3'
+        kotlinxSerializationJsonVersion = '1.6.3'
```
fadi-george, line 20: do we need this change?

abdulraqeeb33: yes, so i shifted away from using org.json.JsonObject to using kotlinxSerializationJson. This will help us in KMP as it needs to be pure kotlin.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
```diff
@@ -0,0 +1,122 @@
...
+            return
+        }
+
+        val tag = ConfigModelChangeTags.REMOTE_FEATURE_FLAGS
+        current.setListProperty(ConfigModel::sdkRemoteFeatureFlags.name, result.enabledKeys, tag)
+        current.setOptStringProperty(ConfigModel::sdkRemoteFeatureFlagMetadata.name, newMetaString, tag)
+    }
+
+    companion object {
+        private const val REFRESH_INTERVAL_MS = 600_000L
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. `IFeatureFlagsBackendService.fetchRemoteFeatureFlags` now returns a sealed `RemoteFeatureFlagsFetchOutcome` (`Success`/`Unavailable`); `fetchAndApply` early-returns on `Unavailable` so cached flags are not overwritten on HTTP failure, invalid version, or JSON parse failure. A legitimate server response with an empty `features` array still comes back as `Success(RemoteFeatureFlagsResult(emptyList(), null))` and is applied normally.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/features/FeatureManager.kt
```diff
@@ -31,13 +42,24 @@ internal class FeatureManager(
...
     ) {
         Logging.debug("OneSignal: FeatureManager.onModelReplaced(tag=$tag)")
-        if (tag == ModelChangeTags.HYDRATE || tag == ModelChangeTags.NORMAL) {
+        if (tag == ModelChangeTags.HYDRATE ||
+            tag == ModelChangeTags.NORMAL ||
+            tag == ConfigModelChangeTags.REMOTE_FEATURE_FLAGS
+        ) {
             try {
                 refreshEnabledFeatures(model, applyNextRunOnlyFeatures = false)
             } catch (t: Throwable) {
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. The `REMOTE_FEATURE_FLAGS` branch was removed from `FeatureManager.onModelReplaced`; remote-flag refreshes correctly flow through `onModelUpdated` when `fetchAndApply` calls `setListProperty`/`setOptStringProperty` on the live model.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
```diff
@@ -0,0 +1,122 @@
...
+                    val appId = configModelStore.model.appId
+                    if (appId.isNotEmpty()) {
+                        try {
+                            fetchAndApply(appId)
+                        } catch (t: Throwable) {
+                            Logging.warn("FeatureFlagsRefreshService: fetch failed", t)
+                        }
+                    }
+                    delay(REFRESH_INTERVAL_MS)
+                }
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. The polling loop now explicitly rethrows `CancellationException` before the generic `catch (e: Exception)`, so normal unfocus/restart cancellation no longer produces spurious warning logs.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
```diff
@@ -0,0 +1,119 @@
...
+    IApplicationLifecycleHandler,
+    ISingletonModelStoreChangeHandler<ConfigModel> {
+    private var pollJob: Job? = null
+
+    override fun start() {
+        applicationService.addApplicationLifecycleHandler(this)
+        configModelStore.subscribe(this)
+        if (applicationService.isInForeground) {
+            onFocus(firedOnSubscribe = true)
+        }
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 3720ec5. The explicit `onFocus` call was removed from `start()`; we rely on `addApplicationLifecycleHandler` firing `onFocus(true)` when an Activity is already in the foreground.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
```diff
@@ -0,0 +1,119 @@
...
+                    if (appId.isNotEmpty()) {
+                        try {
+                            fetchAndApply(appId)
+                        } catch (t: Throwable) {
+                            Logging.warn("FeatureFlagsRefreshService: fetch failed", t)
+                        }
+                    }
+                    delay(REFRESH_INTERVAL_MS)
+                }
+            }
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 3720ec5 and 64a10683. The poller now launches via `OneSignalDispatchers.launchOnIO` (no blanket catch) and the inner `try/catch` explicitly rethrows `CancellationException` before `catch (e: Exception)`, so normal unfocus/restart cancellation does not produce warning logs.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
```diff
@@ -0,0 +1,119 @@
...
+                    if (!applicationService.isInForeground) {
+                        break
+                    }
+                    val appId = configModelStore.model.appId
+                    if (appId.isNotEmpty()) {
+                        try {
+                            fetchAndApply(appId)
+                        } catch (t: Throwable) {
+                            Logging.warn("FeatureFlagsRefreshService: fetch failed", t)
+                        }
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. The inner catch was narrowed: `catch (e: CancellationException) { throw e }` runs first, then `catch (e: Exception)` handles actual fetch errors. The coroutine now cancels cleanly on unfocus.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/features/FeatureManager.kt
```diff
@@ -66,7 +91,12 @@ internal class FeatureManager(
         model: ConfigModel,
         applyNextRunOnlyFeatures: Boolean,
     ) {
-        val enabledFeatureKeys = (model.features + localFeatureOverrides).toSet()
+        val enabledFeatureKeys =
+            (
+                model.features +
+                    model.sdkRemoteFeatureFlags +
+                    localFeatureOverrides
+            ).toSet()
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. `refreshEnabledFeatures` now canonicalizes all unioned keys (from `features`, `sdkRemoteFeatureFlags`, and `localFeatureOverrides`) to lowercase via `canonicalizeFeatureKey` before `FeatureFlag.isEnabledIn`. `FeatureFlag.key` values are regex-validated lowercase in `FeatureFlagTests`, so legacy cached keys from older SDK versions can't latch the wrong value.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModelChangeTags.kt
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * [ModelChangeTags][com.onesignal.common.modeling.ModelChangeTags]-style values used only for
+ * [ConfigModel] / [ConfigModelStore] replace notifications.
+ */
+internal object ConfigModelChangeTags {
+    /**
+     * Remote feature-flags HTTP endpoint updated [ConfigModel] SDK flag fields only
+     * ([sdkRemoteFeatureFlags], [sdkRemoteFeatureFlagMetadata]).
+     */
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. The KDoc now uses fully-qualified `[ConfigModel.sdkRemoteFeatureFlags]` and `[ConfigModel.sdkRemoteFeatureFlagMetadata]` links.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/features/FeatureFlag.kt
```diff
@@ -17,21 +17,19 @@ internal enum class FeatureActivationMode {
...
 internal enum class FeatureFlag(
     val key: String,
     val activationMode: FeatureActivationMode
 ) {
     // Threading mode is selected once per app startup to avoid mixed-mode behavior mid-session.
     //
-    // Naming convention:
-    //   SDK_<MMmmpp>_<FEATURE_DESCRIPTION>
+    // Remote key (lowercase) must match backend / ConfigCat flag id.
     //
```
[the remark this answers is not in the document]
abdulraqeeb33: Addressed in 64a10683. The ConfigCat reference was removed from the `FeatureFlag` comment; it now says "Remote key (lowercase) must match backend / Turbine flag id." No `ConfigCat` references remain in the codebase.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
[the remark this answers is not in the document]
sherwinski: @abdulraqeeb33 I can't find that commit (64a10683), did it ever get pushed up?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/features/FeatureFlag.kt
```diff
@@ -17,21 +17,19 @@ internal enum class FeatureActivationMode {
...
-    // Example:
-    //   SDK_050800_BACKGROUND_THREADING
-    //
-    SDK_050800_BACKGROUND_THREADING(
-        "SDK_050800_BACKGROUND_THREADING",
```
sherwinski, line 34: Does this need a migration path? Is there any downside to just dropping it for existing users?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
```diff
@@ -0,0 +1,133 @@
...
+            return
+        }
+        if (model.appId.isNotEmpty() && applicationService.isInForeground) {
+            restartForegroundPolling()
+        }
+    }
+
+    private fun restartForegroundPolling() {
+        pollJob?.cancel()
+        pollJob =
```
sherwinski: do these need to be synchronized/made atomic? I'm wondering what would happen if `onFocus()` and `onModeulUpdated()` get invoked in quick succession?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/impl/FeatureFlagsJsonParser.kt
```diff
@@ -0,0 +1,150 @@
...
+ * Payload must be valid JSON (e.g. a comma between `"features": [...]` and the next property).
+ *
+ * Flag ids from the `features` array are normalized with [Char.lowercaseChar] (Unicode case mapping, no
+ * `java.util.Locale`) so they align with [com.onesignal.core.internal.features.FeatureFlag.key]. Sibling
+ * objects are resolved with exact key, then canonical key, then a case-insensitive match against other root
+ * properties (excluding `features`).
+ *
+ * Uses only Kotlin stdlib + kotlinx.serialization (Kotlin Multiplatform-friendly).
+ */
+internal object FeatureFlagsJsonParser {
```
nan-li, line 30: nit: the method implementation and kdocs in this class are hard to grasp

the metadata is speculative and no install would behave different if the backend started sending this, so users would still have to upgrade, is storing this is meant to prime the SDK on a potential future upgrade?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
```diff
@@ -300,6 +300,25 @@ class ConfigModel : Model() {
             setListProperty(::features.name, value)
         }
 
+    /**
+     * Feature keys from the dedicated SDK feature-flags HTTP endpoint (see [com.onesignal.core.internal.backend.IFeatureFlagsBackendService]).
+     * Unioned with [features] for [com.onesignal.core.internal.features.FeatureManager].
```
nan-li: what are we doing with `features`, do we need to maintain both properties

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModelChangeTags.kt
```diff
@@ -0,0 +1,13 @@
+package com.onesignal.core.internal.config
+
+/**
+ * [ModelChangeTags][com.onesignal.common.modeling.ModelChangeTags]-style values used only for
+ * [ConfigModel] / [ConfigModelStore] replace notifications.
+ */
+internal object ConfigModelChangeTags {
+    /**
+     * Remote feature-flags HTTP endpoint updated [ConfigModel] SDK flag fields only
```
nan-li: nit: some of the kdocs are a little confusingly worded?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/features/FeatureFlag.kt
nan-li: we never released a 5.8.x based on main

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/backend/impl/FeatureFlagsBackendServiceTests.kt
```diff
@@ -0,0 +1,145 @@
...
+ * Verifies that [FeatureFlagsBackendService] surfaces enough diagnostic information for a
+ * developer to understand why a remote feature-flags fetch returned
+ * [RemoteFeatureFlagsFetchOutcome.Unavailable]. In particular:
+ * - 4xx failures (e.g. 403 Forbidden for apps not enrolled in Turbine) should log at [LogLevel.WARN]
+ *   and include a body snippet so the Turbine error payload shows up in logcat by default.
+ * - 5xx / transient failures should stay at [LogLevel.DEBUG] to avoid noisy logs for flaky networks.
+ * - A 200 with an unexpected body shape should log at [LogLevel.WARN] with a body snippet since it
+ *   indicates a real contract/config issue that would otherwise be silent.
+ */
+class FeatureFlagsBackendServiceTests : FunSpec({
```
nan-li, line 27: There's FeatureFlagsBackendServiceTest and FeatureFlagsBackendServiceTests, what is the difference?

OneSignalSDK/onesignal/core/build.gradle
```diff
@@ -1,6 +1,7 @@
 plugins {
     id 'com.android.library'
     id 'kotlin-android'
+    id 'org.jetbrains.kotlin.plugin.serialization'
```
nan-li, line 4: do we need this one?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/features/FeatureFlag.kt
abdulraqeeb33: we have neither deployed `SDK_050800_BACKGROUND_THREADING` in the backend nor on the SDK side.
so there is no need of migration.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/FeatureFlagsRefreshService.kt
abdulraqeeb33: good call. Claude also mentioned the same. thanks

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/impl/FeatureFlagsJsonParser.kt
abdulraqeeb33: Updated the documentation for ease of reading.

The metadata will contain values that we want to change on the SDK without having to release a new sdk.

ex: In the future, maybe we want to increase/decrease the threads spawned for the background threading.
So from config cat we could pass the int value, maybe 3 or 4 or 5... this will be read by the SDK and used appropriately.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
abdulraqeeb33: yes we do, features has logging level and we cant get rid of it as there SDK versions that are already out there. We could consider moving to this latest way but once we have this settled.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModelChangeTags.kt
abdulraqeeb33: updated.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/backend/impl/FeatureFlagsBackendServiceTests.kt
abdulraqeeb33: Combined them.

OneSignalSDK/onesignal/core/build.gradle
abdulraqeeb33: We use kotlinx.serialization for JSON so the same types and parser can live in shared Kotlin. org.json is fine on Android/JVM but isn’t a portable choice for KMP common code; kotlinx.serialization is.”

sherwinski (approved):

OneSignalSDK/onesignal/core/build.gradle
nan-li: I think that comes from org.jetbrains.kotlinx:kotlinx-serialization-json below right?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt
nan-li: Does `features` include logging? I don't see it on [ConfigModel](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.7.7/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/ConfigModel.kt) in 5.7.7, I don't think we shipped `features` property on ConfigModel yet: https://github.com/OneSignal/OneSignal-Android-SDK/commit/c7794754fcbe7c8084901b52eb935ec06c426232

abdulraqeeb33: @nan-li you are right. we dont need features. Logging was RemoteLoggingParams

fadi-george (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/ConfigModelStoreListener.kt
```diff
@@ -95,7 +95,6 @@ internal class ConfigModelStoreListener(
                     params.locationShared?.let { config.locationShared = it }
                     params.requiresUserPrivacyConsent?.let { config.consentRequired = it }
                     params.opRepoExecutionInterval?.let { config.opRepoExecutionInterval = it }
-                    config.features = params.features
                     params.influenceParams.notificationLimit?.let { config.influenceParams.notificationLimit = it }
```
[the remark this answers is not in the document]
abdulraqeeb33: Nice catch — confirmed and fixed in 87868520.

Went with a scoped re-copy: right before `replace()` we re-read `sdkRemoteFeatureFlags` / `sdkRemoteFeatureFlagMetadata` from the live model back into the snapshot, so any in-place writes from `FeatureFlagsRefreshService` that landed during the params fetch survive `Model.initializeFromModel`'s `data.clear() + putAll()`.

I kept it limited to those two fields rather than rewriting `fetchParams` into in-place updates — the snapshot+replace pattern is shared with the rest of the params payload, and a broader rework feels like its own change.

Regression test in `ConfigModelStoreListenerTests` reproduces the race deterministically: a getter side-effect on `params.locationShared` writes to the live model after `fetchParams` snapshots but before it calls `replace()`. Fails on the pre-fix code (live ends up `[]` / `null`), passes after.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/backend/impl/FeatureFlagsJsonParser.kt
```diff
@@ -0,0 +1,146 @@
...
+                    ?.content
+                    ?.trim()
+                    ?.takeIf { it.isNotEmpty() }
+                    ?.let { raw -> raw to canonicalFeatureFlagId(raw) }
+            }.distinctBy { it.second }
+
+        if (flagEntries.isEmpty()) {
+            return RemoteFeatureFlagsResult(emptyList(), null)
+        }
+
```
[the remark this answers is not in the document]
abdulraqeeb33: Confirmed, applied in 87868520.

Took the suggested two-line fix in `parseRootStrict`: when `flagEntries` is empty we now return `Success(empty)` only if the original `features` array was also empty (the authoritative empty-config case). A non-empty array that filtered down to empty due to element-type violations returns `null` instead, which surfaces as `RemoteFeatureFlagsFetchOutcome.Unavailable` upstream so `FeatureFlagsRefreshService` preserves the cached list rather than overwriting it with `[]`.

Added coverage in `FeatureFlagsJsonParserTests` for the pathological shapes you called out (`[1,2,3]`, `[null]`, `[{}]`, `[[]]`, `[true,false]`, `[""]`, `["   "]`) plus a guard test pinning the existing tolerance contract — `["good", 1, null, {}, ""]` still returns `Success(["good"])`.

nan-li (approved):

PR #2614, opened 2026-04-15
feat: hash PII (email/SMS) in SharedPreferences at rest

sherwinski:
# Description
## One Line Summary
Hash email and SMS subscription addresses (SHA-256) before writing to SharedPreferences so PII is not stored in plain text on device.

### Motivation
Email addresses and phone numbers are stored in plain text in `OneSignal.xml` under the `MODEL_STORE_subscriptions` key. On rooted devices or via ADB backup, this PII is directly readable. This PR hashes those values at the serialization boundary so the on-disk representation is opaque.

### Scope
- **Email and SMS subscription addresses only.** Push tokens are not hashed — they are not PII and are required for push delivery.
- **On-disk only.** The in-memory `SubscriptionModel.address` always holds the raw value at runtime (after server hydration). Network requests continue to send the real email/phone to the backend API.
- **No public API changes.** The `IEmailSubscription.email` and `ISmsSubscription.number` getters return the raw value post-hydration, or an empty string during the brief cold-start window before `RefreshUserOperationExecutor` restores the real value.

### How it works

1. **`PIIHasher`** — new utility: deterministic SHA-256 hashing and `isHashed()` detection (64-char lowercase hex).
2. **`ModelStore.transformJsonForPersistence()`** — new protected hook called during `persist()`. Default is identity; subclasses can override to transform JSON before it hits SharedPreferences.
3. **`SubscriptionModelStore`** — overrides the hook to hash the `address` field for EMAIL and SMS subscriptions. Skips PUSH. Skips already-hashed values (idempotency guard against double-hashing on restart).
4. **Hash-aware removal/lookup** — `SubscriptionManager.removeEmailSubscription()`, `removeSmsSubscription()`, `SubscriptionList.getByEmail()`, and `getBySMS()` now compare the raw input against both the raw and hashed `model.address`, so lookups work whether the model is pre- or post-hydration.
5. **Log redaction** — `SubscriptionManager.addSubscriptionToModels()` debug log now prints the hash instead of the raw address.
6. Public getter safety — EmailSubscription.email and SmsSubscription.number return "" if the underlying address is detected as a hash (cold-start to hydration window).

### Trade-off
Between cold start and server hydration (~0.5–1s), `model.address` contains a hash loaded from SharedPreferences. During this window, public getters return an empty string. Once `RefreshUserOperationExecutor` hydrates the model from the server response, the real value is restored and persisted as a new hash.

Note: The `IEmailSubscription.email` and `ISmsSubscription.number` getters exist but are not reachable through the current public API (`IUserManager` does not expose email/SMS subscription collections). If a future API exposes them, they will return an empty string during the brief cold-start to hydration window.

# Testing
## Unit testing
All 560 existing unit tests pass. The `ModelingTests` deadlock test exercised the new `transformJsonForPersistence` path — an initial NPE (accessing `model.type` on an uninitialized model) was caught and fixed by reading the type from the JSON object instead.

New tests were added for the following: 
- **PIIHasherTests**: hash output format, determinism, known digest,
  isHashed detection for valid/invalid inputs
- **SubscriptionModelStoreTests**: persist hashes email/SMS but not push,
  idempotent on already-hashed values, in-memory model stays raw
- **SubscriptionManagerTests**: hash-aware removal for email/SMS when
  model.address is hashed (pre-hydration), public getter returns
  empty string for hashed addresses, getByEmail/getBySMS find
  subscriptions with hashed addresses

## Manual testing
Tested on a Pixel emulator (API 35) with the OneSignal demo app:

1. **Fresh email/SMS add** — Added email and SMS via demo app. Dumped SharedPreferences → addresses are 64-char SHA-256 hashes. Push token is unchanged (not hashed).
2. **Network requests** — Verified via logcat that API requests to OneSignal backend contain the raw email/phone, not hashes.
3. **Post-hydration removal** — Removed email after hydration completed → subscription removed from SharedPreferences successfully.
4. **Pre-hydration removal** — Added `removeEmail()` call immediately after `initWithContext` (before `RefreshUserOperationExecutor` runs). The hash-aware comparison matched the raw input against the hashed stored value → subscription removed correctly.
5. **Upgrade migration (plain text → hashed)** — Installed `5.7-main` build, added email (stored as plain text). Installed feature branch build over the top (no uninstall). On first launch, the plain-text address was automatically hashed during the first persist cycle.
7. **Removal after migration** — `removeEmail()` with the raw email successfully removed the now-hashed subscription.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/PIIHasher.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/PIIHasher.kt
@@ -1,0 +1,23 @@
+package com.onesignal.common
+
+import java.security.MessageDigest
+
+/**
+ * Deterministic SHA-256 hashing for PII fields (email, phone number) so that
+ * sensitive data stored in SharedPreferences is not readable in plain text on
+ * rooted devices or via ADB backup.
+ *
+ * The hash is hex-encoded and always 64 characters long.
+ */
+object PIIHasher {
+    private const val SHA256_HEX_LENGTH = 64
+    private val SHA256_HEX_REGEX = Regex("^[a-f0-9]{$SHA256_HEX_LENGTH}$")
+
+    fun hash(value: String): String {
+        val digest = MessageDigest.getInstance("SHA-256")
+        val bytes = digest.digest(value.toByteArray(Charsets.UTF_8))
+        return bytes.joinToString("") { "%02x".format(it) }
+    }
+
+    fun isHashed(value: String): Boolean = SHA256_HEX_REGEX.matches(value)
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/PIIHasher.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/PIIHasher.kt
@@ -13,11 +13,13 @@ object PIIHasher {
     private const val SHA256_HEX_LENGTH = 64
     private val SHA256_HEX_REGEX = Regex("^[a-f0-9]{$SHA256_HEX_LENGTH}$")
 
+    /** Returns the lowercase hex-encoded SHA-256 hash of [value]. */
     fun hash(value: String): String {
         val digest = MessageDigest.getInstance("SHA-256")
         val bytes = digest.digest(value.toByteArray(Charsets.UTF_8))
         return bytes.joinToString("") { "%02x".format(it) }
     }
 
+    /** Returns `true` if [value] looks like a 64-char lowercase hex SHA-256 digest. */
     fun isHashed(value: String): Boolean = SHA256_HEX_REGEX.matches(value)
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/EmailSubscription.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/EmailSubscription.kt
@@ -1,5 +1,6 @@
 package com.onesignal.user.internal
 
+import com.onesignal.common.PIIHasher
 import com.onesignal.user.internal.subscriptions.SubscriptionModel
 import com.onesignal.user.subscriptions.IEmailSubscription
 
@@ -7,5 +8,8 @@ internal class EmailSubscription(
     model: SubscriptionModel,
 ) : Subscription(model), IEmailSubscription {
     override val email: String
-        get() = model.address
+        get() {
+            val address = model.address
+            return if (PIIHasher.isHashed(address)) "" else address
+        }
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
@@ -1,5 +1,6 @@
 package com.onesignal.user.internal
 
+import com.onesignal.common.PIIHasher
 import com.onesignal.user.internal.subscriptions.SubscriptionModel
 import com.onesignal.user.subscriptions.ISmsSubscription
 
@@ -7,5 +8,8 @@ internal class SmsSubscription(
     model: SubscriptionModel,
 ) : Subscription(model), ISmsSubscription {
     override val number: String
-        get() = model.address
+        get() {
+            val address = model.address
+            return if (PIIHasher.isHashed(address)) "" else address
+        }
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
@@ -1,8 +1,10 @@
 package com.onesignal.user.internal.subscriptions
 
+import com.onesignal.common.PIIHasher
 import com.onesignal.common.modeling.ModelChangeTags
 import com.onesignal.common.modeling.SimpleModelStore
 import com.onesignal.core.internal.preferences.IPreferencesService
+import org.json.JSONObject
 
 open class SubscriptionModelStore(prefs: IPreferencesService) : SimpleModelStore<SubscriptionModel>({
     SubscriptionModel()
@@ -31,5 +33,18 @@ open class SubscriptionModelStore(prefs: IPreferencesService) : SimpleModelStore
             }
             super.replaceAll(models, tag)
         }
+    }
+
+    override fun transformJsonForPersistence(
+        model: SubscriptionModel,
+        json: JSONObject,
+    ): JSONObject {
+        if (model.type == SubscriptionType.PUSH) return json
+
+        val address = json.optString("address", "")
+        if (address.isNotEmpty() && !PIIHasher.isHashed(address)) {
+            json.put("address", PIIHasher.hash(address))
+        }
+        return json
     }
 }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
@@ -39,7 +39,8 @@ open class SubscriptionModelStore(prefs: IPreferencesService) : SimpleModelStore
         model: SubscriptionModel,
         json: JSONObject,
     ): JSONObject {
-        if (model.type == SubscriptionType.PUSH) return json
+        val type = json.optString("type", "")
+        if (type.isEmpty() || type == SubscriptionType.PUSH.toString()) return json
 
         val address = json.optString("address", "")
         if (address.isNotEmpty() && !PIIHasher.isHashed(address)) {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
@@ -5,6 +5,7 @@ import com.onesignal.common.AndroidUtils
 import com.onesignal.common.DeviceUtils
 import com.onesignal.common.IDManager
 import com.onesignal.common.OneSignalUtils
+import com.onesignal.common.PIIHasher
 import com.onesignal.common.events.EventProducer
 import com.onesignal.common.modeling.IModelStoreChangeHandler
 import com.onesignal.common.modeling.ModelChangedArgs
@@ -93,7 +94,11 @@ internal class SubscriptionManager(
     }
 
     override fun removeEmailSubscription(email: String) {
-        val subscriptionToRem = subscriptions.emails.firstOrNull { it is EmailSubscription && it.email == email }
+        val hashedEmail = PIIHasher.hash(email)
+        val subscriptionToRem =
+            subscriptions.emails.firstOrNull {
+                it is EmailSubscription && (it.model.address == email || it.model.address == hashedEmail)
+            }
 
         if (subscriptionToRem != null) {
             removeSubscriptionFromModels(subscriptionToRem)
@@ -101,7 +106,11 @@ internal class SubscriptionManager(
     }
 
     override fun removeSmsSubscription(sms: String) {
-        val subscriptionToRem = subscriptions.smss.firstOrNull { it is SmsSubscription && it.number == sms }
+        val hashedSms = PIIHasher.hash(sms)
+        val subscriptionToRem =
+            subscriptions.smss.firstOrNull {
+                it is SmsSubscription && (it.model.address == sms || it.model.address == hashedSms)
+            }
 
         if (subscriptionToRem != null) {
             removeSubscriptionFromModels(subscriptionToRem)
@@ -113,7 +122,7 @@ internal class SubscriptionManager(
         address: String,
         status: SubscriptionStatus? = null,
     ) {
-        Logging.log(LogLevel.DEBUG, "SubscriptionManager.addSubscription(type: $type, address: $address)")
+        Logging.log(LogLevel.DEBUG, "SubscriptionManager.addSubscription(type: $type, address: ${PIIHasher.hash(address)})")
 
         val subscriptionModel = SubscriptionModel()
         subscriptionModel.id = IDManager.createLocalId()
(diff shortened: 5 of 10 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/PIIHasher.kt
```diff
@@ -0,0 +1,23 @@
...
+import java.security.MessageDigest
+
+/**
+ * Deterministic SHA-256 hashing for PII fields (email, phone number) so that
+ * sensitive data stored in SharedPreferences is not readable in plain text on
+ * rooted devices or via ADB backup.
+ *
+ * The hash is hex-encoded and always 64 characters long.
+ */
+object PIIHasher {
```
sherwinski, line 12: I'm open to a different/more general class name here.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
```diff
@@ -1,11 +1,15 @@
...
 import com.onesignal.user.subscriptions.ISmsSubscription
 
 internal class SmsSubscription(
     model: SubscriptionModel,
 ) : Subscription(model), ISmsSubscription {
     override val number: String
-        get() = model.address
+        get() {
+            val address = model.address
+            return if (PIIHasher.isHashed(address)) "" else address
```
fadi-george, line 13: why return empty, would that affect getBySMS check?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/EmailSubscription.kt
```diff
@@ -1,11 +1,15 @@
 package com.onesignal.user.internal
 
+import com.onesignal.common.PIIHasher
```
fadi-george, line 3: why return empty, would that affect getByEmail check?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
```diff
@@ -113,7 +122,7 @@ internal class SubscriptionManager(
         address: String,
         status: SubscriptionStatus? = null,
     ) {
-        Logging.log(LogLevel.DEBUG, "SubscriptionManager.addSubscription(type: $type, address: $address)")
+        Logging.log(LogLevel.DEBUG, "SubscriptionManager.addSubscription(type: $type, address: ${PIIHasher.hash(address)})")
```
fadi-george: dont need to hash push token,
also how does this work?
will api requests be done with original values then get hashed on success?

sherwinski:
## Manual Testing Steps

### Setup

```bash
# Build and install the demo app (from OneSignalSDK/ directory)
./gradlew :app:installGmsDebug

# Useful commands for inspecting state
adb shell run-as com.onesignal.sdktest cat /data/data/com.onesignal.sdktest/shared_prefs/OneSignal.xml | rg address
adb shell run-as com.onesignal.sdktest cat /data/data/com.onesignal.sdktest/shared_prefs/OneSignal.xml | rg EMAIL
adb logcat -s "OneSignal" | grep -E "SDK-4202|addSubscription|RefreshUserOperationExecutor|Request Sent"

# Force-kill the app
adb shell am force-stop com.onesignal.sdktest

# Relaunch the app
adb shell am start -n com.onesignal.sdktest/.ui.main.MainActivity
```

---

### Core Hashing

- [x] **Email address is hashed in SharedPreferences**
  Added `test@example.com` via demo app. Dumped SharedPreferences:
  ```
  "address":"973dfe463ec85785f5f95af5ba3906eedb2d931c24e69824a89ea65dba4e813b"
  ```
  64-char SHA-256 hex hash confirmed.

- [x] **Push token is not hashed**
  Same dump shows push token stored as raw FCM token:
  ```
  "type":"PUSH","address":"dz1A0qydQGCYM9dDgo6rB_:APA91bEq..."
  ```

- [x] **Debug log shows hash, not raw email**
  Logcat output for `addSubscription` shows the hashed address, not the plain-text email.

- [x] **Network call sends raw email to backend**
  Logcat shows the POST/PATCH request body contains `"token":"test@example.com"` (the raw value), not a hash.

- [x] **Backend accepts the request (202)**
  Server responded with `STATUS: 202` for the subscription creation.

---

### Removal — Post-Hydration (same session)

- [x] **Remove after add (same session)**
  Added an email, then removed it via the demo app UI. Confirmed the EMAIL entry disappeared from SharedPreferences:
  ```bash
  adb shell run-as com.onesignal.sdktest cat /data/data/com.onesignal.sdktest/shared_prefs/OneSignal.xml | rg EMAIL
  # (no output — EMAIL subscription removed)
  ```

---

### Removal — Pre-Hydration (cold start, hash-aware comparison)

- [x] **Remove before hydration completes**
  Added `another1@email.com`, force-killed the app. Modified `MainApplication.kt` to call `OneSignal.User.removeEmail("another1@email.com")` immediately after `initWithContext()` (before `RefreshUserOperationExecutor` runs). Rebuilt and relaunched:
  ```bash
  ./gradlew :app:installGmsDebug
  adb shell am start -n com.onesignal.sdktest/.ui.main.MainActivity
  ```
  Dumped SharedPreferences — the EMAIL entry was successfully removed, confirming the hash-aware comparison in `removeEmailSubscription()` correctly matched the raw input against the hashed `model.address`.

---

### Cold Start + Hydration Cycle

- [x] **Hash persists across cold start**
  Added email, force-killed app, relaunched. Dumped SharedPreferences — address is still a 64-char hash (loaded from disk, not reverted to plain text).

- [x] **Hash remains consistent after hydration re-persist**
  After `RefreshUserOperationExecutor` hydrated the model with the raw value from the server, the next persist cycle re-hashed it. Dumped SharedPreferences — same 64-char hash as before (no double-hashing).

- [x] **Server response contains raw email**
  Logcat for the `GET /users/by/onesignal_id/...` response shows:
  ```
  "token":"getter@email.com"
  ```
  Confirming the server sends back the real email, which hydrates `subscriptionModel.address`.

---

### Upgrade Migration (plain text → hashed)

- [x] **Step 1 — Install base branch build (`5.7-main`), add email**
  ```bash
  git checkout 5.7-main
  cd OneSignalSDK && ./gradlew :app:installGmsDebug
  ```
  Added `upgrade@test.com` via demo app. Dumped SharedPreferences:
  ```
  "address":"upgrade@test.com"
  ```
  Plain text confirmed on the base branch.

- [x] **Step 2 — Install feature branch over the top (no uninstall)**
  ```bash
  git checkout sherwin/sdk-4202-sensitive-user-data-stored-in-plain-text-onesignalxml
  cd OneSignalSDK && ./gradlew :app:installGmsDebug
  ```
  Launched the app. The first persist cycle hashed the existing plain-text address. Dumped SharedPreferences:
  ```
  "address":"9b2506faa92a13822f34bdd50c78820b04bb310bb7b4f41775b4ac361558c9bd"
  ```
  Migration from plain text to hash confirmed.

- [x] **Step 3 — Removal works after migration**
  Removed `upgrade@test.com` via demo app. Dumped SharedPreferences:
  ```bash
  adb shell run-as com.onesignal.sdktest cat /data/data/com.onesignal.sdktest/shared_prefs/OneSignal.xml | rg EMAIL
  # (no output — EMAIL subscription removed)
  ```
  Hash-aware removal works correctly on migrated data.

---

### Not Tested / Known Limitations

- **Public getter pre-hydration window** — `IEmailSubscription.email` returns `""` when the underlying address is a hash. Could not be directly observed from the demo app because the public API (`IUserManager`) does not expose email/SMS subscription collections, and the hydration window is ~700ms (too fast for the demo UI to render).

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
```diff
@@ -113,7 +122,7 @@ internal class SubscriptionManager(
         address: String,
```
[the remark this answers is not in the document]
sherwinski: Addressed in [a8482de](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2614/commits/a8482de8cc825ae6f8855f1bcac432ff10def19a)

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
sherwinski: `getBySMS` reads `model.address` directly (not the public getter), so there's no impact. The empty return is to protect against leaking a hash to app developers during the cold-start-to-hydration window. I opted for this approach to signal to devs that the value isn't available yet as opposed to providing a confusing value.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/EmailSubscription.kt
sherwinski: same as above

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/impl/SubscriptionManager.kt
sherwinski: > dont need to hash push token

Addressed in a8482de

> will api requests be done with original values then get hashed on success?

Hashing only happens at the serialization boundary. `model.address` always holds the raw value, which is what gets read during network operations, so they always send the raw email/phone in that case.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/PIIHasher.kt
```diff
@@ -0,0 +1,23 @@
...
+    private const val SHA256_HEX_LENGTH = 64
+    private val SHA256_HEX_REGEX = Regex("^[a-f0-9]{$SHA256_HEX_LENGTH}$")
+
+    fun hash(value: String): String {
+        val digest = MessageDigest.getInstance("SHA-256")
+        val bytes = digest.digest(value.toByteArray(Charsets.UTF_8))
+        return bytes.joinToString("") { "%02x".format(it) }
```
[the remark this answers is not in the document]
abdulraqeeb33: I believe we can still stick to SHA-256 as we dont want to bind these keys to a device where it was created. If the user switches the device then it will be impossible to recover.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
```diff
@@ -32,4 +34,18 @@ open class SubscriptionModelStore(prefs: IPreferencesService) : SimpleModelStore
             super.replaceAll(models, tag)
         }
     }
+
+    override fun transformJsonForPersistence(
+        model: SubscriptionModel,
+        json: JSONObject,
+    ): JSONObject {
+        val type = json.optString("type", "")
+        if (type.isEmpty() || type == SubscriptionType.PUSH.toString()) return json
```
abdulraqeeb33, line 43: what if we hash the PUSH token as well? Would that cause us any issues?

fadi-george:
Ill defer for ARs final review

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
abdulraqeeb33: instead of this can we pass an Enum back.
Something like 

Hash("value")
Address("value")
Unknown

then its more clear and explicit

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
sherwinski: Yes, I think that would cause issues. On cold start (before server hydration), `model.address` holds whatever was loaded from SharedPreferences. If the push token were hashed, any network operation during that window would send the hash instead of the real FCM/HMS token, which could break push delivery.

Take [CreateSubscriptionOperation](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5f7187baee9bd09664e59adbd963d15ef1698fc5/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/listeners/SubscriptionModelStoreListener.kt#L29) for example, which sends `model.address` in the POST/PATCH request body to the API. If the push token were hashed on disk and loaded back as a hash on cold start, we would be sending the hashed value to the backend.

Is there a reason to hash the push token? It's not as personally-identifiable as email or phone number.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
sherwinski: I agree with you, but I'd want to clarify one thing first.

`ISmsSubscription` and `IEmailSubscription` expose public getters for email and sms, so passing an enum back would _technically_ be a breaking change. The wrinkle here is that the getters are not reachable because `IUserManager` doesn't expose them, although I believe we do access those values directly in the demo app.
Given this is the case, would we still be ok with that approach?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModelStore.kt
abdulraqeeb33: Ok makes sense. I was assuming that just like email and phone number these values are not really required and a one way hash would be applicable there as well. But we are using the token to Create the Subscription then yeah, we cant.

Yeah I was trying to push more so on the consistency side.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/SmsSubscription.kt
abdulraqeeb33: ah right, these Interface are exposed in the `IUserManager` - would cause a breaking change. I think we should be with this although its might not super clear IMHO for a developer.

Can we update documentation that states - 
Before hydration

- if null then this is what it means

After hydration
- if null then this is what it means

abdulraqeeb33 (approved):

PR #2600, opened 2026-04-07
fix: login race condition and JWT FAIL_UNAUTHORIZED blocking callers

abdulraqeeb33:
## Summary

- **Non-suspend `login()` race condition:** Split `LoginHelper` into `switchUser()` (synchronous local model swap) + `enqueueLogin()` (async network call). The non-suspend `login()` now calls `switchUser()` synchronously on the calling thread, then fires off `enqueueLogin()` in the background via `suspendifyOnIO` (FF ON) or `Thread { runBlocking }` (FF OFF). This ensures tags target the correct user without blocking the main thread or causing ANRs. `loginSuspend()` still uses the full `login()` which awaits the backend operation.
- **JWT `FAIL_UNAUTHORIZED` / `FAIL_PAUSE_OPREPO` hanging callers:** Wake `enqueueAndWait` waiters with `false` before re-queuing operations. Re-queued items use `waiter = null` so future retries (after `updateUserJwt`) don't reference stale waiters. This prevents `loginSuspend` from hanging forever when a JWT is invalid or the op repo pauses.

## Files Changed

- `LoginHelper.kt` — Split into `switchUser()` (instant, synchronous model swap under lock) + `enqueueLogin()` (suspend, network call) + `login()` (both, for `loginSuspend`)
- `OneSignalImp.kt` — `login()` calls `switchUser()` synchronously then dispatches `enqueueLogin()` respecting the FF flag; `logout()` unchanged
- `OperationRepo.kt` — `FAIL_UNAUTHORIZED` and `FAIL_PAUSE_OPREPO` branches wake waiters before re-queuing with `waiter = null`

## Test Results

All tests run on emulator (Medium Phone AVD - API 16) with fresh app data cleared between each run.

### Test code

**Non-suspend (called from `onCreate` on main thread):**
```kotlin
OneSignal.login("alice")            // or OneSignal.login("alice", VALID_JWT)
OneSignal.User.addTag("alice1", "alice1")

OneSignal.login("bob")              // or OneSignal.login("bob", INVALID_JWT)
OneSignal.User.addTag("bob2", "bob2")

OneSignal.login("chris")            // or OneSignal.login("chris", VALID_JWT)
OneSignal.User.addTag("chris3", "chris3")
```

**Suspend (called from `CoroutineScope(Dispatchers.IO)`):**
```kotlin
OneSignal.loginSuspend("alice")     // or OneSignal.loginSuspend("alice", VALID_JWT)
OneSignal.User.addTag("alice1", "alice1")

[...]

PR #2617, opened 2026-04-17
closed

nan-li:
## One Line Summary
Dedupe LoginUserOperation at enqueue time to prevent duplicate login ops from crashing the executor.

### Motivation
After #2600, `OneSignal.login()` was split into `switchUser()` (synchronous) and `enqueueLogin()` (async via `suspendifyOnIO`). This introduced a window between the two steps where `RecoverFromDroppedLoginBug.start()` can fire and observe:
- `externalId` is set (login was just called)
- `onesignalId` is a local ID (user not yet created on backend)
- No `LoginUserOperation` in the queue yet (enqueue is still pending)

This matches the "dropped login bug" criteria, so it enqueues its own recovery `LoginUserOperation`. Shortly after, the real `enqueueLogin()` adds another one. When both ops land in the queue they get grouped (same `modifyComparisonKey`), and `LoginUserOperationExecutor.createUser()` throws "Unrecognized operation" on the second `LoginUserOperation` in the batch — dropping both ops and logging "Could not login user". Before #2600, `login()` enqueued synchronously so `RecoverFromDroppedLoginBug` would always see the op already queued and skip.

### Scope
- `OperationRepo.internalEnqueue()` — added a second dedupe check: if a `LoginUserOperation` for the same `onesignalId` is already queued, drop the incoming one.
- The dedupe path wakes the dropped op's waiter with `true` so `loginSuspend()` callers don't hang on `enqueueAndWait.waitForWake()` forever.
- Not IV/JWT-specific — this race affects any login flow.

# Testing
## Manual testing
- Fresh install with cached anonymous user: call `OneSignal.login()` right after `initWithContext` → user created on backend, no "Unrecognized operation" error in logs.
- Verified the executor no longer receives a grouped `[login-user, create-subscription, login-user]` batch.

# Affected code checklist
   - [x] REST API requests

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -173,6 +173,19 @@ internal class OperationRepo(
                 return
             }
 
+            // Dedupe LoginUserOperation for the same user — prevents RecoverFromDroppedLoginBug
+            // and the real login() call from both enqueuing a login op during the timing window.
+            // Wake the waiter as if we succeeded, since the already-queued op will do the work
+            // and enqueueAndWait callers (e.g. loginSuspend) would otherwise hang forever.
+            val op = queueItem.operation
+            if (op is LoginUserOperation &&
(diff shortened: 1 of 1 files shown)


nan-li:
closing in favor of https://github.com/OneSignal/OneSignal-Android-SDK/pull/2618

PR #2618, opened 2026-04-17
fix: login/logout race causes subsequent calls to target previous user

nan-li:
# Description
## One Line Summary
**_(Not JWT related)_**
Implementation of https://github.com/OneSignal/OneSignal-Android-SDK/pull/2600

Fix a race in `OneSignal.login()` and `OneSignal.logout()` where SDK calls made immediately after either operation (e.g. `addTag`) landed on the previous user's identity.

### Motivation
Both `OneSignal.login()` and `OneSignal.logout()` scheduled the local user switch asynchronously via `suspendifyOnIO`, so the API call returned before the identity/properties/subscription models were updated. Calls made right after targeted the OLD user.

Example reproduction:
```kotlin
OneSignal.login("alice")
OneSignal.User.addTag("name_aa", "alice")    // lands on old user, not alice

OneSignal.login("bob")
OneSignal.User.addTag("name_bb", "bob")      // lands on alice (or whoever), not bob

OneSignal.logout()
OneSignal.User.addTag("post_logout", "val")  // lands on bob, not the new anonymous user
```

### Scope
Split both `LoginHelper.login()` and `LogoutHelper.logout()` into two phases each:
- **`switchUser()`** — synchronous under the login/logout lock. Swaps the local identity/properties/subscription models immediately so subsequent SDK calls see the new user.
- **`enqueueLogin()` / `enqueueLogout()`** — async. Enqueues the `LoginUserOperation` (and for login, awaits completion of the backend user creation).

`OneSignalImp.login()`, `loginSuspend()`, `logout()`, and `logoutSuspend()` all call `switchUser()` synchronously, then launch the enqueue in the coroutine scope.

Because `switchUser` returns before the enqueue runs, there's a brief window where `RecoverFromDroppedLoginBug` can observe the intermediate state (new `externalId`, local `onesignalId`, no `LoginUserOperation` in queue) and enqueue its own recovery login. When the real enqueue then adds another login op, the executor crashes on a grouped batch containing two `LoginUserOperation`s with "Unrecognized operation".

To handle this:
- `OperationRepo.internalEnqueue` dedupes `LoginUserOperation` by `onesignalId`. If a queued op already exists for the same user, the incoming op is dropped. The incoming op's `existingOnesignalId` is merged into the queued op first so the anonymous-user conversion link isn't lost (the recovery op always has `existingOnesignalId = null`, the real login op may carry the anon UUID needed for the backend merge).
- The incoming op's waiter is transferred to the queued op so `enqueueAndWait` callers (e.g. `loginSuspend`) receive the real execution result.
- `LoginUserOperation.existingOnesignalId` setter relaxed from `private` to `internal` to allow the in-place merge.

Also applies the `FAIL_PAUSE_OPREPO` waker fix from #2600: ops re-queued on pause detach their waiters (wake with `false`) so `enqueueAndWait` callers don't hang on a paused queue.

Related: #2600 made the same split on `feat/identity_verification_5.8`, but that branch hasn't merged. This port brings the fix to `main` so all users get the race fix in the next release, and extends the pattern symmetrically to logout.

# Testing
## Unit testing
- Updated `LoginHelperTests` to use the new `switchUser()` + `enqueueLogin()` split; existing tests pass.
- Updated `LogoutHelperTests` to use the new `switchUser()` + `enqueueLogout()` split; existing tests pass.
- Added a new `OperationRepoTests` case for the dedupe+merge path: seed a `LoginUserOperation(existingOnesignalId=null)` in the queue (simulates `RecoverFromDroppedLoginBug`), enqueue an incoming op with the same `onesignalId` and a non-null `existingOnesignalId`, verify the queued op is merged in place and the incoming op is dropped. Uses `waitForInternalEnqueue` helper for determinism.
- Updated the `loadSavedOperations` test to construct `LoginUserOperation` with full args (the previous no-args construction stopped working once the dedupe reads `onesignalId`).

## Manual testing
- Reproduced the original login bug in the demo app with back-to-back `login()` + `addTag()` calls — confirmed each tag landed on the wrong user before the fix.
- After the fix: each tag lands on the intended user.
- Verified the symmetric logout case: `logout()` + `addTag()` now attaches the tag to the new anonymous user instead of the previously logged-in one.
- Reproduced the RecoverFromDroppedLoginBug race on fresh install + quick `login()` call — confirmed the executor no longer crashes with "Unrecognized operation" and the anonymous-user conversion link is preserved.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -333,9 +333,15 @@ internal class OperationRepo(
                     Logging.error("Operation execution failed with eventual retry, pausing the operation repo: $operations")
                     // keep the failed operation and pause the operation repo from executing
                     paused = true
-                    // add back all operations to the front of the queue to be re-executed.
+                    // Unblock any enqueueAndWait callers so loginSuspend doesn't hang.
+                    ops.forEach { it.waiter?.wake(false) }
+                    // Re-queue with waiter = null: the operation is preserved for retry
+                    // on next cold start, but the original waiter is detached since it
+                    // was already woken above.
                     synchronized(queue) {
-                        ops.reversed().forEach { queue.add(0, it) }
+                        ops.reversed().forEach {
+                            queue.add(0, OperationQueueItem(it.operation, waiter = null, bucket = it.bucket, retries = it.retries))
+                        }
                     }
                 }
             }
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -32,7 +32,7 @@ internal class OperationRepo(
 ) : IOperationRepo, IStartableService {
     internal class OperationQueueItem(
         val operation: Operation,
-        var waiter: WaiterWithValue<Boolean>? = null,
+        var waiter: WaiterWithValue<Boolean>? = null, // waiter may transfer during operation de-dupe
         val bucket: Int,
         var retries: Int = 0,
     ) {
@@ -162,10 +162,7 @@ internal class OperationRepo(
                 return
             }
 
-            // Dedupe LoginUserOperation for the same user. Merge the incoming
-            // existingOnesignalId so the anon-user conversion isn't lost when
-            // RecoverFromDroppedLoginBug wins the race (it enqueues with null).
-            // Transfer the waiter so enqueueAndWait callers get the real result.
+            // Dedupe LoginUserOperation by onesignalId.
             val op = queueItem.operation
             if (op is LoginUserOperation) {
                 val existing =
@@ -174,12 +171,14 @@ internal class OperationRepo(
                     }
                 if (existing != null) {
                     val existingOp = existing.operation as LoginUserOperation
+                    // Preserve the anon-user conversion link if the queued op lacks it (e.g. RecoverFromDroppedLoginBug enqueued with null).
                     if (op.existingOnesignalId != null && existingOp.existingOnesignalId == null) {
                         Logging.debug("OperationRepo: internalEnqueue - merging existingOnesignalId=${op.existingOnesignalId} into queued LoginUserOperation for onesignalId: ${op.onesignalId}.")
                         existingOp.existingOnesignalId = op.existingOnesignalId
                     } else {
                         Logging.debug("OperationRepo: internalEnqueue - LoginUserOperation for onesignalId: ${op.onesignalId} already exists in the queue.")
                     }
+                    // Transfer the waiter so enqueueAndWait callers see the queued op's real execution result.
                     if (queueItem.waiter != null && existing.waiter == null) {
                         existing.waiter = queueItem.waiter
                     } else {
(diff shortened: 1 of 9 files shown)


nan-li:
@claude review

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -161,6 +162,31 @@ internal class OperationRepo(
...
+                        _operationModelStore.remove(queueItem.operation.id)
+                    }
+                    queueItem.waiter?.wake(true)
+                    return
+                }
+            }
+
             if (index != null) {
                 queue.add(index, queueItem)
             } else {
```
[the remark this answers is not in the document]
nan-li: Persist is async internally via `PreferencesService.doWorkAsync` so no disk I/O happens in the synchronized block, just a brief in-memory lock. Following the external pattern here adds ceremony (data class + out-of-lock juggling) without meaningful safety gain.

nan-li:
@claude review

nan-li:
@claude review

nan-li:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -161,6 +163,41 @@ internal class OperationRepo(
                 return
             }
 
+            // Dedupe LoginUserOperation by onesignalId.
+            val op = queueItem.operation
+            if (op is LoginUserOperation) {
+                val existing =
+                    queue.firstOrNull {
+                        it.operation is LoginUserOperation && it.operation.onesignalId == op.onesignalId
+                    }
```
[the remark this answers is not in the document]
nan-li: LoginUserOperation.onesignalId is non-null

nan-li:
@claude review

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -161,6 +163,41 @@ internal class OperationRepo(
...
+                        existingOp.existingOnesignalId = incomingExistingId
+                    } else {
+                        Logging.debug("OperationRepo: internalEnqueue - LoginUserOperation for onesignalId: ${op.onesignalId} already exists in the queue.")
+                    }
+                    // Transfer the waiter so enqueueAndWait callers see the queued op's real execution result.
+                    if (queueItem.waiter != null && existing.waiter == null) {
+                        existing.waiter = queueItem.waiter
+                    } else {
+                        queueItem.waiter?.wake(true)
+                    }
```
sherwinski, line 193: Might be worth writing a test for this (and L36) that confirms the waiters get switched.

sherwinski: Something like: seed the queue with an `OperationQueueItem(queuedOp, waiter = seededWaiter, bucket = 0)`, then call `enqueueAndWait` for the same
  `onesignalId`, let the executor run the queued op (mock it to return `SUCCESS`), and assert both the seeded `waiter` and the `enqueueAndWait` caller wake with true.

nan-li: added new test

abdulraqeeb33 (approved):

PR #2622, opened 2026-04-23
fix: retrieve    
  push token on startup when notification permission already granted

### LINKED ISSUE #2621: Subscription stuck at NO_PERMISSION after reinstall when notification permission already granted
## Environment
- react-native-onesignal: 5.4.2
- OneSignal Android SDK: 5.7.7
- Platform: Android

## Steps to Reproduce
1. Install app and grant notification permission
2. Reinstall the app (or clear app data)
3. Reopen — do NOT toggle the notification permission

## Expected Behavior
Notifications work normally after reinstall. Subscription status becomes `SUBSCRIBED`.

## Actual Behavior
OneSignal dashboard shows "Permission Not Granted". `getOptedInAsync()` returns `false`.
The state may eventually recover (hours later), or never recover at all.

## Root Cause

After reading the SDK source, the chain is:

**1. `NotificationsManager.permission` is initialized correctly:**
```kotlin
// NotificationsManager.kt
override var permission: Boolean = NotificationHelper.areNotificationsEnabled(appContext)
```
After reinstall, permission is already granted → `permission = true` from the start.

**2. Fresh subscription has `status = NO_PERMISSION`:**
```kotlin
// UserSwitcher.kt — createAndSwitchToNewUser()
status = currentPushSubscription?.status ?: SubscriptionStatus.NO_PERMISSION
```
After data clear, no previous subscription exists → `status = NO_PERMISSION`.

**3. `retrievePushTokenAndUpdateSubscription()` — the function that fixes `status` — has only two triggers:**
```kotlin
// DeviceRegistrationListener.kt
override fun onNotificationPermissionChange(permission: Boolean) {
    retrievePushTokenAndUpdateSubscription()  // trigger A
}
override fun onModelReplaced(model: ConfigModel, tag: String) {
    if (tag != ModelChangeTags.HYDRATE) return
    retrievePushTokenAndUpdateSubscription()  // trigger B
}
```

**4. Neither trigger fires:**
- **Trigger A** (`onNotificationPermissionChange`): never fires because `setPermissionStatusAndFire()` only fires when `oldPermission != newPermission`. Since `permission` starts as `true` (correctly read at init), and the system permission never changes, no event is emitted.
- **Trigger B** (`onModelReplaced(HYDRATE)`): only fires when the `ConfigModel` is re-fetched from the backend. If the config cache is still valid, no re-fetch occurs → never fires.

**Result:** `status` stays `NO_PERMISSION` indefinitely. The subscription is `optedIn = model.optedIn && status != NO_PERMISSION` = `false`.

## Suggested Fix

In `DeviceRegistrationListener.start()`, add a startup check:

```kotlin
override fun start() {
    _configModelStore.subscribe(this)
    _notificationsManager.addPermissionObserver(this)
    _subscriptionManager.subscribe(this)

    // Fix: if permission is already granted but subscription status is NO_PERMISSION,
    // trigger token retrieval immediately on startup.
    if (_notificationsManager.permission) {
        retrievePushTokenAndUpdateSubscription()
    }
}
```

This is safe — `retrievePushTokenAndUpdateSubscription()` is idempotent and already handles the case where a token exists.

tp26610:
# Description
## One Line Summary
Fix push subscription stuck at `NO_PERMISSION` after reinstall when notification permission is already granted.

### Motivation
After reinstall or app data clear, the push subscription is created fresh with `status = NO_PERMISSION`. If the system notification permission was already granted before reinstall, no permission-change event fires on startup — because `NotificationsManager.permission` is initialized to the correct value of `true`, so `oldPermission == newPermission` and `setPermissionStatusAndFire()` never emits an event.

The `onModelReplaced(HYDRATE)` path only fires when the config cache is invalid, so it cannot be relied on either.

Result: `retrievePushTokenAndUpdateSubscription()` is never called → subscription stays stuck at `NO_PERMISSION` indefinitely → user cannot receive push notifications even though system permission is granted.

Fixes #2621

### Scope
Only affects `DeviceRegistrationListener.start()`. The change adds one conditional call to `retrievePushTokenAndUpdateSubscription()` which is already called from two other places in the same class. No behaviour changes when permission is not granted at startup.

# Testing
## Unit testing
No unit tests added — the existing test suite for `DeviceRegistrationListener` covers `onNotificationPermissionChange` and `onModelReplaced`. A new test case for the `start()` path when `permission = true` would be the appropriate addition, but I don't have a local Android build environment to verify it compiles and passes CI.

## Manual testing
Reproduced with react-native-onesignal 5.4.2 (Android SDK 5.7.7) on a Pixel device (Android 14):
1. Install app → grant notification permission → confirm `getOptedInAsync()` returns `true`
2. Clear app data via Settings → reopen app (permission still granted)
3. Without fix: `getOptedInAsync()` returns `false` indefinitely
4. With fix: `getOptedInAsync()` returns `true` within a few seconds

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/listeners/DeviceRegistrationListener.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/listeners/DeviceRegistrationListener.kt
@@ -39,15 +39,30 @@ internal class DeviceRegistrationListener(
         _subscriptionManager.subscribe(this)
 
         // If notification permission is already granted at startup (e.g. after reinstall or
-        // app data clear), the push subscription status may still be NO_PERMISSION because
-        // no permission-change event will fire — permission was already true when the SDK
-        // initialized, so oldPermission == newPermission and no event is emitted.
-        // The config-hydration path (onModelReplaced/HYDRATE) only fires when the config
-        // cache is invalid, so it cannot be relied on either.
-        // Eagerly retrieve the push token here to ensure the subscription is active.
-        if (_notificationsManager.permission) {
+        // app data clear), the cached push subscription can be stuck at NO_PERMISSION:
+        //   - The permission observer (onNotificationPermissionChange) won't fire because
+        //     permission did not change — NotificationsManager.setPermissionStatusAndFire
+        //     only emits when oldPermissionStatus != isEnabled.
+        //   - The config-hydration path (onModelReplaced/HYDRATE) is triggered by
+        //     ConfigModelStoreListener.fetchParams on every startup, but it runs on IO and
+        //     depends on a successful params backend call, so it leaves a window of stale
+        //     state and may not run at all when the device is offline.
+        // Eagerly retrieve the push token here to close that window, but only when the
+        // cached push subscription doesn't already reflect a healthy SUBSCRIBED state —
+        // this avoids a redundant FCM round-trip on every warm start.
+        if (_notificationsManager.permission && needsPushTokenRefresh()) {
             retrievePushTokenAndUpdateSubscription()
         }
+    }
+
+    private fun needsPushTokenRefresh(): Boolean {
+        val pushModel = _subscriptionManager.pushSubscriptionModel
+        // An uninitialized push subscription has an empty model.id (see UninitializedPushSubscription);
+        // a real push subscription always has an id (local UUID or server-assigned).
+        if (pushModel.id.isEmpty()) {
+            return true
+        }
+        return pushModel.status != SubscriptionStatus.SUBSCRIBED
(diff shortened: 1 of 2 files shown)


a0979470582 (approved):

melody0000tw (approved):

abdulraqeeb33:
@claude review

nan-li (approved):

PR #2624, opened 2026-04-24
feat(jwt): add externalId and requiresJwt to Operation base class (2/6)

nan-li:
# Description

## One Line Summary
PR 2 of 6 against the Identity Verification integration branch — mechanical refactor that adds `externalId: String?` and `requiresJwt: Boolean` to the `Operation` base class so the next PR can implement queue-level IV gating. No runtime behavior change.

### Motivation
The next PR in the series (PR 3/6, "queue-level IV") needs to read `op.externalId` polymorphically from `OperationRepo`. Today `externalId` is declared on `LoginUserOperation` and `TrackCustomEventOperation` only — `OperationRepo` can't read it from a generic `Operation` reference without casting to each of the 16 subclass types. This PR hoists the property (and a companion `requiresJwt` flag) to the base class so downstream consumers can read them uniformly.

Kept deliberately narrow so the base-class refactor has its own reviewable surface; the behavior changes (pre-HYDRATE deferral, FAIL_UNAUTHORIZED handling, `IdentityVerificationService` with anon-op purge) come in the next PR.

### Scope
**`Operation` base class:**
- New `var externalId: String?` — captures the externalId of the user the op was enqueued for. Nullable because anonymous users have no externalId; base-class declaration keeps the type `String?` uniformly across all subclasses.
- New `open val requiresJwt: Boolean get() = true` — declares whether an op's backend endpoint needs a JWT when IV is active. Subclasses may override to `false` for endpoints that don't require auth (none do yet — default `true` covers every current op).

**Operation subclasses (16):**
- All 16 constructors gain `externalId: String?` as the 3rd parameter (after `appId` and `onesignalId`).
- `LoginUserOperation` and `TrackCustomEventOperation` shed their local `externalId` property in favor of the inherited base-class property (previously defined on each subclass independently).
- `TransferSubscriptionOperation` is the one exception in parameter position — its constructor is `(appId, subscriptionId, onesignalId, externalId)` to match the existing layout where `onesignalId` comes after `subscriptionId`.

**Call sites updated to pass `externalId` at construction time (10 files):**
- `IdentityModelStoreListener`, `PropertiesModelStoreListener`, `SubscriptionModelStoreListener` — read `_identityModelStore.model.externalId`.
- `SessionListener` (session start + end).
- `TrackGooglePurchase` (purchase tracking).
- `UserSwitcher` (legacy player ID migration).
- `RebuildUserService` (reconstructs queued ops from cached models).
- `UserRefreshService` (foreground user refresh).
- `LoginUserOperationExecutor`, `LoginUserFromSubscriptionOperationExecutor`, `SubscriptionOperationExecutor` (ops they emit as side effects, e.g. post-login `RefreshUserOperation`, post-subscription-recreate `CreateSubscriptionOperation`).

**`PropertiesModelStoreListener` gains `IdentityModelStore` as a constructor dependency** — the properties model doesn't carry externalId, so the listener has to read it from the identity model.

### Placement decision: base class vs per-subclass
`onesignalId` is declared on each subclass individually, not on the base. Considered matching that convention for `externalId`/`requiresJwt`, but chose base-class placement:
- `externalId` is naturally nullable; `String?` works uniformly across every subclass without weakening any type.
- Cross-cutting consumers (next PR's `hasValidJwtIfRequired`, `removeOperationsWithoutExternalId`) need to read `op.externalId` polymorphically — that requires at least an abstract declaration on the base. Per-subclass with abstract-on-base would mean ~80 lines of boilerplate (override + getter/setter × 16 subclasses) vs the 5-line base-class declaration.
- `requiresJwt` has natural override semantics (`open val` with default), which is idiomatic on a base class.

`name` on `Operation` is precedent for a non-nullable universal property on the base, so there's no technical constraint keeping `onesignalId` off the base — that placement is a historical style preference that this PR does not touch.

### What is NOT in this PR
- `IdentityVerificationService` — PR 3/6.
- `OperationRepo` IV gating (`hasValidJwtIfRequired` dispatch, FAIL_UNAUTHORIZED handler, pre-HYDRATE deferral in `getNextOps`) — PR 3/6.
- HTTP layer JWT attachment — PR 4/6.
- Public API (`login(externalId, jwt)`, `updateUserJwt`, listeners) — PR 5/6.
- Logout + IAM integration — PR 6/6.

# Testing

## Unit testing
- Compile verified — every call site updated to pass `externalId` (see compile errors list narrowed from 50+ to 0).
- All existing test fixtures updated to the new constructor signature. Most were mechanical substitutions (add `null` as 3rd arg for anon/un-scoped tests; add the relevant externalId where the test exercises it).
- 791 core tests run, 789 pass, 2 fail — both are pre-existing `SDKInitTests` failures on the integration branch (unrelated: one is about an error-message assertion, the other about init not blocking). Same two failures exist on `feat/iv-foundation-01` without this PR.

## Manual testing
Not applicable — this PR adds a property + parameter with no runtime behavior change. Every call site still passes identically-shaped data; the only difference is that `externalId` is now available for polymorphic reads in the next PR.

# Affected code checklist
   - [ ] Notifications
   - [ ] Outcomes
   - [x] Sessions (session listener passes externalId to track-session ops)
   - [ ] In-App Messaging
   - [x] REST API requests (only op shape; no new network traffic)
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing — adds two properties to `Operation` and updates every subclass + call site
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs (no public API changes)

## Testing
   - [x] I have included test coverage for these changes (existing tests cover via construction; no new behavior to test in this PR)
   - [x] All automated tests pass locally (pre-existing `SDKInitTests` failures are unrelated — see Manual testing)
   - [x] Manual testing not applicable — no runtime behavior change

## Final pass

### DIFF
--- a/OneSignalSDK/detekt/detekt-baseline-core.xml
+++ b/OneSignalSDK/detekt/detekt-baseline-core.xml
@@ -42,6 +42,7 @@
     <ID>ConstructorParameterNaming:InfluenceManager.kt$InfluenceManager$private val _configModelStore: ConfigModelStore</ID>
     <ID>ConstructorParameterNaming:InfluenceManager.kt$InfluenceManager$private val _sessionService: ISessionService</ID>
     <ID>ConstructorParameterNaming:InstallIdService.kt$InstallIdService$private val _prefs: IPreferencesService</ID>
+    <ID>ConstructorParameterNaming:JwtTokenStore.kt$JwtTokenStore$private val _prefs: IPreferencesService</ID>
     <ID>ConstructorParameterNaming:LanguageContext.kt$LanguageContext$private val _propertiesModelStore: PropertiesModelStore</ID>
     <ID>ConstructorParameterNaming:LoginUserFromSubscriptionOperationExecutor.kt$LoginUserFromSubscriptionOperationExecutor$private val _identityModelStore: IdentityModelStore</ID>
     <ID>ConstructorParameterNaming:LoginUserFromSubscriptionOperationExecutor.kt$LoginUserFromSubscriptionOperationExecutor$private val _propertiesModelStore: PropertiesModelStore</ID>
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/CoreModule.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/CoreModule.kt
@@ -45,6 +45,7 @@ import com.onesignal.location.ILocationManager
 import com.onesignal.location.internal.MisconfiguredLocationManager
 import com.onesignal.notifications.INotificationsManager
 import com.onesignal.notifications.internal.MisconfiguredNotificationsManager
+import com.onesignal.user.internal.jwt.JwtTokenStore
 
 internal class CoreModule : IModule {
     override fun register(builder: ServiceBuilder) {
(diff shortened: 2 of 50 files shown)


nan-li:
@claude review

PR #2625, opened 2026-04-24
feat(jwt): OperationRepo IV runtime — pre-HYDRATE deferral, FAIL_UNAUTHORIZED handler, IdentityVerificationService (3/6)

nan-li:
# Description

## One Line Summary
PR 3 of 6 against the Identity Verification integration branch — wires queue-level IV behavior (pre-HYDRATE deferral, FAIL_UNAUTHORIZED handler, anonymous-op purge) on top of PR 2's `Operation.externalId` base-class refactor.

### Motivation
PR 1 landed the gate state (`IdentityVerificationGates`) and JWT store; PR 2 gave every operation an `externalId`. This PR is where IV starts to actually do something at runtime. It lands the three behavioral pieces of `OperationRepo` + `IdentityVerificationService` that are necessary before the public API can be added:

1. **Pre-HYDRATE deferral** — the SDK can't dispatch ops until it knows whether `jwt_required` is true or false, otherwise a Phase 2 user's first-launch ops would race past the JWT gate and 401.
2. **`FAIL_UNAUTHORIZED` handling** — without this, a 401 drops the op and the developer has no way to recover; the SDK needs to notify the app so it can supply a fresh JWT, then retry.
3. **Anonymous-op purge** — when IV becomes active, ops queued under the previous anonymous user can't authenticate and would otherwise block the queue indefinitely.

### Scope

**`IOperationRepo` gains two methods** (both used only by internal SDK components — external callers don't see these change on the public surface):

- `fun setJwtInvalidatedHandler(handler: ((externalId: String) -> Unit)?)` — single-handler setter, set once by PR 5's `UserManager` with a lambda that fires the developer-facing listener `EventProducer`. Intentionally not an `EventProducer` here: there's exactly one internal subscriber.
- `fun removeOperationsWithoutExternalId()` — drops anon ops from both the in-memory queue and the persistent `OperationModelStore` so they don't resurrect on cold start.

**`OperationRepo`:**
- Injects `JwtTokenStore` directly. We considered a narrow `IJwtStore` interface defined in core (implemented by user-layer `JwtTokenStore`) to preserve core→user layering, but found `OperationRepo` already imports user-layer classes directly (e.g. `LoginUserOperation`, `NewRecordsState`), and 34 core files import from user overall. The core/user split is organizational, not architectural, so adding an abstraction for this one case would be inconsistent with the codebase.
- **Pre-HYDRATE deferral** at the top of `getNextOps`:
  ```kotlin
  if (_configModelStore.model.useIdentityVerification == JwtRequirement.UNKNOWN) return null
  ```
  Unconditional — applies even to Phase 1 non-IV users — because pre-HYDRATE we can't distinguish phases. First-launch delay is typically <2 s; `useIdentityVerification` is cached in `ConfigModel` after the first successful params fetch so subsequent launches skip this.
- **Outer `newCodePathsRun` dispatch** at the `getNextOps` filter and the `FAIL_UNAUTHORIZED` executor-result branch. Ops on the old code path (feature flag off, `jwt_required=false`) are unchanged. Split `FAIL_UNAUTHORIZED` out of the combined `FAIL_NORETRY`/`FAIL_CONFLICT` branch because its handling is now conditional.

**New `OperationRepoIvExtensions.kt`** (side-by-side with `OperationRepo.kt` per the PR plan's Option A file layout):
- `hasValidJwtIfRequired(jwtTokenStore, op)` — inner `ivBehaviorActive` guard short-circuits to `true` when IV behavior is off, so Phase 3 users (new code paths on, IV off) run through the dispatch site but no actual JWT gating happens. Returns `false` for anonymous ops when IV is active; otherwise checks `jwtTokenStore.getJwt(op.externalId) != null`.
- `handleFailUnauthorized(startingOp, ops, jwtTokenStore, handler)` — same inner gate. Returns `false` when IV inactive or op is anon so the caller falls through to the default drop-on-fail path. When handled: invalidates the JWT, **fires the handler before waking waiters** (otherwise an `enqueueAndWait` caller could return before the handler propagates to the app), wakes waiters with `false`, and re-queues ops without their waiters so they can retry once a fresh JWT arrives.

Both extensions take their dependencies as parameters rather than accessing `OperationRepo` private fields — keeps encapsulation intact and the extensions pure-ish.

**New `IdentityVerificationService`** (DI-registered `IStartableService`):
- Subscribes to `ConfigModelStore`; reacts only to HYDRATE.
- Schedules the purge via `suspendifyOnIO { operationRepo.awaitInitialized(); ... }` — **fix for an earlier race** where anon-op purge ran against an empty in-memory queue on cold start because it fired before `loadSavedOperations` completed.
- Purges anon ops only when the new `useIdentityVerification` is `REQUIRED`. Always calls `forceExecuteOperations` to wake the queue from the pre-HYDRATE deferral regardless of the new state.

**`CoreModule`** registers `IdentityVerificationService` as an `IStartableService`.

### Two-layer gating

The memory-plan design is preserved:
- **Outer gate** (dispatch site, `OperationRepo.kt`): `if (IdentityVerificationGates.newCodePathsRun) extensionFunction(...)`. When off, old code path runs untouched.
- **Inner gate** (extension body): `if (!IdentityVerificationGates.ivBehaviorActive) return <no-op>`. Lets Phase 3 users (new code paths on, IV off) run through the new dispatch without any actual IV gating applied — this is how we validate the new code paths structurally before flipping IV active for each customer.

### What is NOT in this PR
- Public API (`OneSignal.login(externalId, jwt)`, `updateUserJwt`, JWT-invalidated listeners) — **PR 4**.
- HTTP layer `Authorization: Bearer` header attachment — **PR 5**.
- Logout IV-aware behavior + IAM integration — **PR 6**.
- `JwtTokenStore.pruneToExternalIds()` invocation on cold start — deferred; not needed yet.

# Testing

## Unit testing
11 new tests, 789 existing tests still pass (2 pre-existing `SDKInitTests` failures on the integration branch are unrelated — same 2 that fail on `feat/iv-foundation-01` / `feat/iv-oprepo-gating-02`).

**New `IdentityVerificationServiceTests` (5):**
- `start subscribes to ConfigModelStore`.
- `HYDRATE with REQUIRED purges anon ops then wakes the queue (after awaitInitialized)` — uses `coVerifyOrder` to assert the purge runs *after* `awaitInitialized()`, which is the whole point of the race fix.
- `HYDRATE with NOT_REQUIRED wakes the queue but does NOT purge`.
- `HYDRATE with UNKNOWN wakes the queue (no purge)`.
- `non-HYDRATE model replacement is ignored`.

**New `OperationRepoTests` (6):**
- Pre-HYDRATE deferral blocks execution (queue retains op; executor not called).
- Flipping `useIdentityVerification` to `NOT_REQUIRED` and calling `forceExecuteOperations` unblocks the queue.
- `removeOperationsWithoutExternalId` filters the queue AND removes the persisted ops from the store.
- `setJwtInvalidatedHandler` round-trip via internal accessor.
- `FAIL_UNAUTHORIZED` with IV active: invalidates JWT, fires handler, re-queues ops (not dropped), wakes `enqueueAndWait` caller with `false`.
- `FAIL_UNAUTHORIZED` with IV inactive: handler NOT fired, default drop-on-fail path runs (op removed from store).

Each IV test scopes its `IdentityVerificationGates` mutations in `try/finally` to reset the singleton; `Mocks()` sets `configModel.useIdentityVerification = JwtRequirement.NOT_REQUIRED` by default so the pre-HYDRATE deferral doesn't hold up tests that aren't exercising IV state.

## Manual testing
Not applicable for this PR — no runtime behavior change for Phase 1 users (feature flag off, `jwt_required` eventually `false`) once HYDRATE completes; IV-specific behavior is gated off pending the rest of the rollout series. The pre-HYDRATE deferral is covered by unit tests.

# Affected code checklist
   - [ ] Notifications
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests (ops can now be held/re-queued/dropped based on IV state; no change to request shapes)
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing — wires queue-level IV behavior onto the refactor from PR 2
   - [x] Any Public API changes are explained in the PR details (none in this PR)

## Testing
   - [x] I have included test coverage for these changes
   - [x] All automated tests pass locally (pre-existing `SDKInitTests` failures are unrelated — same 2 on integration branch)
   - [x] Manual testing N/A (IV-specific behavior is gated off pending PR 4+ wiring)

## Final pass


nan-li:
@claude review

nan-li:
@claude review

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -415,12 +473,25 @@ internal class OperationRepo(
...
+        // Pre-HYDRATE deferral: wait until we know whether IV is required before dispatching
+        // any op, otherwise we could send an unsigned request when the customer has IV enabled.
+        // `isInitializedWithRemote` would be wrong here: pre-IV SDKs persisted it as `true`
+        // without ever reading `jwt_required`, so cached config from an upgrade looks
+        // "initialized" while `useIdentityVerification` is still UNKNOWN. Gating on UNKNOWN
+        // directly is correct because the IV-aware backend ships alongside this SDK — every
+        // successful HYDRATE will populate the field with REQUIRED or NOT_REQUIRED.
+        if (_configModelStore.model.useIdentityVerification == JwtRequirement.UNKNOWN) {
+            return null
+        }
```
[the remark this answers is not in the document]
nan-li: Accepted as a known limitation. A 403 from `/params` doesn't tell us whether IV is required — it tells us the app/account is rejected. Defaulting `useIdentityVerification = NOT_REQUIRED` on that path would silently disable JWT attachment for IV-required customers if a transient 403 ever occurred. The 403 branch is already documented as terminal ("omitting further retries") — the SDK is non-functional in that state regardless of IV. Won't fix.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -415,12 +452,26 @@ internal class OperationRepo(
...
+        // whether IV is required. Gated on `isInitializedWithRemote` (set unconditionally by
+        // `ConfigModelStoreListener` on any successful HYDRATE) rather than on
+        // `useIdentityVerification` being non-UNKNOWN, because the backend may legitimately
+        // omit `require_ident_auth` (older deployments, dev/test environments, partial
+        // rollouts) — if we gated on the param itself we would deadlock the queue for those
+        // users. Post-HYDRATE with a silent backend, `useIdentityVerification` stays UNKNOWN
+        // and `ivBehaviorActive` is false, so ops flow through the old code path normally.
+        if (!_configModelStore.model.isInitializedWithRemote) {
+            return null
+        }
```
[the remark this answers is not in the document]
nan-li: See response on https://github.com/OneSignal/OneSignal-Android-SDK/pull/2625#discussion_r3155524351 — accepted as known limitation.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/IdentityVerificationService.kt
```diff
@@ -0,0 +1,70 @@
...
+    /** Whether IV-specific behavior (JWT attachment, auth error handling) applies. UNKNOWN reads as `false`. */
+    val ivBehaviorActive: Boolean
+        get() = configModelStore.model.useIdentityVerification == JwtRequirement.REQUIRED
+
+    /** Whether new IV-related code paths should run. `featureFlag_IV_ON || jwt_required == REQUIRED`. */
+    val newCodePathsRun: Boolean
+        get() = featureManager.isEnabled(FeatureFlag.SDK_IDENTITY_VERIFICATION) || ivBehaviorActive
+
+    @Volatile
+    private var _onJwtConfigHydrated: ((ivRequired: Boolean) -> Unit)? = null
```
abdulraqeeb33: make a that val and try to use a sync block instead.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -244,6 +262,44 @@ internal class OperationRepo(
...
+     * to evict anon ops that can no longer execute.
+     */
+    internal fun removeOperationsWithoutExternalId() {
+        val removedIds: List<String> =
+            synchronized(queue) {
+                val anonymous = queue.filter { it.operation.externalId == null }
+                anonymous.forEach { it.waiter?.wake(false) }
+                queue.removeAll(anonymous)
+                Logging.debug("OperationRepo: removeOperationsWithoutExternalId removed ${anonymous.size} of ${anonymous.size + queue.size} operations")
+                anonymous.map { it.operation.id }
```
abdulraqeeb33, line 276: why is this needed?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -305,7 +361,24 @@ internal class OperationRepo(
...
+                            handleFailUnauthorized(
+                                startingOp,
+                                ops,
+                                _jwtTokenStore,
+                                _jwtInvalidatedHandler,
+                                _identityVerificationService.ivBehaviorActive,
+                            )
+                    if (!handled) {
+                        // IV inactive or anon op: drop and wake waiters, matching FAIL_NORETRY.
+                        Logging.warn("Operation execution failed without retry: $operations")
```
abdulraqeeb33, line 371: can we create a resuable method/extension method instead of duplicating code?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationRepo.kt
```diff
@@ -42,6 +42,14 @@ interface IOperationRepo {
...
 
     fun forceExecuteOperations()
+
+    /**
+     * Registers a handler invoked with the externalId of the user whose op just hit a
+     * FAIL_UNAUTHORIZED response. Set once by the public-API layer so it can forward to
+     * developer-facing [com.onesignal.user.internal.jwt.IUserJwtInvalidatedListener]s
+     * (single handler — not multi-subscriber).
+     */
+    fun setJwtInvalidatedHandler(handler: ((externalId: String) -> Unit)?)
```
abdulraqeeb33: lets check if there is a better way to do this.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: Ah it was due to being a block expression, the last expression of `synchronized(queue) { ... }` is its return value, assigned to `removedIds`. We capture IDs inside the lock, remove from the store outside to avoid holding the queue lock across N store calls.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/IOperationRepo.kt
nan-li: removed the `JwtInvalidatedHandler` and added invalidated callback to the existing `IJwtUpdateListener` instead in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2625/changes/aec7562ff8411cd5e600e0b24950ad92dbc69387

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
nan-li: done in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2625/commits/cce340a646d4f4b70e2d9512f80de3458ed89ef6

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/config/impl/IdentityVerificationService.kt
nan-li: Done in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2625/commits/cce340a646d4f4b70e2d9512f80de3458ed89ef6, note there's only 1 listener so instead of handling multiple, keep this way for YAGNI

nan-li:
@claude review

PR #2636, opened 2026-05-06
fix: [SDK-4474] self-heal SDK-4388-stuck push subscriptions on session start

sherwinski:
# Description
## One Line Summary
On session start, detect when the server's record of the device's push subscription is disabled while the local model says enabled-and-opted-in, and re-assert local truth via `PATCH /subscriptions/{id}` so users already stuck by [SDK-4388](https://linear.app/onesignal/issue/SDK-4388/android-sdk-subscription-state-permanently-stuck-at-never-subscribed) self-heal on next launch.

### Motivation
The fix in #2627 (parent PR) prevents *new* installs from getting stuck by flipping `POST /subscriptions` to `PATCH /subscriptions/{id}` in `SubscriptionOperationExecutor` when the subscription id is server-assigned. That covers the forward path, but it does **not** rescue the cohort that already triggered the bug on an older SDK build:

- The buggy `POST /subscriptions` returned `200 {}` (server-side no-op for an existing id).
- The merged `create-subscription + update-subscription(SUBSCRIBED)` group was therefore acked and dropped from the local op queue.
- On in-place upgrade to the fixed SDK there is nothing left in the queue to retry, and the server view stays at `enabled:false, notification_types:0` permanently.

This is confirmed with the SDK-4388 demo-app reproduction:
- force-stop on the buggy build (server stuck)
- upgrade to the fixed build
- relaunch — no `PATCH /subscriptions/{id}` in the logs.

### Scope
- **Affected:** `RefreshUserOperationExecutor.getUser`. On every session start the SDK already issues `GET /users/by/onesignal_id/{osid}`, whose response carries `enabled` / `notification_types` per subscription. We now inspect the entry whose `id` matches the device's cached push subscription id, and if `local.enabled && !server.enabled`, we emit a follow-up `UpdateSubscriptionOperation` returned via `ExecutionResponse.operations`. That op runs through the standard executor path and produces one `PATCH /subscriptions/{id}` carrying the device's true state.
- **Unchanged:** \"Device is the source of truth for push\" policy. We still do **not** hydrate the server's push subscription into `SubscriptionModelStore` (the existing `if (subscriptionModel.type != SubscriptionType.PUSH)` filter is preserved). The new code only enforces that policy across the wire when divergence is detected.
- **No-op for healthy users.** Server-matches-local triggers no PATCH.
- **Respects user intent.** A locally-opted-out subscription (`optedIn=false`, status `UNSUBSCRIBE`) or one with `NO_PERMISSION` does not trigger a reconcile — those server-side disables are correct, not stuck.

### Cost
One extra `PATCH /subscriptions/{id}` per stuck device on its first foreground session under the new SDK build. Healthy devices pay nothing. Once the PATCH succeeds, server matches local and no further reconciles fire.

# Testing
## Unit testing
Added 4 cases to `RefreshUserOperationExecutorTests`:

1. **divergence** — server `enabled:false, notification_types:0` + local opted-in + SUBSCRIBED + non-empty token → returns exactly one `UpdateSubscriptionOperation` with `enabled=true, status=SUBSCRIBED`, correct ids.
2. **healthy** — server `enabled:true, notification_types:1` + local opted-in → no follow-up op.
3. **opted-out** — server `enabled:false, notification_types:-2` + local `optedIn=false` → no follow-up op (UNSUBSCRIBE is intentional).
4. **NO_PERMISSION** — server `enabled:false, notification_types:0` + local `status=NO_PERMISSION` → no follow-up op (OS denial is real).

The existing happy-path test `refresh user is successful and models are hydrated properly` was updated to set `notificationTypes = 1` on the cached push subscription so it stays on the new healthy path.

Test gate: `./gradlew :OneSignal:core:testReleaseUnitTest --tests \"...RefreshUserOperationExecutorTests\"` → 10/10 PASS. Two unrelated `SDKInitTests` failures exist on the parent branch and are not introduced by this PR.

Detekt + production assemble (`:OneSignal:core:detekt`, `:OneSignal:core:assembleRelease`) clean.

## Manual testing
End-to-end repro and validation are intended to run on the demo app shipped with the parent SDK-4388 branch (which has the \"Reproduce SDK-4388\" + \"Self-heal test\" sections):

1. Force-stop demo on the **broken** SDK build with cached stuck state (`enabled:false, notification_types:0` on dashboard).
2. Upgrade in place to this branch, relaunch.
3. Expect on first session: `GET /users/by/onesignal_id/{osid}` followed by a single `PATCH /subscriptions/{id}` with `enabled:true, notification_types:1`.
4. Dashboard subscription flips from \"Never Subscribed\" to \"Subscribed\".
5. On subsequent sessions: only `GET` (no further PATCH) — divergence check is a no-op once reconciled.

# Affected code checklist
- [ ] Notifications
   - [ ] Display
   - [ ] Open
   - [ ] Push Processing
   - [ ] Confirm Deliveries
- [ ] Outcomes
- [ ] Sessions
- [ ] In-App Messaging
- [x] REST API requests — adds at most one `PATCH /subscriptions/{id}` per stuck device on first session post-upgrade
- [ ] Public API changes

# Checklist
## Overview
- [x] PR does one thing — adds a single divergence detection branch in `RefreshUserOperationExecutor.getUser`
- [x] No public API changes

## Testing
- [x] Unit test coverage added for all four boundary cases
- [x] All automated tests pass (pre-existing `SDKInitTests` failures are unrelated to this change)
- [ ] Personally tested on device — pending end-to-end verification with the demo-app repro flow described above

## Final pass
- [x] Code is as readable as possible — divergence logic isolated in a single helper `buildPushSelfHealOperationOrNull` with explanatory comments referencing SDK-4388 / SDK-4474
- [x] Self-reviewed

---

**Branching:** This PR targets the SDK-4388 fix branch (#2627 parent) so the two changes can land together as a complete \"prevent new + heal old\" pair. If the team prefers to ship the SDK-4388 fix first and follow up with this independently, retarget to `main`.

Refs: [SDK-4474](https://linear.app/onesignal/issue/SDK-4474), [SDK-4388](https://linear.app/onesignal/issue/SDK-4388)

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
@@ -224,7 +224,7 @@ internal class RefreshUserOperationExecutor(
             return null
         }
 
-        Logging.warn(
+        Logging.info(
             "RefreshUserOperationExecutor: push subscription $pushSubscriptionId diverged from server " +
                 "(server enabled=${serverSubscription.enabled} notificationTypes=${serverSubscription.notificationTypes}; " +
                 "local opted-in and SUBSCRIBED). Enqueuing follow-up update-subscription op to re-assert local " +
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
@@ -107,7 +107,7 @@ internal class RefreshUserOperationExecutor(
             val pushSubscriptionIdFromConfig = _configModelStore.model.pushSubscriptionId
 
             val subscriptionModels = mutableListOf<SubscriptionModel>()
-            var pushSelfHealOperation: UpdateSubscriptionOperation? = null
+            var pushSelfHealOperationForStuckSubscription: UpdateSubscriptionOperation? = null
             for (subscription in response.subscriptions) {
                 val subscriptionModel = SubscriptionModel()
                 subscriptionModel.id = subscription.id!!
@@ -135,7 +135,7 @@ internal class RefreshUserOperationExecutor(
                 // so we don't want to cache these subscriptions from the backend.
                 if (subscriptionModel.type != SubscriptionType.PUSH) {
                     subscriptionModels.add(subscriptionModel)
-                } else if (subscription.id == pushSubscriptionIdFromConfig && pushSelfHealOperation == null) {
+                } else if (subscription.id == pushSubscriptionIdFromConfig && pushSelfHealOperationForStuckSubscription == null) {
                     // Self-heal for users stuck at "Never Subscribed". Older SDK builds dispatched
                     // the merged create-subscription + update-subscription(SUBSCRIBED) batch as a
                     // POST /subscriptions carrying the already-existing server-side id; the server
@@ -147,7 +147,7 @@ internal class RefreshUserOperationExecutor(
                     // re-assert local truth via PATCH. "Device is the source of truth" is the
                     // existing policy for push; this just enforces it across the wire when the
                     // server has drifted out of sync.
-                    pushSelfHealOperation = buildPushSelfHealOperationForStuckSubscription(op, subscription, pushSubscriptionIdFromConfig)
+                    pushSelfHealOperationForStuckSubscription = buildPushSelfHealOperationForStuckSubscription(op, subscription, pushSubscriptionIdFromConfig)
                 }
             }
 
@@ -167,7 +167,7 @@ internal class RefreshUserOperationExecutor(
 
             return ExecutionResponse(
                 ExecutionResult.SUCCESS,
-                operations = pushSelfHealOperation?.let { listOf<Operation>(it) },
+                operations = pushSelfHealOperationForStuckSubscription?.let { listOf<Operation>(it) },
             )
         } catch (ex: BackendException) {
             val responseType = NetworkUtils.getResponseStatusType(ex.statusCode)
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
@@ -207,7 +207,6 @@ internal class RefreshUserOperationExecutor(
      * Caller has already verified that [serverSubscription.id] matches the cached
      * `pushSubscriptionId`, so this method only inspects the local model + server flags.
      */
-    @Suppress("ReturnCount")
     private fun buildPushSelfHealOperationForStuckSubscription(
         op: RefreshUserOperation,
         serverSubscription: SubscriptionObject,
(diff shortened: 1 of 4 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -93,7 +96,13 @@ internal class RefreshUserOperationExecutor(
             // No longer hydrate timezone from remote, set locally
             propertiesModel.timezone = TimeUtils.getTimeZoneId()
 
+            // Retrieve the push subscription ID from the backend configuration model store. Used
+            // both for re-attaching the locally-cached push model below and for the SDK-4474
+            // self-heal divergence check inside the loop.
+            val pushSubscriptionIdFromConfig = _configModelStore.model.pushSubscriptionId
+
             val subscriptionModels = mutableListOf<SubscriptionModel>()
+            var pushSelfHealOperation: UpdateSubscriptionOperation? = null
```
sherwinski: Open to different naming if someone has a strong preference here

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -170,6 +194,49 @@ internal class RefreshUserOperationExecutor(
...
+    /**
+     * SDK-4474 self-heal builder. Returns an [UpdateSubscriptionOperation] iff the cached
+     * push subscription model resolves to enabled-and-opted-in but the server response
+     * recorded the same id as disabled. Returns null otherwise (no divergence — no-op).
+     *
+     * Caller has already verified that [serverSubscription.id] matches the cached
+     * `pushSubscriptionId`, so this method only inspects the local model + server flags.
+     */
+    @Suppress("ReturnCount")
+    private fun buildPushSelfHealOperationOrNull(
```
abdulraqeeb33:
```suggestion
    private fun buildPushSelfHealOperationForStuckSubscription
```

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -170,6 +194,49 @@ internal class RefreshUserOperationExecutor(
...
+            return null
+        }
+
+        val (localEnabled, localStatus) = SubscriptionModelStoreListener.getSubscriptionEnabledAndStatus(cachedPushSubscriptionModel)
+        val serverEnabled = (serverSubscription.enabled == true) && ((serverSubscription.notificationTypes ?: 0) > 0)
+        if (!localEnabled || serverEnabled) {
+            return null
+        }
+
+        Logging.warn(
```
abdulraqeeb33: can we reduce it to an info?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
```diff
@@ -170,6 +194,49 @@ internal class RefreshUserOperationExecutor(
...
 
+    /**
+     * SDK-4474 self-heal builder. Returns an [UpdateSubscriptionOperation] iff the cached
+     * push subscription model resolves to enabled-and-opted-in but the server response
+     * recorded the same id as disabled. Returns null otherwise (no divergence — no-op).
+     *
+     * Caller has already verified that [serverSubscription.id] matches the cached
+     * `pushSubscriptionId`, so this method only inspects the local model + server flags.
+     */
+    @Suppress("ReturnCount")
```
abdulraqeeb33: can we find a way to avoid using this Suppression?

abdulraqeeb33 (commented):
a few nits that i think we should change

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
sherwinski:
```suggestion
        Logging.info(
```

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
sherwinski: Done in 4dffa22a1 — collapsed the divergence branch into a trailing `return if (divergent) { ... } else null` so the function body now uses exactly two `return` keywords (early null/type guard + final expression). `@Suppress("ReturnCount")` removed; detekt clean on a forced rerun (`./gradlew :OneSignal:core:detekt --rerun-tasks`).

While I was at it I also caught two unrelated bugs introduced when 803aa366b (the rename suggestion) was applied: the trailing `(` on the function declaration was dropped, and the call site in `getUser` was never updated. The branch wouldn't compile in that state. Replayed cleanly as 22010102e (rename, co-authored with you preserved) and re-cherry-picked the warn->info change as 1d7ef5bbe. CI should compile now.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
abdulraqeeb33: do you want to rename it to the similar thing as the method name? `buildPushSelfHealOperationForStuckSubscription`

abdulraqeeb33 (approved):
LGTM with just one nit

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/RefreshUserOperationExecutor.kt
sherwinski: addressed in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2636/changes/cd1c0a7e041beb47785a2e10bec8bb73bb5ea262

sherwinski:
@claude[agent] review

PR #2638, opened 2026-05-06
fix: defensive JSON parsing for params + IAM (SDK-4478, SDK-4494)

abdulraqeeb33:
Two related fixes that stop the SDK from generating noisy / dangerous `JSONException`s when an upstream payload is the wrong shape.

Fixes:
- [SDK-4478](https://linear.app/onesignal/issue/SDK-4478/paramsbackendservicefetchparams-silently-kills-params-refresh-loop-on) — `ParamsBackendService.fetchParams` silently kills the params refresh loop on a non-JSON 2xx response.
- [SDK-4494](https://linear.app/onesignal/issue/SDK-4494/webviewmanagerpagerecttoviewheight-emits-noisy-error-level) — `WebViewManager.pageRectToViewHeight` emits an ERROR-level `JSONException` (with full stack trace) on missing `rect`, flooding OTel/Datadog with non-actionable alerts.

Common theme: stop the SDK from throwing `JSONException` for benign/recoverable conditions, and route any remaining surprise through `Logging.warn` with a bounded body snippet so issues are still observable without being alert-worthy.

## SDK-4478 — Params refresh loop

`ParamsBackendService.fetchParams` calls `JSONObject(response.payload!!)` unconditionally on any 2xx response. When the body isn't valid JSON (intercepting proxy like Burp, captive portal, CDN error HTML, charset/gzip corruption, etc.) the resulting `JSONException` escapes the caller's retry loop because `ConfigModelStoreListener.fetchParams` only catches `BackendException`.

This is **not** an app-level crash today: `suspendifyOnIO` wraps the block in a top-level `try { ... } catch (Exception) { Logging.error(...) }`, so the exception is caught at the dispatcher boundary and printed via `Log.e`. The actual symptom is a **silent functional regression** — the `do { ... } while (!success)` retry/backoff loop is unwound entirely, so for the rest of that session `android_params.js` is never re-fetched and the SDK keeps running on cached/default params (FCM params, notification channels, outcomes config, log level, etc.).

> Note: in older revisions of `ThreadUtils.suspendifyOnIO` that didn't have a top-level `catch (Exception)`, this same `JSONException` could have escaped to `Thread.UncaughtExceptionHandler` and actually crashed the process. On current `main` it's silent. The fix is correct either way.

Reported logcat trace:

```
org.json.JSONException: Value <html><head><title>Burp of type java.lang.String cannot be converted to JSONObject
    at org.json.JSON.typeMismatch(JSON.java:112)
    at org.json.JSONObject.<init>(JSONObject.java:163)
    at org.json.JSONObject.<init>(JSONObject.java:176)
    at com.onesignal.core.internal.backend.impl.ParamsBackendService.fetchParams(ParamsBackendService.kt:42)
    ...
```

**Fix:** wrap the `JSONObject(...)` parse in a try/catch and rethrow as `BackendException(statusCode, payload, retryAfterSeconds)`. This routes a malformed body into the existing retry/backoff path in `ConfigModelStoreListener` exactly like any other transient backend failure. Original status code is preserved so the existing 403 short-circuit still works. A `null` payload (which would have NPE'd on `payload!!`) is also handled the same way.

**Additional companion test in `FeatureFlagsBackendServiceTests`** asserts that the Turbine SDK feature-flags endpoint already handles a fully non-JSON 200 body safely (its `FeatureFlagsJsonParser.parseSuccessful` already catches `Throwable`). No production change there — just locking the contract so a future refactor can't drop the catch silently.

## SDK-4494 — IAM page-rect noisy ERROR log

Production OTel logs (sample insertId `4133ooyq0ths9s5j`) show this at ERROR level with a full stack trace:

```
org.json.JSONException: No value for rect
    at org.json.JSONObject.get(JSONObject.java:398)
    at org.json.JSONObject.getJSONObject(JSONObject.java:618)
    at com.onesignal.inAppMessages.internal.display.impl.W.pageRectToViewHeight(SourceFile:7)
    at com.onesignal.inAppMessages.internal.display.impl.W.calculateHeightAndShowWebViewAfterNewActivity$lambda$0(SourceFile:13)
    ...
log.message: "[main] pageRectToViewHeight could not get page height"
severity: ERROR
```

Like SDK-4478 this is **not** a crash — `pageRectToViewHeight` already catches `JSONException`, logs `Logging.error(...)` and returns `-1`. The IAM keeps working. But the `WebView.evaluateJavascript("getPageMetaData()")` callback can deliver a payload without `rect` for benign, recoverable reasons:

- The IAM HTML's JS hasn't finished defining `getPageMetaData()` yet (race with activity rotation / new activity).
- The page returned partial metadata (custom IAM template).
- The function returned `undefined` (callback delivers the literal string `"null"`, which `JSONObject("null")` rejects).

Logging this at ERROR floods OTel/Datadog and triggers operator alerts for a non-actionable race.

**Fix:**

- Switch `pageRectToViewHeight` to `optJSONObject("rect")` / `optInt("height", -1)` so missing fields stop generating thrown exceptions; demote the log to a single WARN with a 200-char body snippet (same pattern as `FeatureFlagsBackendService.bodySnippet`).
- In `calculateHeightAndShowWebViewAfterNewActivity`, bail out early when `evaluateJavascript` returns `null` or the literal `"null"`, instead of letting `JSONObject(...)` throw.
- Replace the bare `e.printStackTrace()` (which bypassed the SDK logging pipeline) with `Logging.warn(..., e)` + body snippet so any remaining surprise is still observable.

IAM display behavior on the happy path is unchanged.

## Test plan

- [x] `ParamsBackendServiceTests` (4 new tests): HTML 200, null 200, non-2xx passthrough, happy path. All pass.
- [x] `FeatureFlagsBackendServiceTests` (15 tests including new HTML-on-200 contract test). All pass.
- [x] `:OneSignal:core:detekt` clean for touched files.
- [x] `:OneSignal:in-app-messages:assembleRelease` clean (warnings only on pre-existing unrelated lines).
- [x] `:OneSignal:in-app-messages:detekt` clean.
- [x] `:OneSignal:in-app-messages:testReleaseUnitTest` — full module suite passes.
- [x] Full `:OneSignal:core:testReleaseUnitTest` — only 2 pre-existing `SDKInitTests` failures, unrelated and reproducible on bare `main` (verified by stashing this fix).

## Commits

1. `fix: SDK-4478: handle non-JSON params response without killing refresh loop` (`ParamsBackendService` + tests)
2. `test: pin FeatureFlags non-JSON 200 contract (companion to SDK-4478)` (no production change, regression test only)
3. `fix: SDK-4494: stop logging IAM page-rect JSONException at ERROR level` (`WebViewManager`)

### DIFF
--- a/OneSignalSDK/detekt/detekt-baseline-core.xml
+++ b/OneSignalSDK/detekt/detekt-baseline-core.xml
@@ -174,6 +174,7 @@
     <ID>InstanceOfCheckForException:HttpClient.kt$HttpClient$t is UnknownHostException</ID>
     <ID>LongMethod:ApplicationService.kt$ApplicationService$override suspend fun waitUntilSystemConditionsAvailable(): Boolean</ID>
     <ID>LongMethod:ConfigModelStoreListener.kt$ConfigModelStoreListener$private fun fetchParams()</ID>
+    <ID>LongMethod:FeatureFlagsBackendService.kt$FeatureFlagsBackendService$override suspend fun fetchRemoteFeatureFlags(appId: String): RemoteFeatureFlagsFetchOutcome</ID>
     <ID>LongMethod:HttpClient.kt$HttpClient$private suspend fun makeRequestIODispatcher( url: String, method: String?, jsonBody: JSONObject?, timeout: Int, headers: OptionalHeaders?, ): HttpResponse</ID>
     <ID>LongMethod:IdentityOperationExecutor.kt$IdentityOperationExecutor$override suspend fun execute(operations: List&lt;Operation>): ExecutionResponse</ID>
     <ID>LongMethod:LoginUserOperationExecutor.kt$LoginUserOperationExecutor$private suspend fun createUser( createUserOperation: LoginUserOperation, operations: List&lt;Operation>, ): ExecutionResponse</ID>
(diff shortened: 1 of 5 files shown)


fadi-george (approved):

sherwinski (approved):
LGTM but seems like code cov is below threshold

nan-li (approved):

PR #2645, opened 2026-05-12
fix: warm OneSignalDispatchers on init to avoid cold-start ANRs

abdulraqeeb33:
## Summary

Move the `OneSignalDispatchers` lazy-chain construction cost off the main thread on cold start. Closes the structural root cause that `#2644` covers tactically with FF-gated wrappers — a future call site that slips past the FF gate (or a `JobService` delivered to main before init has run) would otherwise still pay the construction cost on the calling thread.

Linear: SDK-4512
Base branch: `ar/sdk-4506-notifications-onfocus-anr` (#2644)

Note: the branch name `ar/sdk-4507` is numerically sequential with the rest of this series but is unrelated to Linear `SDK-4507` (which is an unrelated Web ticket). The tracking ticket for this PR is `SDK-4512`.

## Motivation

ANR-dump analysis of `logs/2026-05-12 (500 ANR entries, all on sdk_background_threading)` shows **23 / 500 (4.6 %)** of ANRs ending in `SyncJobService.onStartJob -> suspendifyOnIO`, all bottoming out in the same `OneSignalDispatchers` lazy chain:

```
ThreadPoolExecutor.execute -> LinkedBlockingQueue.offer
CoroutineDispatcher.dispatch -> kotlinx.coroutines first-launch
OneSignalDispatchers.IOScope.<init>   (by lazy)
OneSignalDispatchers.IO               (by lazy)
OneSignalDispatchers.ioExecutor       (by lazy)
```

The first IO consumer in the process pays the executor + dispatcher + scope construction + the kotlinx.coroutines `MainDispatcherFactory` `ServiceLoader` scan on its thread. Under `sdk_background_threading` whichever main-thread caller wins the race eats **5–20 s** before the watchdog fires.

#2644 routes the five known onFocus / onUnfocused handlers through `runOnSerialIOIfBackgroundThreading` so they no longer fire on main, but the deeper structural problem is the lazy chain itself — a future call site that slips past the FF gate (or a `JobService` delivered to main before init has run) hits the same stall.

## Fix: `OneSignalDispatchers.prewarm()`

* Spawns a dedicated short-lived `OneSignal-prewarm` daemon thread that submits one empty launch on each of `IO` / `Default` / `SerialIO`.
* That single thread pays the lazy-init cost end-to-end (executor → dispatcher → scope → `MainDispatcherFactory` `ServiceLoader` scan → worker thread spin-up) so the next production caller — even on the main thread — only sees the cheap submit-to-already-constructed-executor path.
* Double-checked-locked `prewarmStarted` flag — repeat calls from init / suspend init / `SyncJobService.onStartJob` no-op cheaply. An internal `resetPrewarmForTest()` lets specs exercise the "first call wins" branch independently.
* Daemon at `NORM_PRIORITY - 2` so prewarm never blocks process exit or starves UI work.
* Failures log and swallow. The existing `Dispatchers.IO` / `SerialIO` fallback paths still apply if anything goes wrong, so a failed prewarm just means the first real caller pays the original cost.

Called from:
* `OneSignalImp.initWithContext(context, appId)` (sync variant)
* `OneSignalImp.initWithContextSuspend(context, appId)` (suspend variant, used by re-entrant suspend callers)
* `SyncJobService.onStartJob` **before** `suspendifyOnIO`, because the `JobService` can fire before the host app's `initWithContext` runs and would otherwise be the first IO consumer.

## Base branch

Stacked on `ar/sdk-4506-notifications-onfocus-anr` (#2644). Together with #2643 and #2644 this covers the full **~94 / 500** main-thread-ANR bucket from `logs/2026-05-12` attributable to SDK threading helpers (47 onFocus + 23 JobService + 25 SessionService).

### Static
* `:OneSignal:core:detekt` — clean.

### Automated
* `:OneSignal:core:testReleaseUnitTest` full suite — green, including the new `OneSignalDispatchersTests` prewarm cases:
  * `prewarm` returns immediately on the caller and the daemon thread brings `IO` / `Default` / `SerialIO` and their scopes to `Active`.
  * `prewarm` is idempotent — second call does not spawn another `OneSignal-prewarm` thread (verified via thread-name scan).

### Manual

Cold-start ANR rate on the `sdk_background_threading=ON` cohort should drop ~9 % from the prewarm fix alone (the 47 onFocus entries also benefit indirectly because the FF-on path no longer races for first-IO-consumer status).

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
@@ -235,7 +235,6 @@ object OneSignalDispatchers {
             prewarmStarted = false
         }
     }
-
     internal fun getPerformanceMetrics(): String {
         return try {
             val serialQueueSize =
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
@@ -9,6 +9,8 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.ThreadFactory
 import java.util.concurrent.ThreadPoolExecutor
@@ -43,6 +45,8 @@ object OneSignalDispatchers {
         "$BASE_THREAD_NAME-IO" // Thread name prefix for I/O operations
     private const val DEFAULT_THREAD_NAME_PREFIX =
         "$BASE_THREAD_NAME-Default" // Thread name prefix for CPU operations
+    private const val SERIAL_IO_THREAD_NAME =
+        "$BASE_THREAD_NAME-SerialIO" // Single, named thread for order-sensitive work
 
     private class OptimizedThreadFactory(
         private val namePrefix: String,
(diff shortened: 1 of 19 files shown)


abdulraqeeb33:
@copilot resolve the merge conflicts in this pull request

fadi-george (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -172,6 +172,69 @@ object OneSignalDispatchers {
...
+     *
+     * Idempotent and fire-and-forget: a no-op on second and subsequent calls. Safe to invoke
+     * from any thread; the heavy lifting always happens on the daemon thread we spawn here, not
+     * on the caller. Failures are logged and swallowed because the executors retain their
+     * existing fallback paths (e.g. `Dispatchers.IO.limitedParallelism(1)` for [SerialIO]) and a
+     * failed prewarm will simply mean the first production caller pays the original cost.
+     */
+    fun prewarm() {
+        if (prewarmStarted) return
+        synchronized(prewarmLock) {
```
nan-li, line 205: For my understanding, what is the purpose of having both `volatile` flag and synchronization when writing to `prewarmStarted`?

nan-li (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
nan-li: figurd it out

PR #2646, opened 2026-05-12
fix: SDK-4508 offload SessionService.onFocus / onUnfocused off the main thread

abdulraqeeb33:
## Summary

ANR-dump analysis of `logs/2026-05-12 (500 ANR entries, all on sdk_background_threading)` attributes **25 of 500 ANRs (5.0%)** to `SessionService.onFocus` blocking the main thread. Stack signature:

```
java.util.concurrent.LinkedBlockingQueue.offer
java.util.concurrent.ThreadPoolExecutor.execute
kotlinx.coroutines.scheduling.CoroutineScheduler.dispatch
com.onesignal.common.threading.OneSignalDispatchers.<lazy chain>
... operation-repo / IAM trigger eval handlers ...
com.onesignal.session.internal.session.impl.b$c.invoke   (sessionLifeCycleNotifier.fire)
com.onesignal.session.internal.session.impl.b.onFocus    (SessionService.onFocus)
com.onesignal.core.internal.application.impl.a.handleFocus  (fanout on main)
```

`SessionService.onFocus` runs synchronously on the main thread via `ApplicationService.handleFocus -> applicationLifecycleNotifier.fire`. Inside, `sessionLifeCycleNotifier.fire { onSessionStarted / onSessionActive }` invokes the registered session-lifecycle handlers (operation repo, IAM trigger eval, etc.) synchronously; the first one to touch `OneSignalDispatchers` pays the cold-init cost on the main thread (5–20s blocks before the watchdog fires).

Same SDK-4505 / SDK-4506 / SDK-4507 fanout pattern — but on a pre-existing handler that predates background threading.

Linear: SDK-4508

## Fix

Wrap `onFocus` / `onUnfocused` bodies in `runOnSerialIOIfBackgroundThreading` (the same gated helper SDK-4505, SDK-4506, and SDK-4507 use). FF off retains the original inline semantics for the rollout control cohort and existing synchronous tests; FF on routes through the dedicated serial IO dispatcher so the fanout no longer runs on main.

### Timing semantics preserved

The body of `onFocus` writes `session.startTime = _time.currentTimeMillis` and `session.focusTime = _time.currentTimeMillis` to record when Android delivered the lifecycle event. If we simply deferred everything to the serial dispatcher, those timestamps would shift to whenever the dispatcher got around to the block — which can be hundreds of ms later under load.

To keep the existing semantics:

- `onFocus` captures `_time.currentTimeMillis` on the caller's thread BEFORE the wrapper, passes it into the deferred `handleOnFocus(firedOnSubscribe, focusTimeMs)` block.
- `onUnfocused` does the same so `activeDuration` accounting stays accurate (`dt = unfocusTimeMs - session.focusTime`, both captured on the main thread).

## Base branch

Stacked on `ar/sdk-4506-notifications-onfocus-anr` (#2644) because the fix relies on `runOnSerialIOIfBackgroundThreading`, which was introduced in that PR. Rebases cleanly onto `main` after #2643 and #2644 land.

## Test plan

- `SessionServiceTests` existing assertions about state mutation continue to pass (the FF defaults to off, so the wrapper runs inline; the new private `handleOnFocus` / `handleOnUnfocused` preserve behavior).
- Adds three new regression tests:
  - `onFocus` dispatches the session-mutation body through `runOnSerialIOIfBackgroundThreading`.
  - `onUnfocused` dispatches the `activeDuration` update through `runOnSerialIOIfBackgroundThreading`.
  - Rapid `onUnfocused -> onFocus` burst routes both events through the same gated helper in submission order (mirrors SDK-4505 `BackgroundManager` burst coverage; protects against `activeDuration` drift if the events ever raced across an unbounded IO pool).
- `:core detekt` — no new findings.

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
@@ -9,6 +9,8 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.ThreadFactory
 import java.util.concurrent.ThreadPoolExecutor
@@ -43,6 +45,8 @@ object OneSignalDispatchers {
         "$BASE_THREAD_NAME-IO" // Thread name prefix for I/O operations
     private const val DEFAULT_THREAD_NAME_PREFIX =
         "$BASE_THREAD_NAME-Default" // Thread name prefix for CPU operations
+    private const val SERIAL_IO_THREAD_NAME =
+        "$BASE_THREAD_NAME-SerialIO" // Single, named thread for order-sensitive work
 
     private class OptimizedThreadFactory(
         private val namePrefix: String,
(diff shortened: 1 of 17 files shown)


abdulraqeeb33:
Closing — the SessionService.onFocus / onUnfocused offload from this PR has been consolidated into #2644 (the unified onFocus / onUnfocused offload PR). All five lifecycle handlers (BackgroundManager, NotificationsManager, NotificationPermissionController, FeatureFlagsRefreshService, SessionService) now move together on the same FF in #2644.

PR #2656, opened 2026-06-01
fix: [SDK-4735] route requestPermission fallbackToSettings to settings when permanently denied

fadi-george:
# Description
## One Line Summary
Fix `requestPermission(fallbackToSettings = true)` so the "open settings" prompt actually appears when a permission is permanently denied.

Before, SDK wasn't showing settings prompt:
Screenshot 2026-05-29 at 7 36 51 PM

### Motivation
When a notification (or location) permission is permanently denied, `requestPermission(true)` is supposed to route the user to the app's settings. Two separate bugs prevented this:

1. **Settings dialog created then immediately dismissed (regression).** The fallback dialog was parented to the transparent `PermissionsActivity`, which finishes immediately after the reject callback. Its `onDestroy()` then called `AlertDialogPrepromptForAndroidSettings.dismissCurrentDialog()`, tearing down the dialog the instant it was shown. Symptom: `showSettings == true` in the debugger, but nothing visible on screen. This regressed in the Oct 2025 ViewModel refactor (`6fa661dfb`), which moved the activity-finish to after the callback and added the `onDestroy` dismiss; the pre-refactor flow finished the activity first and showed the dialog on the resumed host activity.

2. **Stuck permanently-denied state.** `shouldShowSettings` only persisted the "resolved denial" preference on the exact `shouldShowRequestPermissionRationale` `true -> false` transition. If a permission was already permanently blocked when OneSignal first observed it (denied via system Settings, in a prior session/version, or process death mid-prompt), that transition is never seen, the pref stays `false`, and the fallback is permanently stuck.

### Scope
- Affects the notification and location permission settings-fallback flow only.
- `NotificationPermissionController` / `LocationPermissionController`: the fallback dialog is now deferred until a non-`PermissionsActivity` host activity is in the foreground, then shown there.
- `PermissionsActivity.onDestroy -> dismissCurrentDialog()` removed, along with the now-dead `currentDialog`/`dismissCurrentDialog` tracking in `AlertDialogPrepromptForAndroidSettings`.
- `PermissionsViewModel` + `IPreferencesService`: new `PROMPTED_PERMISSION_` pref records that OneSignal has requested a permission at least once, used by `shouldShowSettings` to recover the permanently-blocked state without having witnessed the rationale transition.
- No public API changes. Normal first-prompt and grant flows are unchanged.
- Also corrects a broken `IModule` import in `LocationModule` and reformats its builder chains.

# Testing
## Unit testing
Added 3 cases to `PermissionsViewModelTests` covering the recovery path: shows settings when permanently blocked and previously prompted, does not show settings on a first denial when never prompted before, and persists the prompted-before flag after a request. Existing permission tests updated/retained.

## Manual testing
Built the demo app against the local SDK in Android Studio. Verified that, on an install whose notification permission was permanently denied, tapping PROMPT PUSH now surfaces the "open settings" dialog (and it stays visible) rather than nothing.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

Note: touches notification and location runtime-permission prompting (no dedicated checklist item exists for permissions).

# Checklist
## Overview
   - [x] PR does one thing
   - [x] Any Public API changes are explained in the PR details and conform to existing APIs (none)

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/activities/PermissionsActivity.kt
@@ -8,7 +8,6 @@ import androidx.activity.viewModels
 import androidx.core.app.ActivityCompat
 import androidx.lifecycle.lifecycleScope
 import com.onesignal.core.R
-import com.onesignal.core.internal.permissions.AlertDialogPrepromptForAndroidSettings
 import com.onesignal.core.internal.permissions.PermissionsViewModel
 import com.onesignal.core.internal.permissions.PermissionsViewModel.Companion.INTENT_EXTRA_ANDROID_PERMISSION_STRING
 import com.onesignal.core.internal.permissions.PermissionsViewModel.Companion.INTENT_EXTRA_CALLBACK_CLASS
@@ -69,12 +68,6 @@ class PermissionsActivity : ComponentActivity() {
         // Reset waiting state when activity loses focus
         // This ensures permission dialog can be shown again if activity was interrupted
         viewModel.resetWaitingState()
-    }
-
-    override fun onDestroy() {
-        super.onDestroy()
-        // Dismiss any active dialogs to prevent WindowLeaked errors
-        AlertDialogPrepromptForAndroidSettings.dismissCurrentDialog()
     }
 
     private fun finishActivity() {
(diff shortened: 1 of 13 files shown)


fadi-george:
@claude review

nan-li (approved):
I repro'd original issue and now it works as expected

PR #2662, opened 2026-06-17
chore: [SDK-4783] deprecate blocking OneSignal APIs in favor of suspend variants

abdulraqeeb33:
# Description
## One Line Summary
Mark the blocking (non-suspend) OneSignal public APIs as `@Deprecated`, pointing each to its suspend equivalent.

### Motivation
Closes [SDK-4783](https://linear.app/onesignal/issue/SDK-4783/mark-initwithcontext-as-deprecated).

The blocking APIs (`initWithContext`, `login`, `logout`, `updateUserJwt`, the manager accessors, and the config properties) can block the calling thread until the SDK is initialized, which risks ANRs when invoked on the main thread. The SDK already exposes suspend equivalents for all of them. This PR formally deprecates the blocking variants so developers get a compile-time signal (with an IDE quick-fix via `ReplaceWith`) and a runtime log nudging them toward the suspend APIs.

### Scope
Deprecations are applied consistently across all three layers:
- `IOneSignal` (interface contract)
- `OneSignalImp` (implementation overrides — also where the runtime `Logging.warn` deprecation notices live for the blocking methods)
- `OneSignal` (public static object app developers call)

Deprecated members and their replacements:

| Deprecated (blocking) | Replacement (suspend) |
| --- | --- |
| `initWithContext(context, appId)` | `initWithContextSuspend(context, appId)` |
| `login(externalId)` / `login(externalId, jwtBearerToken)` | `loginSuspend(...)` |
| `logout()` | `logoutSuspend()` |
| `updateUserJwt(externalId, token)` | `updateUserJwtSuspend(externalId, token)` |
| `User` / `Session` / `Notifications` / `Location` / `InAppMessages` | `getUserSuspend()` / `getSessionSuspend()` / `getNotificationsSuspend()` / `getLocationSuspend()` / `getInAppMessagesSuspend()` |
| `consentRequired` / `consentGiven` / `disableGMSMissingPrompt` | `getX Suspend()` / `setX Suspend(...)` |

**Not changed:**
- The internal-only suspend `initWithContext(context)` path used by the SDK itself.
- `debug` (no suspend equivalent; usable before init).
- Property accessors deliberately do **not** emit runtime logs (they can be read frequently and would flood logs); the `@Deprecated` annotation is the signal there.

### Notes
- Deprecation level is `WARNING` (not `ERROR`) so the SDK's own unit tests that exercise the blocking APIs still compile.
- All internal SDK callers already use the suspend variants, so main source compiles with zero new deprecation warnings.

# Testing
## Unit testing
No new tests; this is an annotation/logging change. Existing tests that call the blocking APIs continue to compile and pass (they now surface expected deprecation warnings).

## Manual testing
Verified `:OneSignal:core`, `:OneSignal:notifications`, `:OneSignal:location`, and `:OneSignal:in-app-messages` compile successfully (`compileDebugKotlin`) with no new deprecation warnings in SDK source.

# Affected code checklist
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -72,12 +72,15 @@ internal class OneSignalImp(
     override val isInitialized: Boolean
         get() = initState == InitState.SUCCESS
 
-    @Deprecated(
+    @get:Deprecated(
         message =
-        "Accessing this property may block the calling thread until the SDK is initialized and " +
-            "cause ANRs when called on the main thread. Use the suspend functions getConsentRequired() " +
-            "and setConsentRequired(required) instead.",
+            "Reading this property may block the calling thread until the SDK is initialized and " +
+                "cause ANRs when called on the main thread. Use the suspend function getConsentRequired() instead.",
         replaceWith = ReplaceWith("getConsentRequired()"),
+    )
+    @set:Deprecated(
+        message = "Use the suspend function setConsentRequired(required) instead.",
+        replaceWith = ReplaceWith("setConsentRequired(required)"),
     )
     override var consentRequired: Boolean
         get() =
@@ -93,12 +96,15 @@ internal class OneSignalImp(
             }
         }
 
-    @Deprecated(
+    @get:Deprecated(
         message =
-        "Accessing this property may block the calling thread until the SDK is initialized and " +
-            "cause ANRs when called on the main thread. Use the suspend functions getConsentGiven() " +
-            "and setConsentGiven(value) instead.",
+            "Reading this property may block the calling thread until the SDK is initialized and " +
+                "cause ANRs when called on the main thread. Use the suspend function getConsentGiven() instead.",
         replaceWith = ReplaceWith("getConsentGiven()"),
+    )
+    @set:Deprecated(
+        message = "Use the suspend function setConsentGiven(value) instead.",
+        replaceWith = ReplaceWith("setConsentGiven(value)"),
     )
     override var consentGiven: Boolean
         get() =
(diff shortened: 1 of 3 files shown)


sherwinski:
@claude review once

abdulraqeeb33:
@claude[agent] review

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -72,6 +72,16 @@ internal class OneSignalImp(
...
 
+    @get:Deprecated(
+        message =
+            "Reading this property may block the calling thread until the SDK is initialized and " +
+                "cause ANRs when called on the main thread. Use the suspend function getConsentRequired() instead.",
+        replaceWith = ReplaceWith("getConsentRequired()"),
+    )
+    @set:Deprecated(
+        message = "Use the suspend function setConsentRequired(required) instead.",
+        replaceWith = ReplaceWith("setConsentRequired(required)"),
```
sherwinski: claude seems to think this (and the other two `ReplaceWith(setConsentRequired())` instances) should be `value` instead of `required`

sherwinski (approved):

nan-li:
re-ran tests

fadi-george:
Why not just replace the original methods

fadi-george (changes requested):
Why not just replace the original methods?
going forward people will use the new methods like OneSignal.User. loginSuspend ?
Thats really ugly

abdulraqeeb33:
> Why not just replace the original methods? going forward people will use the new methods like OneSignal.User. loginSuspend ? Thats really ugly

we can't really rename the original methods, it will break backward compatibility. Our original plan was to slowly nudge the users to start using the suspend methods. And hopefully in a major release we get rid of the old methods.

PR #2663, opened 2026-06-17
fix: [SDK-4672] start location updates when permission is granted

nan-li:
# Description
## One Line Summary
When registering for ongoing location updates fails, the SDK now records the failure and registers again the next time location starts, instead of assuming the first attempt worked and never retrying.

### Motivation
If an app turns on location sharing (`isShared = true`) before the user has granted location permission, the SDK tries to register for ongoing location updates, but that call fails because the permission is missing. The failure was caught but not reported back, so the SDK recorded the location-updates listener as registered when it actually was not. After the user granted permission, the SDK sent a single saved location and did not register for ongoing updates again, so no further location changes were reported until the next time the app was sent to the background and brought back to the foreground.

This change makes the registration call report whether it actually succeeded, so the SDK no longer treats a failed attempt as registered, and it registers for ongoing updates again as soon as permission is granted. Ongoing updates now start right away.

### Scope
- Affects the Google Play Services location path only; the Huawei path is unchanged.
- Ongoing updates now start at the moment permission is granted.
- No public API changes.

# Testing
## Unit testing
Added two tests in `GmsLocationControllerTests`: one checks that calling `start()` again registers for location updates a second time, and one checks that a failed registration is not recorded as active. All location unit tests and detekt pass.

## Manual testing
Reproduced on the demo app (`com.onesignal.example`, built from this branch) on a Pixel 7 / Android 14 emulator: turned on `isShared` before granting permission, then chose "While using the app". Before the change, ongoing updates started only after the app was sent to the background and brought back; after the change, the SDK registers for ongoing updates right after the grant with no error, and the latitude/longitude reaches OneSignal in the same foreground session.

<details>
<summary>Verification logs</summary>

Turn on `isShared` before permission — the registration call fails, so no ongoing updates are started:

```text
16:37:45.358 D LocationManager.setIsShared(value: true)
16:37:45.358 D LocationManager.startGetLocation()
16:37:45.388 D GMSLocationController GoogleApiClientListener onConnected
16:37:45.395 D GMSLocationController GoogleApiClient requestLocationUpdates!
16:37:45.398 W FusedLocationApi.requestLocationUpdates failed!
16:37:45.398 W java.lang.SecurityException: uid 10204 does not have any of
              [ACCESS_FINE_LOCATION, ACCESS_COARSE_LOCATION]
```

Grant "While using the app" — the SDK registers for ongoing updates immediately, with no background/foreground cycle in between:

```text
16:38:15.663 D ...key=USER_RESOLVED_PERMISSION_android.permission.ACCESS_FINE_LOCATION
16:38:15.664 D LocationManager.startGetLocation()
16:38:15.676 D ApplicationService.onActivityResumed: MainActivity
16:38:15.684 D LocationController fireCompleteForLocation: Location[fused 37.421998,-122.084 ...]
16:38:15.695 D GMSLocationController GoogleApiClient requestLocationUpdates!
16:38:20.699 D HttpClient: PATCH .../users/...  Body: {"properties":{"lat":37.4219983,"long":-122.084}}
16:38:20.917 D HttpClient: PATCH ... STATUS: 202
```

The `requestLocationUpdates!` at the grant has no `SecurityException` after it, and there is no `LocationUpdateListener.onFocus()`/`onUnfocused()` between the grant and the re-registration — so ongoing updates start from the permission grant itself, not from a background/foreground cycle.
</details>

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing
   - [x] No public API changes

## Testing
   - [x] I have included test coverage for these changes
   - [x] All automated tests pass
   - [x] I have personally tested this on an emulator

## Final pass
   - [x] I have reviewed this PR myself

### DIFF
--- a/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/FusedLocationApiWrapperImpl.kt
+++ b/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/FusedLocationApiWrapperImpl.kt
@@ -24,16 +24,20 @@ internal class FusedLocationApiWrapperImpl : IFusedLocationApiWrapper {
         googleApiClient: GoogleApiClient,
         locationRequest: LocationRequest,
         locationListener: LocationListener,
-    ) {
-        try {
+    ): Boolean {
+        return try {
             if (Looper.myLooper() == null) {
                 Looper.prepare()
             }
             if (googleApiClient.isConnected) {
                 LocationServices.FusedLocationApi.requestLocationUpdates(googleApiClient, locationRequest, locationListener)
+                true
+            } else {
+                false
             }
         } catch (t: Throwable) {
             Logging.warn("FusedLocationApi.requestLocationUpdates failed!", t)
+            false
         }
     }
 
--- a/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/GmsLocationController.kt
+++ b/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/GmsLocationController.kt
@@ -56,6 +56,10 @@ internal class GmsLocationController(
                             setLocationAndFire(localLastLocation)
                         }
                     }
+                    // Retry the subscription in case the first attempt failed because
+                    // permission was not yet granted, otherwise it would only recover on
+                    // the next app focus change.
+                    locationUpdateListener?.refreshRequest()
                     wasSuccessful = true
                 } else {
                     try {
@@ -193,7 +197,7 @@ internal class GmsLocationController(
             }
         }
 
-        private fun refreshRequest() {
+        internal fun refreshRequest() {
             if (!googleApiClient.isConnected) {
                 Logging.warn("Attempt to refresh location request but not currently connected!")
                 return
@@ -217,8 +221,7 @@ internal class GmsLocationController(
                     .setMaxWaitTime((updateInterval * 1.5).toLong())
                     .setPriority(LocationRequest.PRIORITY_BALANCED_POWER_ACCURACY)
             Logging.debug("GMSLocationController GoogleApiClient requestLocationUpdates!")
-            _fusedLocationApiWrapper.requestLocationUpdates(googleApiClient, locationRequest, this)
-            hasExistingRequest = true
+            hasExistingRequest = _fusedLocationApiWrapper.requestLocationUpdates(googleApiClient, locationRequest, this)
         }
 
         override fun onLocationChanged(location: Location) {
--- a/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/GmsLocationController.kt
+++ b/OneSignalSDK/onesignal/location/src/main/java/com/onesignal/location/internal/controller/impl/GmsLocationController.kt
@@ -168,6 +168,8 @@ internal class GmsLocationController(
         private val googleApiClient: GoogleApiClient,
         private val _fusedLocationApiWrapper: IFusedLocationApiWrapper,
     ) : LocationListener, IApplicationLifecycleHandler, Closeable {
+        // Guards hasExistingRequest and the cancel/register sequence against concurrent threads.
+        private val requestLock = Any()
         private var hasExistingRequest = false
 
         init {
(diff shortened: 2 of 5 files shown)


nan-li:
@claude review

fadi-george (approved):

fadi-george:
Is this similar to the other location pr?

nan-li:
> Is this similar to the other location pr?

No, that one catches throwables in LocationServices API only. 

This one calls an API `requestLocationUpdates` that already is wrapped in a try-catch but it has to do with making sure the SDK actually subscribes to location updates instead of silently failing. I updated the **Motivation** section to be more clear.

abdulraqeeb33 (approved):
Hey Nan, took a pass through this with a couple of reviewers and overall it's a solid fix — really like that requestLocationUpdates now reports real success instead of us optimistically assuming it worked. The re-arm-on-grant flow makes a lot of sense and the manual repro in the description is super clear. Nice work tracking this one down.

A few things I'd love your thoughts on before we ship:

Re-arm churn on every start() — since the early-path now calls refreshRequest() unconditionally, and refreshRequest() cancels + re-registers (and resets the interval timer) even when the subscription is already healthy, repeated start() calls (toggling isShared, permission callbacks) end up churning the subscription. Could we guard it so we only re-arm when hasExistingRequest == false?

hasExistingRequest threading — it's an unsynchronized var that's now touched from the IO start() path (under startStopMutex) and from onFocus/onUnfocused on the main thread (no lock). This race technically predates your change, but this PR widens it a bit. Worth at least a @Volatile, or serializing the refresh through the same lock?

The new tests — the mocked applicationService no-ops addApplicationLifecycleHandler, but the real one fires onFocus(true) synchronously, so the callCount shouldBe 2 assertion doesn't quite mirror production's re-entrant path. Might be worth a test that exercises that flow + asserts the failed-first / success-second transition directly.

None of these are blockers from my side

abdulraqeeb33 (approved):
Hey Nan, took a pass through this with a couple of reviewers and overall it's a solid fix — really like that requestLocationUpdates now reports real success instead of us optimistically assuming it worked. The re-arm-on-grant flow makes a lot of sense and the manual repro in the description is super clear. Nice work tracking this one down.

A few things I'd love your thoughts on before we ship:

Re-arm churn on every start() — since the early-path now calls refreshRequest() unconditionally, and refreshRequest() cancels + re-registers (and resets the interval timer) even when the subscription is already healthy, repeated start() calls (toggling isShared, permission callbacks) end up churning the subscription. Could we guard it so we only re-arm when hasExistingRequest == false?

hasExistingRequest threading — it's an unsynchronized var that's now touched from the IO start() path (under startStopMutex) and from onFocus/onUnfocused on the main thread (no lock). This race technically predates your change, but this PR widens it a bit. Worth at least a @Volatile, or serializing the refresh through the same lock?

The new tests — the mocked applicationService no-ops addApplicationLifecycleHandler, but the real one fires onFocus(true) synchronously, so the callCount shouldBe 2 assertion doesn't quite mirror production's re-entrant path. Might be worth a test that exercises that flow + asserts the failed-first / success-second transition directly.

None of these are blockers from my side

fadi-george:
@nan-li merge?

nan-li:
Thanks for the careful review, all three were fair, and I've pushed a change for each:

**1. Re-arm churn on every start()** — done. `refreshRequest()` now takes an `onlyIfInactive` flag, and the early `start()` path passes `onlyIfInactive = true`, so an already-active request is left in place instead of being cancelled and re-registered (which also reset the interval) on repeated `start()`s. Grant-recovery still works: after a failed attempt the request isn't active, so it re-registers there.

**2. hasExistingRequest threading** — done. `@Volatile` alone wouldn't cover it, since cancel → register → assign is a check-then-act, so I moved `hasExistingRequest` and that whole sequence behind a dedicated lock shared by `refreshRequest()` and `close()`. That serializes the IO `start()/stop()` path with the main-thread `onFocus()/onUnfocused()` callbacks (the coroutine mutex isn't usable there since those callbacks aren't suspend).

**3. Tests** — done. Added a regression test that drives the real transition — a first `requestLocationUpdates` that fails, then a second `start()` that succeeds — asserting failed-first → success-second directly. I also flipped the old "re-arms on a subsequent start" test (which was really asserting the churn from #1) to verify a healthy request is left alone.

PR #2666, opened 2026-06-22
fix: [SDK-4793] prevent main-thread ANR in blocking SDK APIs during/after init

abdulraqeeb33:
## Summary

Under the `sdk_background_threading` feature flag, OneSignal's synchronous public APIs blocked the calling thread on `runBlocking(OneSignalDispatchers.IO)` to wait for initialization — even after init had already reached `SUCCESS`, and even though the wait itself needs no IO work. On Flutter, the plugin resolves managers synchronously on the platform/main thread in `lifecycleInit`, so during a cold-start-contended IO pool the main thread parks waiting for a free IO worker, producing ANRs.

Affected entry points (all reached only under `sdk_background_threading`):
- Service accessors `getUser()` / `getInAppMessages()` / `getNotifications()` / `getSession()` / `getLocation()`, and `login` / `logout` / `updateUserJwt` / JWT-invalidated listeners — via `waitForInit`.
- Consent getters `consentRequired` / `consentGiven` / `disableGMSMissingPrompt` — via `blockingGet`.

- Linear: SDK-4793
- Reported in Flutter SDK: OneSignal/OneSignal-Flutter-SDK#1163 (5.6.0, 20k+ users, Play Console vitals anomaly)

### Reported stack trace (both reported traces share this shape)

```
main (Timed Waiting)
  LockSupport.parkNanos
  kotlinx.coroutines.BlockingCoroutine.joinBlocking
  kotlinx.coroutines.BuildersKt.runBlocking
  OneSignalImp.waitForInit
  OneSignalImp.waitAndReturn
  OneSignalImp.getServiceWithFeatureGate
  OneSignalImp.getInAppMessages / getUser
  OneSignalInAppMessages.lifecycleInit / OneSignalUser.lifecycleInit  (Flutter plugin, main thread)
  MethodChannel ... Looper.loop ... ActivityThread.main
```

## Root cause

Two compounding issues in the blocking wait primitives (`waitForInit`, `blockingGet`):

1. **No `SUCCESS` short-circuit.** The FF-on path routed every call through `runBlocking(runtimeIoDispatcher)` even when init had long since finished. (The legacy FF-off accessor branch already returned directly at `SUCCESS`.)
2. **The wait was dispatched onto `OneSignalDispatchers.IO`.** The wait body only awaits the init-completion signal — it needs no IO worker — yet it was scheduled on the IO pool. On a saturated pool the caller parks until a worker frees, both to *start* the wait and to *deliver the result after init completes*.

## Fix

Centralized in the two blocking primitives, so every entry point above is covered:

1. **`SUCCESS` fast-path.** `SUCCESS` is terminal and never flips back (a re-`initWithContext` short-circuits on `initState.isSDKAccessible()`; only `FAILED -> IN_PROGRESS` retries exist), so return synchronously without `runBlocking`. `initState` is `@Volatile`, so the read is safe and lock-free — identical to what the FF-off branch already relies on.
2. **Decouple the wait from the IO pool.** For the non-`SUCCESS` path, use plain `runBlocking { ... }` so the wait runs on the caller's own event loop and resumes off the init-completion signal alone, never blocked on IO-worker availability.

`NOT_STARTED` / `FAILED` still throw and `IN_PROGRESS` still blocks until init completes, preserving the existing block-or-throw semantics.

## Impact classification

**Not a crash.** The main thread is parked inside `runBlocking`, bottoming out at `Looper.loop` / `ActivityThread.main`, so the OS surfaces this as an **ANR** (Android vitals anomaly), not a fatal force-close. Self-recovering once init completes and an IO worker frees. **Severity: High** — main-thread block on the hottest startup path with broad user-visible impact.

## Test plan

- [x] `SDKInitTests` saturates the real `OneSignalDispatchers.IO` and asserts SUCCESS-state accessors / `login` / consent getters return synchronously.
- [x] `OneSignalImpTests` injects a throwing `CoroutineDispatcher` to prove the SUCCESS fast-path never dispatches, plus a new IN_PROGRESS regression test proving the wait resumes off the completion signal without touching the IO dispatcher.
- [x] Each new test verified to **fail** without its corresponding fix and **pass** with it.
- [x] Full `:onesignal:core` unit test suite passes.

## Known limitation / follow-ups

- Under FF-on, init itself runs on `OneSignalDispatchers.IO` (`suspendifyOnIO`). If that pool is *fully* saturated, init cannot complete and an `IN_PROGRESS` caller must still wait — no waiter-side change can avoid that. This PR removes the SUCCESS-state park and the post-completion IO dependency; the durable fix is Flutter-side: resolve managers off the platform thread in `lifecycleInit` (OneSignal-Flutter-SDK).
- Optional hardening (not in this PR): bound the wait with a timeout so a wedged init degrades to a logged error instead of an indefinite block.

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -609,6 +609,19 @@ internal class OneSignalImp(
      * @param operationName Optional operation name to include in error messages (e.g., "login", "logout")
      */
     private fun waitForInit(operationName: String? = null) {
+        // Fast-path: once init reaches SUCCESS the state is terminal and never flips back (a
+        // re-initWithContext short-circuits on initState.isSDKAccessible(); only FAILED ->
+        // IN_PROGRESS retries exist), so there is nothing to await. Return synchronously instead
+        // of hopping onto the (cold-start-contended) IO dispatcher via runBlocking, which parks
+        // the calling thread -- e.g. Flutter's main-thread lifecycleInit -- and is the ANR in
+        // OneSignal-Flutter-SDK#1163. Centralizing the fast-path here covers every blocking entry
+        // point (service accessors via waitAndReturn, plus login / logout / updateUserJwt /
+        // add|removeUserJwtInvalidatedListener). The non-terminal states (IN_PROGRESS / FAILED /
+        // NOT_STARTED) still go through waitUntilInitInternal, preserving block-until-ready / throw
+        // semantics. initState is @Volatile, so the read is safe and lock-free.
+        if (initState == InitState.SUCCESS) {
+            return
+        }
         runBlocking(runtimeIoDispatcher) {
             waitUntilInitInternal(operationName)
         }
@@ -726,14 +739,9 @@ internal class OneSignalImp(
 
     private fun <T> getServiceWithFeatureGate(getter: () -> T): T {
         if (isBackgroundThreadingEnabled) {
-            // Once init reaches SUCCESS the state is terminal and never flips back, so the
-            // requested service is already constructable. Answer synchronously instead of
-            // hopping onto the (cold-start-contended) IO dispatcher via runBlocking, which
-            // parks the calling thread -- e.g. Flutter's main-thread lifecycleInit -- and is
-            // the ANR in OneSignal-Flutter-SDK#1163. Only the genuinely-not-yet-ready states
-            // (IN_PROGRESS / FAILED / NOT_STARTED) need waitAndReturn, which preserves the
-            // existing block-or-throw semantics.
-            return if (initState == InitState.SUCCESS) getter() else waitAndReturn(getter)
+            // waitAndReturn -> waitForInit short-circuits synchronously once init is SUCCESS
+            // (see waitForInit), so this no longer parks the caller on the IO dispatcher.
+            return waitAndReturn(getter)
         }
         return when (initState) {
             InitState.SUCCESS -> {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -613,7 +613,10 @@ internal class OneSignalImp(
         if (initState == InitState.SUCCESS) {
             return
         }
-        runBlocking(runtimeIoDispatcher) {
+        // Wait on the caller's own runBlocking event loop, not OneSignalDispatchers.IO: the body
+        // only awaits the init-completion signal, so coupling it to a saturated IO pool can park
+        // the caller even after init finishes (OneSignal-Flutter-SDK#1163).
+        runBlocking {
             waitUntilInitInternal(operationName)
         }
     }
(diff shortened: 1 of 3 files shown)


fadi-george (dismissed):

fadi-george:
@claude review

fadi-george:
I think we should also address `blockingGet()` before merging, or narrow the comment in `waitForInit()`.

The new `SUCCESS` fast path covers `waitForInit()` callers, but `consentRequired`, `consentGiven`, and `disableGMSMissingPrompt` still go through `blockingGet()`, which unconditionally does `runBlocking(runtimeIoDispatcher)`. With `sdk_background_threading` enabled, that still means `runBlocking(OneSignalDispatchers.IO)` after init has already reached `SUCCESS`, so those getters can hit the same saturated-IO/main-thread parking shape this PR is fixing.

Could we either add the same `SUCCESS` synchronous path to `blockingGet()`, or update the comment so it does not claim this covers every blocking entry point?

fadi-george (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -609,7 +609,14 @@ internal class OneSignalImp(
      * @param operationName Optional operation name to include in error messages (e.g., "login", "logout")
      */
     private fun waitForInit(operationName: String? = null) {
-        runBlocking(runtimeIoDispatcher) {
+        // SUCCESS is terminal, so initialized callers can return without parking behind the IO dispatcher.
```
nan-li: nit on comments that discuss past implementations, they will start to clutter the SDK, and these aren't really relevant for a current or future viewer. I would drop this line since `if (initState == InitState.SUCCESS)` feels pretty clear and reading "parking behind the IO dispatcher" would be confusing for someone unless they knew the previous implementation, as there is no usage of IO dispatcher in this method. And same for the 3-line comment below, I would drop or condense

nan-li (approved):

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
```diff
@@ -46,6 +50,46 @@ class SDKInitTests : FunSpec({
...
+            }
+        thread.start()
+        thread.join(2_000)
+
+        thread.isAlive shouldBe false
+        error.get() shouldBe null
+    }
+
     beforeAny {
         Logging.logLevel = LogLevel.NONE
```
sherwinski, line 94: nit: this comment is no longer accurate. The new `enableBackgroundThreadingBeforeInit` helper flips the FF on without having to go through SharedPreferences. We should update this to avoid confusion.

sherwinski (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
abdulraqeeb33: Good call — dropped the SUCCESS-line comment (the `if (initState == InitState.SUCCESS)` check speaks for itself) and condensed the block below to a short present-tense note ("intentionally dispatcher-less") with no reference to the prior implementation. Done in 8bdace27c.

fadi-george (approved):

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/application/SDKInitTests.kt
abdulraqeeb33: Good catch — updated the comment to reflect the `enableBackgroundThreadingBeforeInit` helper (seeds `configModelStore.model.sdkRemoteFeatureFlags` directly, no SharedPreferences planting) and dropped the obsolete bit about it conflicting with `BlockingPrefsContext`. Done in 846a9c0ff.

abdulraqeeb33: done

PR #2676, opened 2026-06-29
fix: [SDK-4822] make ANR watchdog app-state aware to stop false background ANRs

abdulraqeeb33:
## Summary

The Otel ANR watchdog (`OtelAnrDetector`) monitored the main thread continuously with **no awareness of foreground/background**, so any backgrounded main-thread block was reported as a crash-class ANR. A pull of ~25k recent ANRs showed ~87.5% are backgrounded main-thread, and the long-duration reports (median block ~6 min, up to ~25h) are impossible as real hangs — they are freeze/suspension artifacts. The net effect was an ANR metric inflated well beyond what Android actually counts (user-perceived ANRs are foreground input-dispatch timeouts).

## Approach

Detection is now **app-state aware**, and background blocks are **downgraded but not discarded**:

- **Foreground** block > `5s` → crash-class ANR (unchanged).
- **Background** block > `10s` → recorded via a distinct `BackgroundMainThreadBlockException` on the **same retained, disk-buffered crash path**, so it stays out of the ANR/crash metric but remains queryable. `app.state` + SDK version are auto-attached, and the exception message carries a compact fingerprint (top frame + first OneSignal frame) for triage. This preserves visibility into real background regressions (broadcast `onReceive`, `JobScheduler`, main-thread SQLite, lock/binder waits) instead of dropping them.
- **Frozen-process guard:** if the watchdog thread's own sleep overran the check interval beyond a 2s slack, the process was descheduled (Doze / cached) rather than the main thread being stuck → the measurement is suppressed.
- **Monotonic clock** (`SystemClock.uptimeMillis`) instead of wall-clock, so NTP / DST / manual clock changes and deep sleep can't fabricate block time.
- **Independent foreground/background dedup** (see review thread): a stream of background warnings can never suppress a real foreground ANR.

Foreground state is derived from the existing `platformProvider.appState`; `"unknown"` is treated as foreground so a genuine ANR is never silently downgraded.

## Testability & coverage

- Timing / classification / dedup extracted into a pure, Android-free `AnrCheckEvaluator` (100% JVM-covered).
- Android touch points (main-thread `Handler`, stack capture, monotonic clock) abstracted behind an injectable `AnrWatchdogPlatform` seam so the whole detector runs under **plain-JVM tests instead of Robolectric** — which is what lets JaCoCo actually measure it. The shell went from **0% → ~73%**, and aggregate diff coverage is **~84%** (was 0%, which was failing the gate).

## Test plan

- [x] `./gradlew :OneSignal:core:testDebugUnitTest` — all pass
- [x] Diff coverage gate passes (~84% aggregate on touched lines)
- [x] Evaluator matrix: fg ANR, responsive, bg-warning threshold, frozen-process precedence + slack boundary, **independent fg/bg dedup**, post-boot dedup sentinel, fingerprint
- [x] Shell wiring: `reportAnr` (OneSignal-at-fault / not), background-warning path, start/stop lifecycle, unknown-state → foreground

Closes SDK-4822.

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/AnrCheckEvaluator.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/AnrCheckEvaluator.kt
@@ -1,0 +1,182 @@
+package com.onesignal.debug.internal.crash
+
+import java.util.concurrent.atomic.AtomicLong
+
+/**
+ * Pure, Android-free decision core for the ANR watchdog.
+ *
+ * All timing/classification/deduplication state lives here so it can be exercised deterministically
+ * on the JVM (with an injected clock) without a `Handler`, `Looper`, or real background thread. The
+ * Android shell ([OtelAnrDetector]) owns the thread, the real sleep, and the reporting side effects,
+ * and delegates every per-iteration decision to [evaluate].
+ *
+ * Foreground and background blocks keep independent dedup timestamps: a stream of backgrounded
+ * warnings must never suppress a genuine foreground ANR (and vice versa).
+ */
+internal class AnrCheckEvaluator(
+    private val anrThresholdMs: Long,
+    private val checkIntervalMs: Long,
+    private val backgroundThresholdMs: Long,
+    private val frozenSlackMs: Long,
+    private val dedupWindowMs: Long,
+    private val now: () -> Long,
+) {
+    // Monotonic timestamps (from `now`); see OtelAnrDetector for why the clock must be monotonic.
+    private val lastResponseTime = AtomicLong(now())
+    private val lastForegroundReportTime = AtomicLong(NEVER_REPORTED)
+    private val lastBackgroundReportTime = AtomicLong(NEVER_REPORTED)
+
+    /** Re-baselines the responsiveness clock, e.g. at start() so a construction->start gap isn't a block. */
+    fun resetBaseline() {
+        lastResponseTime.set(now())
+    }
+
+    /** Records that the main thread ran the heartbeat runnable. */
+    fun recordHeartbeat() {
+        lastResponseTime.set(now())
+    }
+
+    /**
+     * Evaluates a single watchdog iteration given how long the watchdog's own sleep actually took and
+     * the current app state. Returns the action the shell should take; all dedup bookkeeping is done
+     * here so the shell only performs side effects.
+     */
+    fun evaluate(actualSleepMs: Long, inForeground: Boolean): AnrCheckResult {
+        val timeSinceLastResponse = now() - lastResponseTime.get()
+
+        return when (
+            classifyBlock(
+                timeSinceLastResponseMs = timeSinceLastResponse,
+                actualSleepMs = actualSleepMs,
+                checkIntervalMs = checkIntervalMs,
+                frozenSlackMs = frozenSlackMs,
+                anrThresholdMs = anrThresholdMs,
+                backgroundThresholdMs = backgroundThresholdMs,
+                inForeground = inForeground,
+            )
+        ) {
+            BlockClassification.FROZEN_PROCESS -> {
+                // The watchdog thread itself was descheduled (Doze / cached-process freeze). Anything
+                // measured for the main thread is a freeze artifact, so re-baseline and treat as
+                // responsive to avoid firing on the next iteration.
+                lastResponseTime.set(now())
+                AnrCheckResult.FrozenProcess(actualSleepMs = actualSleepMs, expectedSleepMs = checkIntervalMs)
+            }
+            BlockClassification.RESPONSIVE -> {
+                clearReportTimestamps()
+                AnrCheckResult.Responsive
+            }
+            BlockClassification.FOREGROUND_ANR ->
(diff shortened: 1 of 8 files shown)


fadi-george:
One concern: `lastAnrReportTime` is shared between background warnings and foreground ANRs. If a background block is recorded, then the app foregrounds while the main thread is still blocked, the foreground crash-class ANR can be suppressed for 30s. That seems to conflict with the intent to keep foreground ANRs reported normally.
Also worth validating the app-state rule: Android can still count some non-foreground work as ANR-relevant, such as broadcasts or foreground services. If OneSignal can block the main thread in those paths, downgrading solely on `appState == "background"` may underreport real ANRs.

abdulraqeeb33:
Thanks Fadi — both good catches, and I reworked the PR around them.

**1. Shared dedup → fixed.** Foreground and background now keep independent dedup timestamps (`lastForegroundReportTime` / `lastBackgroundReportTime` in the new `AnrCheckEvaluator`), so a background warning can never suppress a foreground ANR. If the app foregrounds while the main thread is still blocked, the foreground ANR reports on the next check regardless of any recent background warning. Covered by a dedicated test: *"foreground and background dedup are independent — a bg warning never suppresses a fg ANR"*.

**2. Non-foreground ANR-relevant work → we no longer drop it.** You're right that Android can count some non-foreground work (broadcast `onReceive`, foreground services, `JobScheduler`) and that `appState == "background"` is coarser than that. The important change is that background blocks are no longer discarded — they're routed to a **retained, queryable stream** as a distinct `BackgroundMainThreadBlockException` (same disk-buffered crash path, with `app.state` + SDK version auto-attached and a top-frame / first-OneSignal-frame fingerprint in the message). So we keep full visibility into genuine background main-thread blocks; they're just categorized out of the *user-perceived* ANR metric rather than lost. The data pull supports this: the plausible 5–60s background window does contain real OneSignal main-thread blocks, and this keeps them observable and groupable. We only have foreground/background/unknown granularity from `platformProvider.appState` today (`unknown` → treated as foreground so real ANRs are never downgraded); if the retained stream shows a specific broadcast/FGS path is user-impacting, we can promote those to ANR-class in a follow-up instead of guessing now.

Also worth noting: this pass makes the detector JVM-testable (pure `AnrCheckEvaluator` + an injectable `AnrWatchdogPlatform` seam), which is what finally gives the watchdog real coverage (0% → ~84% aggregate on the diff) instead of the Robolectric 0%.

fadi-george:
Thanks for the update. The independent foreground/background dedupe addresses my main concern.

One follow-up: `reportBackgroundBlock` now uses `crashReporter.saveCrash`, but `OtelCrashReporter.saveCrash` hardcodes `Severity.FATAL`. Since the PR describes these as retained background warnings that stay out of ANR/crash metrics, can we confirm the backend segments `BackgroundMainThreadBlockException` independently and ignores the fatal severity for crash/ANR metrics? If not, this may need an explicit non-fatal/warning reporter path.

I’d also still want product/Android validation on the app-state rule. The new background stream preserves visibility, but the SDK still downgrades solely on `appState == "background"`, while some background entry points like broadcasts/foreground services can still be ANR-relevant.

abdulraqeeb33:
Good catch — you were right, the FATAL severity wasn't enforced anywhere. Fixed in 08cf970.

Background blocks now go through a new `saveNonFatal` path: emitted at `WARN` with an SDK-owned `ossdk.crash.fatal=false` attribute. Real crashes and foreground ANRs stay on `saveCrash` (FATAL, flag `true`). So the backend can exclude on severity *or* that flag, not just the exception classname.

One thing I confirmed while in there: crash reports and normal remote logs go through the same network exporter — there's no separate "crash" channel, so the metric is a query over severity/attributes. WARN + the flag should drop these out cleanly, but if you can confirm the metric keys on one of those I'll call it closed.

Still want your/product read on the app-state rule. We use a flat 5s foreground / 10s background threshold, not Android's per-context timeouts (broadcast 10s, service 20/200s) — so a backgrounded broadcast/FGS main-thread hang that's technically ANR-relevant would land as a non-fatal background block. Upside: it's retained and queryable, so we can promote specific paths to ANR-class once the data shows real user impact instead of guessing now. OK to ship that as the default?

fadi-george:
One API-compat concern: `IOtelCrashReporter` is public in the published `:OneSignal:otel` artifact, and adding `saveNonFatal(...)` is a breaking interface change for any external implementation.
If this module/API is treated as internal-only, that may be fine. Otherwise, can we avoid changing the public interface, or confirm this break is acceptable for the release?

abdulraqeeb33:
> One API-compat concern: `IOtelCrashReporter` is public in the published `:OneSignal:otel` artifact, and adding `saveNonFatal(...)` is a breaking interface change for any external implementation. If this module/API is treated as internal-only, that may be fine. Otherwise, can we avoid changing the public interface, or confirm this break is acceptable for the release?

Yeah this is internal only and currently used only by Android SDK.

nan-li (approved):
LGTM, just calling out that (currently) making background ANRs WARN-level effectively means we're not going to capture them in telemetry to the same extent anymore, because we have so few apps on log level WARN. It may still be enough to actually get usable metrics. And if we were to bump the loglevel to WARN for more apps, the volume would be too great. These would be meaningfully captured if we do something different with our logging in the future.

fadi-george (approved):

PR #2678, opened 2026-06-30
fix: [SDK-4734] move HMS notification finish() inside coroutine so entryState is set before trampoline stops

abdulraqeeb33:
## Summary

Fixes [SDK-4734](https://linear.app/onesignal/issue/SDK-4734).

The HMS notification-open trampoline (`NotificationOpenedActivityHMS`) previously called `finish()` **synchronously**, before the background coroutine processed the open. This created a race:

- The trampoline's `onStop` could reach `ApplicationService` while `entryState` was still `APP_CLOSE`.
- Because the entry state had not yet been set to `NOTIFICATION_CLICK`, the stale-entry reset path was skipped.
- The `NOTIFICATION_CLICK` that the coroutine later set then lingered, causing the **next organic launch to be mis-attributed as a DIRECT notification session**.

This change moves `finish()` inside the coroutine, running it on the main thread only **after** `handleHMSNotificationOpenIntent` completes. That guarantees `entryState` is `NOTIFICATION_CLICK` before `onStop` runs, so the reset path executes correctly and attribution is accurate.

## Test plan

This needs verification on **Huawei (HMS) hardware** since the trampoline path only runs on HMS devices. Manually verify the three scenarios:

- [ ] **URL / browser-only notification**: Tap an HMS push whose action opens a URL in the browser. Confirm the app does not retain a stale `NOTIFICATION_CLICK` entry, and a subsequent organic (launcher) app open is attributed as a normal session, not DIRECT.
- [ ] **Foreground-app notification**: With the app in the foreground, tap an HMS push. Confirm the open is processed correctly and the next organic launch is not mis-attributed as DIRECT.
- [ ] **Background-processed notification**: With the app backgrounded/closed, tap an HMS push so the trampoline cold-starts the process. Confirm the open is processed, the activity finishes only after processing, and a later organic launch is attributed correctly (not DIRECT).

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
@@ -30,6 +30,7 @@ package com.onesignal
 import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
+import com.onesignal.common.AndroidUtils
 import com.onesignal.common.threading.OneSignalDispatchers
 import com.onesignal.common.threading.suspendifyOnDefault
 import com.onesignal.core.internal.application.OneSignalInternalActivity
@@ -75,11 +76,6 @@ class NotificationOpenedActivityHMS :
     }
 
     private fun processIntent() {
-        processOpen(intent)
-        finish()
-    }
-
-    private fun processOpen(intent: Intent?) {
         // HMS notification-open trampoline runs on the main thread and can cold-start the process;
         // warm dispatchers before the first suspendifyOnDefault dispatch.
         OneSignalDispatchers.prewarm()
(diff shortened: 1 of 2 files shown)


fadi-george (approved):
Nit: the HMS fix looks sound, but NotificationOpenedActivityBase still has the same early-return-before-finish shape for the GMS/base trampoline. If OneSignal.initWithContext(...) returns false, AndroidUtils.finishSafely(...) is skipped there.

Might be worth either applying the same try/finally pattern here too, or filing a follow-up since it’s adjacent but outside the HMS-specific fix.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/NotificationOpenedActivityHMS.kt
```diff
@@ -75,22 +76,29 @@ class NotificationOpenedActivityHMS :
...
+                    return@suspendifyOnDefault
+                }
 
-            val notificationPayloadProcessorHMS = OneSignal.getService<INotificationOpenedProcessorHMS>()
-            val self = this
-            notificationPayloadProcessorHMS.handleHMSNotificationOpenIntent(self, intent)
+                val notificationPayloadProcessorHMS = OneSignal.getService<INotificationOpenedProcessorHMS>()
+                notificationPayloadProcessorHMS.handleHMSNotificationOpenIntent(this, intent)
+            } finally {
+                // Finish only after the open is fully processed (or an early return / failure),
```
nan-li, line 91: nit: This comment is quite long and a future reader likely find confusing without the knowledge of this PR and bug, it explains the previous bug, the lifecycle race, attribution impact, failure behavior, threading, and finally semantics all inline. Can we simplify?

PR #2687, opened 2026-07-21
feat(logger): [SDK-4835] adopt shared KMP logger module via OneSignal-KMP-SDK submodule

abdulraqeeb33:
## Summary
Introduces the multiplatform **`:logger`** module — the OpenTelemetry-free replacement for the `otel` module — and wires the Android SDK to consume it from the standalone [OneSignal-KMP-SDK](https://github.com/OneSignal/OneSignal-KMP-SDK) repo as a **git submodule**, so the exact same logging/telemetry pipeline can be shared with the iOS SDK and both release in parity.

This branch is up to date with `main` (latest merge brings in **SDK-4822**, the app-state-aware ANR watchdog + fatal/non-fatal crash indicator), and that work is now ported into the new logger path so it stays at parity with `otel`.

### What's included
- **Shared module consumption**: `OneSignal-KMP-SDK` added as a git submodule; `settings.gradle` remaps `:OneSignal:logger` to the submodule path (project path + dependency substitution unchanged). In-repo logger sources removed.
- **Wire format**: the KMP logger sends **OTLP/protobuf** (custom dependency-free encoder) and matches the old `otel` batch sizes, `LogLimits` (128 attrs / 32000 chars), `service.name=OneSignalDeviceSDK` / scope `loggerBuilder`, and background-thread I/O.
- **Android platform adapters** (host implements the injected interfaces): `AndroidLogger`, `FileLogStore`, `OneSignalLogHttpSender`, `AndroidLogCrashHandler`, `AndroidLogAnrDetector`, `LoggerPlatformProviderAdapter`, composed via `LoggerPlatformFactory` / `LoggerLifecycleManager`.
- **Crash paths at parity with `otel`**:
  - general `log` (level-driven, carries `exception.*`),
  - **fatal** `saveCrash` (Severity.FATAL) for real crashes and foreground ANRs,
  - **non-fatal** `saveNonFatal` (Severity.WARN) for backgrounded main-thread blocks,
  - an explicit, typed **`ossdk.crash.fatal`** OTLP bool attribute so the backend segments crash/ANR metrics on a stable flag rather than inferring intent from severity/exception type.
[...]

PR #2690, opened 2026-07-23
chore: opt in to publishing shared KMP module (com.onesignal:kmp)

abdulraqeeb33:
## Summary

Small stacked PR (on top of #2688) that adds only the host opt-in flag for publishing the shared KMP module.

- Adds `onesignalPublishSharedKmp = true` (with an explanatory comment block) to `OneSignalSDK/gradle.properties`.
- This opts the Android SDK build in as the "host" that publishes the shared `:kmp` module (sourced from the `OneSignal-KMP-SDK` submodule) to Maven Central, so the `com.onesignal:kmp` coordinate referenced by `:core`'s POM resolves for consumers.
- The shared module's `build.gradle` only wires up Maven publishing when this flag is set; the standalone KMP repo build and the iOS host (which ships an XCFramework) never set it.
- Intentionally does NOT include the KMP submodule wiring or `settings.gradle` module remap — those live on the integration branch.

## Test plan

- [ ] CI passes on the stacked branch.
- [ ] Confirm the `onesignalPublishSharedKmp` property is picked up during a release build so the `:kmp` module's Maven publishing tasks are registered.


fadi-george (approved):

fadi-george (changes requested):
rebase branches properly as theres duplicate changes

abdulraqeeb33:
Closing as redundant. The KMP publishing opt-in (com.onesignal:kmp) already landed on `main` via #2687, and the feature-flag work lives in #2688. This branch is based on the pre-merge state and would actually regress `main` (submodule pin 9cf1a33 → would revert aae38ac/v0.1.0, and `:kmp` → back to `:logger`).

PR #2691, opened 2026-07-24
feat(logger): [SDK-4939] wire Android FileLogStore for shared KMP legacy purge

abdulraqeeb33:
## Summary

Linear: [SDK-4939](https://linear.app/onesignal/issue/SDK-4939/harden-legacy-crash-file-cleanup-once-logger-module-is-active)

The KMP `logger` module and the legacy `otel` module share one on-disk crash directory (`{cacheDir}/onesignal/otel/crashes`). Ownership is distinguished by suffix — logger writes `*.otlp`; legacy otel writes bare-millis filenames plus stray `.tmp`s.

**Purge orchestration lives in shared KMP** ([OneSignal-KMP-SDK#8](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/8) — `LogCrashUploader` → `ILogFileStore.deleteUnrecognizedEntries(minAgeMillis)`). This PR provides the Android store implementation and rollout inventory logs.

## Changes

- **Submodule** — pins `OneSignal-KMP-SDK` to KMP #8 (post-upload / `finally` purge, age-gated foreign deletes, cancellation-safe).
- **`FileLogStore.deleteUnrecognizedEntries(minAgeMillis)`** — Android implementation; deletes non-`.otlp` entries at least `minAgeMillis` old; preserves owned `.otlp` and too-young foreign files.
- **`OneSignalCrashUploaderWrapper`** — before/after crash-dir inventory when the logger module is active (no host-side purge call).
- **`FileLogStoreTest`** — stale purge / young preserved / keep-owned / missing-dir / idempotency.

## Merge order

1. Land [OneSignal-KMP-SDK#8](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/8) first.
2. Merge this PR (re-pin submodule to the merged KMP SHA if the tip moves).

## Test plan

- [x] `:OneSignal:core:compileDebugKotlin` (after submodule pin)
- [x] KMP `:kmp:testDebugUnitTest --tests "*LogCrashTest*"`
- [ ] `:OneSignal:core:testDebugUnitTest --tests "*FileLogStoreTest*"`
- [ ] Manual: with `SDK_CUSTOM_LOGGING` on, seed a stale legacy file in the crash dir, confirm KMP purge removes it and after-cleanup inventory logs `legacy=0`

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/OneSignalCrashUploaderWrapper.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/OneSignalCrashUploaderWrapper.kt
@@ -7,9 +7,11 @@ import com.onesignal.core.internal.startup.IStartableService
 import com.onesignal.debug.internal.logging.Logging
 import com.onesignal.debug.internal.logging.logger.LoggerModuleSwitch
 import com.onesignal.debug.internal.logging.logger.android.AndroidLogger
+import com.onesignal.debug.internal.logging.logger.android.CrashDirEntry
 import com.onesignal.debug.internal.logging.logger.android.FileLogStore
 import com.onesignal.debug.internal.logging.logger.android.OneSignalLogHttpSender
 import com.onesignal.debug.internal.logging.logger.android.createAndroidLoggerPlatformProvider
+import com.onesignal.debug.internal.logging.logger.android.formatCrashDirInventory
 import com.onesignal.debug.internal.logging.otel.android.AndroidOtelLogger
 import com.onesignal.debug.internal.logging.otel.android.createAndroidOtelPlatformProvider
 import com.onesignal.logger.LoggerFactory
@@ -74,8 +76,9 @@ internal class OneSignalCrashUploaderWrapper(
                 Logging.info("OneSignal: Crash uploader selecting module=$module (SDK_CUSTOM_LOGGING=$useLogger)")
                 logCrashDirInventory("before-upload")
                 if (useLogger) {
-                    // Shared LogCrashUploader owns the legacy-file purge via
-                    // ILogFileStore.deleteUnrecognizedEntries() (FileLogStore).
+                    // Shared LogCrashUploader.start() is suspend and finishes the owned-record
+                    // upload pass plus the finally-purge before returning, so the after-cleanup
+                    // inventory below is not racing a background purge.
                     loggerUploader.start()
                     logCrashDirInventory("after-cleanup")
                 } else {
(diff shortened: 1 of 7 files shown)


abdulraqeeb33:
Submodule re-pinned to KMP `e636c2b` (exception-safe purge `finally` from KMP #8 review follow-up).

fadi-george:
A few items to address before merging:

1. CI currently fails Spotless, so the new tests have not run. Please apply the import-order fix and confirm `FileLogStoreTest` passes.
2. The submodule is pinned to PR-head SHA `3ececce7`, but KMP #8 was squash-merged as `4864ae1b4265baa433ce60a3588aa153689341ff`. Please re-pin to the merged SHA.
3. Avoid `Logging.info()` inside crash-path `FileLogStore.save()`. `Logging` synchronously invokes app listeners, and a listener exception can make a successfully persisted record return `false`. Raw Logcat would be safer there.
4. Please rethrow `CancellationException` from the Android wrapper and the touched suspend store paths instead of swallowing it through `Throwable`.
5. Bound the crash-directory inventory sample. It currently builds an unlimited filename summary before starting the uploader, and Logcat will truncate large output anyway.

The core purge classification and age-gating otherwise look good.

abdulraqeeb33:
Addressed in the latest push:

1. **Spotless** — import order fixed; `:OneSignal:core:spotlessKotlinCheck` + `FileLogStoreTest` pass locally.
2. **Submodule pin** — re-pinned to merged KMP `v0.1.1` / `4864ae1b4265baa433ce60a3588aa153689341ff`.
3. **Crash-path `save()`** — uses raw `android.util.Log` instead of `Logging.info`/`warn` so app listeners cannot flip a successful write to `false`.
4. **Cancellation** — `CancellationException` is rethrown from the uploader wrapper and from `listReadable` / `delete` / `deleteUnrecognizedEntries`.
5. **Inventory bound** — per-file sample capped at 20 entries with an `…(+N more)` suffix.

fadi-george (approved):
fix failing build/coverage checks too?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/crash/OneSignalCrashUploaderWrapper.kt
```diff
@@ -66,17 +69,73 @@ internal class OneSignalCrashUploaderWrapper(
...
-                if (LoggerModuleSwitch.useLoggerModule(applicationService.appContext)) {
+                val useLogger = LoggerModuleSwitch.useLoggerModule(applicationService.appContext)
+                val module = if (useLogger) "logger" else "otel"
+                Logging.info("OneSignal: Crash uploader selecting module=$module (SDK_CUSTOM_LOGGING=$useLogger)")
+                logCrashDirInventory("before-upload")
+                if (useLogger) {
+                    // Shared LogCrashUploader owns the legacy-file purge via
+                    // ILogFileStore.deleteUnrecognizedEntries() (FileLogStore).
                     loggerUploader.start()
+                    logCrashDirInventory("after-cleanup")
```
sherwinski, line 83: `logCrashDirInventory("after-cleanup")` runs immediately after `loggerUploader.start()` returns. 

The snapshot would only be accurate if `start()` is a suspend function that completes the full upload-and-purge pass before returning. I just want to confirm that these do not run in their parallel in case there's a race condition risk.

fadi-george: Confirmed — no race.

`LogCrashUploader.start()` is a `suspend` function. It runs the owned-record upload pass and then the unrecognized-file purge in a `finally` before returning, so `logCrashDirInventory("after-cleanup")` only runs after that full pass completes on the same coroutine.

Added a clarifying comment at the call site in the latest push.

fadi-george:
Rebased onto latest `main` and addressed the newest review notes:

1. **Suspend / race** — replied on the thread and added a call-site comment: `LogCrashUploader.start()` is suspend and finishes upload + `finally` purge before `after-cleanup` inventory runs.
2. **Coverage** — Robolectric Kotest shells are not attributed by Jacoco here (same constraint documented for `AnrCheckEvaluator`). Extracted ownership/inventory helpers to pure-JVM `CrashDirCleanup` + `CrashDirCleanupTest`, and applied the `Skip Coverage Check` label for the remaining Android shell lines.

sherwinski (approved):

PR #2692, opened 2026-07-24
feat(logger): [SDK-4940] log release-diagnostic info at SDK startup

abdulraqeeb33:
## Summary

Linear: [SDK-4940](https://linear.app/onesignal/issue/SDK-4940/log-more-release-diagnostic-info-at-sdk-startup-sdkkmp-version-active)

During the otel-to-logger rollout it's hard to tell, from a device log capture alone, exactly what a device is running and which observability module is active. Today the SDK version and KMP module version are only sent on the wire / in headers, never logged locally.

This PR adds a single, concise INFO diagnostic line at init and makes the otel crash-upload path equally observable.

## Changes

- **`OneSignalImp.initEssentials`** — one INFO line (`logStartupDiagnostics`) with: `sdkVersion`, active `observabilityModule` + `SDK_CUSTOM_LOGGING` value, `kmpVersion` (`LoggerBuildInfo.KMP_VERSION`, shown only when the logger module is active), OS/API level, device manufacturer/model, host app package + `versionName`, and the crash storage dir. Best-effort, wrapped in try/catch so diagnostics never interfere with init.
- **`OtelCrashUploader`** — upgraded logging: path/level on start, before/after disk inventories, per-batch record count + short preview, and a pass summary. Logging-only.

## Notes

- No PII: only build/version/module/device-model/app-package facts (device model already ships in telemetry).
- One line at INFO; strings built only when the diagnostic runs.

## Test plan

- [x] `:OneSignal:core:compileDebugKotlin` + `:OneSignal:otel:compileDebugKotlin`
- [x] `:OneSignal:core:detekt` + `:OneSignal:otel:detekt` (no new findings)
- [x] `:OneSignal:core:spotlessKotlinCheck` + `:OneSignal:otel:spotlessKotlinCheck`
- [x] `:OneSignal:otel:testDebugUnitTest`
- [ ] Manual: cold start with logger active → confirm `OneSignal init: sdkVersion=... observabilityModule=logger ... kmpVersion=...` appears once in logcat

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
@@ -1,6 +1,7 @@
 package com.onesignal.internal
 
 import android.content.Context
+import android.os.Build
 import com.onesignal.IOneSignal
 import com.onesignal.IUserJwtInvalidatedListener
 import com.onesignal.common.AndroidUtils
@@ -240,8 +241,11 @@ internal class OneSignalImp : IOneSignal,
         // lazy supplier — `enabledFeatureFlags` is read per-event, so resolving the manager
         // can be deferred until services have bootstrapped.
         val featureManagerProvider = { services.getService<IFeatureManager>() }
+        val useLoggerModule =
+            com.onesignal.debug.internal.logging.logger.LoggerModuleSwitch.useLoggerModule(context)
+        logStartupDiagnostics(context, useLoggerModule)
         observabilityManager =
-            if (com.onesignal.debug.internal.logging.logger.LoggerModuleSwitch.useLoggerModule(context)) {
+            if (useLoggerModule) {
                 LoggerLifecycleManager(context = context, featureManagerProvider = featureManagerProvider)
             } else {
                 OtelLifecycleManager(context = context, featureManagerProvider = featureManagerProvider)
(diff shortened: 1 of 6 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -252,6 +256,38 @@ internal class OneSignalImp : IOneSignal,
...
+        try {
+            val module = if (useLoggerModule) "logger" else "otel"
+            val kmpVersion = if (useLoggerModule) com.onesignal.logger.LoggerBuildInfo.KMP_VERSION else "n/a"
+            val appVersion =
+                try {
+                    context.packageManager.getPackageInfo(context.packageName, 0).versionName
+                } catch (t: Throwable) {
+                    "unknown"
+                }
+            Logging.info(
```
sherwinski: This log level is set to `INFO` so this would be suppressed unless the implementing developer specifically allows it. Is that intentional?
Since it looks like the purpose here is to `release triage from a raw log capture` that might call for a more visible debug level. Let me know what you think.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -240,8 +241,11 @@ internal class OneSignalImp : IOneSignal,
         // lazy supplier — `enabledFeatureFlags` is read per-event, so resolving the manager
         // can be deferred until services have bootstrapped.
         val featureManagerProvider = { services.getService<IFeatureManager>() }
+        val useLoggerModule =
+            com.onesignal.debug.internal.logging.logger.LoggerModuleSwitch.useLoggerModule(context)
+        logStartupDiagnostics(context, useLoggerModule)
```
sherwinski: This startup line is emitted before remote telemetry is wired, so it would only log locally.
`logStartupDiagnostics` runs before `observabilityManager` is constructed / `initializeFromCachedConfig()` (line 247-252), which is what calls `setOtelTelemetry`/`setLoggerTelemetry`. At L246, `shouldSendLogLevel` is still the default `{ false }`, so this line it only reaches logcat and not the remote logger.
If we wanted it in remote captures, it should run after telemetry is set (or be re-emitted).

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -252,6 +256,38 @@ internal class OneSignalImp : IOneSignal,
...
+                    "unknown"
+                }
+            Logging.info(
+                "OneSignal init: sdkVersion=${OneSignalUtils.sdkVersion} " +
+                    "observabilityModule=$module (SDK_CUSTOM_LOGGING=$useLoggerModule) " +
+                    "kmpVersion=$kmpVersion " +
+                    "os=Android/${Build.VERSION.RELEASE}(API ${Build.VERSION.SDK_INT}) " +
+                    "device=${Build.MANUFACTURER}/${Build.MODEL} " +
+                    "app=${context.packageName}@$appVersion " +
+                    "crashDir=${context.cacheDir.path}/onesignal/otel/crashes",
```
sherwinski: `crashDir` is a hand-rebuilt path string that duplicates the real construction.

The actual path is built in `createAndroidOtelPlatformProvider` (OtelPlatformProvider.kt:175-178) as `cacheDir/onesignal/otel/crashes`. There's a risk of these strings diverging in the future so we should consider extracting the path into a shared constant both sites use.

sherwinski (changes requested):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
fadi-george: Extracted getOtelCrashStoragePath(context) and now use it from both the platform provider and startup diagnostic.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
fadi-george: Updated the startup diagnostic to WARN so it appears with the default local log level.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
fadi-george: Moved the diagnostic after initializeFromCachedConfig(), so the remote telemetry sink is wired before it is emitted.

sherwinski (approved):

PR #2700, opened 2026-07-29
fix: [SDK-4946] register with Firebase Installation ID when the legacy FCM token API is disabled

fadi-george:
# Description
## One Line Summary
Fall back to Firebase Installation ID registration when an app has disabled the legacy FCM token API, instead of failing to register for push.

### Motivation
Firebase Messaging 25.1.0 deprecated `getToken()`, `deleteToken()`, and `onNewToken()` in favor of Firebase Installation ID (FID) registration. An app opts in by adding `firebase_messaging_installation_id_enabled` to its manifest, and that flag disables the legacy token API for the entire process, not just the `FirebaseApp` that opted in.

That means OneSignal's own `getToken()` call fails with `IllegalStateException: API disabled`, and the device never registers for push. Apps can hit this without doing anything deliberate, since any dependency that sets the flag turns it on for everyone.

### Scope
Only affects the FCM registration path when the legacy token API has been disabled. Apps that have not opted in keep using `getToken()` exactly as before, so there is no behavior change for the vast majority of integrations.

### Apps that set the flag need their own google-services.json
Worth calling out because it is a new requirement, though only for apps that opt in.

Setting `firebase_messaging_installation_id_enabled` to `true` makes FID registration the only way to register. A FID is issued by a Firebase project and is only meaningful within it, and on Android it comes from the app's default `FirebaseApp`, which is initialized from `google-services.json` by the google-services Gradle plugin. OneSignal's own `FirebaseApp` cannot issue a usable one, for the credential reasons described below.

So an app that sets the flag has to ship a `google-services.json` for its own Firebase project, and that project's sender id has to match the sender id configured in the OneSignal dashboard. If either piece is missing, this PR fails registration with an error that says exactly which one.

Apps that leave the flag alone, which is the default, still need nothing beyond their sender id. Nothing here asks existing integrations to add Firebase configuration.

### Why registration runs against the default FirebaseApp
FID registration requires the sender id, app id, and api key to all belong to a single Firebase project. Our `FirebaseApp` deliberately pairs the app's sender id with OneSignal's shared project credentials, so calling `register()` on it is rejected by Play Services with an opaque `ApiException: 8: Internal error encountered.`

When the legacy API is disabled we therefore look for the host app's default `FirebaseApp`, verify its sender id matches the one OneSignal is configured with, and register through it. If there is no default app, or the sender ids disagree, we throw an error that names the problem and how to fix it rather than letting the opaque Play Services error surface.

`FirebaseMessaging.register()` is called reflectively because it was added in firebase-messaging 25.1.0, which is newer than the version this module compiles against.

### Also in this PR: JaCoCo now records Robolectric coverage
The diff coverage gate reported 0% on `PushRegistratorFCM.kt` even though the tests exercise it and pass. JaCoCo was recording nothing from Robolectric tests, because Robolectric loads classes through its own classloader and JaCoCo skips classes with no code source location by default. Every file with recorded coverage in this module was tested by a non-Robolectric test, and Robolectric-tested classes like `NotificationChannelManager` sat at exactly 0.

`coverage/jacoco.gradle` now sets `includeNoLocationClasses` on unit test tasks. These tests have to be Robolectric, since `Tasks.await()` needs `Looper.getMainLooper()` and the `PushRegistratorFCM` constructor decodes the default api key with `android.util.Base64`.

Worth knowing that this changes reported coverage repo wide, not just here. The notifications module goes from 11.4% to 30.3% with no new tests, because a large share of this repo's tests are Robolectric and none of them were being counted. Existing coverage numbers were understated rather than this PR inflating them. Happy to split this into its own PR if reviewers would rather keep the fix isolated.

# Testing
## Unit testing
`FCMTokenProviderTests` covers ten paths through the new logic: the legacy token when the API is enabled, successful FID registration when it is disabled, unrelated `IllegalStateException`s passing through untouched, the missing default app and sender id mismatch diagnostics along with the resolved manifest flag value they report, registration failures propagating, and the three reflection behaviors (invoking `register()`, unwrapping what it throws rather than surfacing the wrapper, and explaining a missing method). `PushRegistratorFCMTests` covers the same wiring through Robolectric against real manifest metadata.

Coverage on the changed lines is 100%, 53 of 53.

## Manual testing
Tested on an Android 16 emulator using the `fadi/SDK-4946-firebase-25-demo` branch. That branch is this fix with the reproduction setup layered on top, since the bug cannot be triggered with the example app as it ships today. It pins the example app to firebase-messaging 25.1.1, sets `firebase_messaging_installation_id_enabled` to `true` in the gms manifest, and applies the google-services plugin so the app has a default `FirebaseApp` to register against. It is a test harness and is not meant to merge.

`google-services.json` is gitignored there since it is specific to one Firebase project, so drop in your own to reproduce the opt-in path, using a project whose sender id matches the one in your OneSignal dashboard. Without the file the plugin is skipped and gms builds still work, they just take the legacy token path.

With the flag on:

- Without this fix, registration fails with `IllegalStateException: API disabled`.
- With this fix and `google-services.json` in place, the app registers and receives notifications sent from the dashboard. The token is a FID, for example `ctmBeCwtQDuDsClbyr6d3z`.
- With this fix and `google-services.json` removed, the new "no default FirebaseApp" error appears instead of a silent failure.

With the flag off, which is the default and covers every existing integration, both on firebase-messaging 25.1.1:

- With `google-services.json` present, registration returns a normal FCM token, for example `fWAQuU2QRcyO-MCoWP5l4D:APA91bF...`. The new code path is never entered.
- With no `google-services.json` and the google-services plugin not applied, which is what a typical OneSignal app looks like, registration also returns a normal FCM token. `FirebaseInitProvider` logs that there is no default app, and it does not matter because nothing asks for one.

I also confirmed the default-app redirect is the part that makes the opt-in case work. Reverting just `PushRegistratorFCM.kt` to an earlier version that called `register()` on OneSignal's own `FirebaseApp`, while leaving a valid `google-services.json` in place, still fails with `ApiException: 8`.

### Firebase version and minification
Ran the flag both ways on firebase-messaging 24.0.0 as well, since that is where most apps sit today and `register()` does not exist in the class there. Legacy token both times, which is expected because the flag does nothing before 25.1.0.

Both flag states also run in a minified release build, which is what exercises the new consumer ProGuard rule. Behavior is the same as debug and there is no `NoSuchMethodException`.

### Token conversions on an existing install
Flipping the flag on an install that already holds a token, without clearing app data:

- Legacy token to FID, the direction an app takes when it adopts the flag: the subscription id is preserved, the token swaps from `dN8Ya5MTR2OO-0BXdqimCI:APA91bFw...` to the FID `cXp9fY-TRFCLSy9kVTqorZ`, and a push sent to that subscription still arrives.
- FID back to legacy token: works when the rollback ships as an app version bump, which is how it would actually reach users. FCM issues a fresh token, the subscription id is unchanged, and delivery works.

Worth knowing before testing the rollback locally: if you flip the flag off by rebuilding at the same `versionCode`, `getToken()` hands back Firebase's cached pre-FID token, which FCM has already invalidated. The backend then gets an unregistered response and parks the subscription at `notification_types: -10`, which our own enum documents as "server side detection only from FCM that the app is no longer installed", so push stops silently while the SDK keeps re-asserting the same dead token. Bumping `versionCode` or clearing app data fixes it. This PR does not change the flag-off path, we return whatever Firebase gives us, but it looks like a bug in the fix if you hit it.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
@@ -1,9 +1,11 @@
 package com.onesignal.notifications.internal.registration.impl
 
 import android.util.Base64
+import com.google.android.gms.tasks.Task
 import com.google.android.gms.tasks.Tasks
 import com.google.firebase.FirebaseApp
 import com.google.firebase.FirebaseOptions
+import com.google.firebase.installations.FirebaseInstallations
 import com.google.firebase.messaging.FirebaseMessaging
 import com.onesignal.core.internal.application.IApplicationService
 import com.onesignal.core.internal.config.ConfigModelStore
@@ -49,22 +51,40 @@ internal class PushRegistratorFCM(
     @Throws(ExecutionException::class, InterruptedException::class)
     override suspend fun getToken(senderId: String): String {
         initFirebaseApp(senderId)
-        return getTokenWithClassFirebaseMessaging()
+        return getTokenWithClassFirebaseMessaging(senderId)
     }
 
     @Throws(ExecutionException::class, InterruptedException::class)
-    private fun getTokenWithClassFirebaseMessaging(): String {
+    private fun getTokenWithClassFirebaseMessaging(senderId: String): String {
         // We use firebaseApp.get(FirebaseMessaging.class) instead of FirebaseMessaging.getInstance()
         //   as the latter uses the default Firebase app. We need to use a custom Firebase app as
         //   the senderId is provided at runtime.
         val fcmInstance = firebaseApp!!.get(FirebaseMessaging::class.java)
-        // FirebaseMessaging.getToken API was introduced in firebase-messaging:21.0.0
-        val tokenTask = fcmInstance.token
-        try {
-            return Tasks.await(tokenTask)
-        } catch (e: ExecutionException) {
-            throw tokenTask.exception ?: e
-        }
+        return FCMTokenProvider.getToken(senderId, { fcmInstance.token }, ::defaultAppRegistration)
+    }
+
+    // Installation ID registration is rejected unless the sender id, app id, and api key all belong
+    //   to the same Firebase project. Our own FirebaseApp pairs the app's sender id with OneSignal's
+    //   shared project credentials, so only the host app's default FirebaseApp can be used for it.
+    private fun defaultAppRegistration(): FCMTokenProvider.InstallationIdRegistration? {
+        val defaultApp =
+            FirebaseApp
+                .getApps(_applicationService.appContext)
+                .firstOrNull { it.name == FirebaseApp.DEFAULT_APP_NAME } ?: return null
+
+        return FCMTokenProvider.InstallationIdRegistration(
+            senderId = defaultApp.options.gcmSenderId,
+            register = { register(defaultApp.get(FirebaseMessaging::class.java)) },
+            installationId = { FirebaseInstallations.getInstance(defaultApp).id },
+        )
+    }
+
+    // FirebaseMessaging.register was added in firebase-messaging:25.1.0, which is newer than the
+    //   version this module compiles against.
+    private fun register(firebaseMessaging: FirebaseMessaging): Task<*> {
+        return firebaseMessaging.javaClass
+            .getMethod("register")
+            .invoke(firebaseMessaging) as Task<*>
     }
 
     private fun initFirebaseApp(senderId: String) {
@@ -78,5 +98,76 @@ internal class PushRegistratorFCM(
                 .setProjectId(projectId)
                 .build()
         firebaseApp = FirebaseApp.initializeApp(_applicationService.appContext, firebaseOptions, FCM_APP_NAME)
+    }
+}
+
+internal object FCMTokenProvider {
+    /**
+     * The Firebase Installation ID registration that replaces the legacy token API, along with the
+     * sender id of the Firebase project it would register against.
+     */
+    class InstallationIdRegistration(
+        val senderId: String?,
+        val register: () -> Task<*>,
+        val installationId: () -> Task<String>,
+    )
+
+    /**
+     * Retrieves an FCM token for [senderId], falling back to Firebase Installation ID registration
+     * when the host app has opted into it. Opting in disables the legacy token API for the whole
+     * app, not just the FirebaseApp that opted in.
+     */
+    fun getToken(
+        senderId: String,
+        legacyToken: () -> Task<String>,
+        installationIdRegistration: () -> InstallationIdRegistration?,
+    ): String {
+        return try {
+            await(legacyToken())
+        } catch (e: IllegalStateException) {
+            if (!isLegacyTokenApiDisabled(e)) throw e
+
+            registerInstallationId(senderId, installationIdRegistration())
+        }
+    }
+
+    private fun registerInstallationId(
+        senderId: String,
+        registration: InstallationIdRegistration?,
+    ): String {
+        if (registration == null) {
+            throw IllegalStateException(
+                "Firebase Installation ID registration is enabled but this app has no default " +
+                    "FirebaseApp to register with. Add your Firebase configuration " +
+                    "(google-services.json), or remove firebase_messaging_installation_id_enabled " +
+                    "from your manifest to keep using the legacy FCM token API.",
+            )
+        }
+
+        if (registration.senderId != senderId) {
+            throw IllegalStateException(
+                "Firebase Installation ID registration is enabled but the default FirebaseApp uses " +
+                    "sender id ${registration.senderId}, while OneSignal is configured with sender " +
+                    "id $senderId. Point both at the same Firebase project, or remove " +
+                    "firebase_messaging_installation_id_enabled from your manifest to keep using " +
+                    "the legacy FCM token API.",
+            )
+        }
+
+        await(registration.register())
+        return await(registration.installationId())
+    }
+
+    private fun isLegacyTokenApiDisabled(exception: IllegalStateException): Boolean {
+        val message = exception.message ?: return false
+        return message.contains("API disabled") && message.contains("register()")
+    }
+
+    private fun <T> await(task: Task<T>): T {
+        try {
+            return Tasks.await(task)
+        } catch (e: ExecutionException) {
+            throw task.exception ?: e
+        }
     }
 }
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
@@ -79,8 +79,9 @@ internal class PushRegistratorFCM(
         )
     }
 
-    // FirebaseMessaging.register was added in firebase-messaging:25.1.0, which is newer than the
-    //   version this module compiles against.
+    // FirebaseMessaging.register was added in firebase-messaging:25.1.0. This module compiles
+    //   against the preferred 24.0.0, but the non-strict Gradle constraint lets apps select newer
+    //   versions through conflict resolution.
     private fun register(firebaseMessaging: FirebaseMessaging): Task<*> {
         return firebaseMessaging.javaClass
             .getMethod("register")
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
@@ -156,7 +156,6 @@ internal object FCMTokenProvider {
      * 25.1.0. This module compiles against the preferred 24.0.0, but the non-strict Gradle
      * constraint lets apps select newer versions through conflict resolution.
      */
-    @Suppress("SwallowedException")
     fun invokeRegister(target: Any): Task<*> {
         val register =
             try {
@@ -175,7 +174,7 @@ internal object FCMTokenProvider {
         return try {
             register.invoke(target) as Task<*>
         } catch (e: InvocationTargetException) {
-            throw e.targetException
+            throw e.targetException ?: e
         }
     }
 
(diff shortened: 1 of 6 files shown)


fadi-george:
@claude review

sherwinski (changes requested):
According to [this comment](https://github.com/OneSignal/OneSignal-Android-SDK/pull/2700/changes#diff-821db332b0f71d8aa0a877c2451201c38b3a5aac833e653356bb35af0b611cfaR82) the `FirebaseMessaging.register` method was added in v25.1.0 but [this package](https://github.com/OneSignal/OneSignal-Android-SDK/blob/a9e2a365a42dbe430ce6645b9d76a1f9fb46bd3c/OneSignalSDK/onesignal/notifications/build.gradle#L79-L84) only resolves against `[23.0.8, 24.0.99]`. So the situation we're guarding against in this PR wouldn't occur as this package won't resolve against firebase-messaging:25.1.0

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -49,22 +51,40 @@ internal class PushRegistratorFCM(
...
+        )
+    }
+
+    // FirebaseMessaging.register was added in firebase-messaging:25.1.0, which is newer than the
+    //   version this module compiles against.
+    private fun register(firebaseMessaging: FirebaseMessaging): Task<*> {
+        return firebaseMessaging.javaClass
+            .getMethod("register")
+            .invoke(firebaseMessaging) as Task<*>
     }
```
sherwinski: `consumer-rules.pro` documents a keep rule for every other reflection target in this module (INotificationServiceExtension, Badger, WorkManager workers, InputMerger). `FirebaseMessaging.register()` is never referenced symbolically by the SDK or, typically, by the host app under R8 full mode (AGP 8+) it can be renamed or stripped. The result is a `NoSuchMethodException` that falls into the catch-all at PushRegistratorAbstractGoogle.kt:159 and surfaces as `"Unknown error getting FCM Token"` / `FIREBASE_FCM_ERROR_MISC_EXCEPTION`. We should add:

```kotlin
-keepclassmembers class com.google.firebase.messaging.FirebaseMessaging {
    public *** register();
}
```

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -49,22 +51,40 @@ internal class PushRegistratorFCM(
...
+            installationId = { FirebaseInstallations.getInstance(defaultApp).id },
+        )
+    }
+
+    // FirebaseMessaging.register was added in firebase-messaging:25.1.0, which is newer than the
+    //   version this module compiles against.
+    private fun register(firebaseMessaging: FirebaseMessaging): Task<*> {
+        return firebaseMessaging.javaClass
+            .getMethod("register")
+            .invoke(firebaseMessaging) as Task<*>
```
sherwinski: `invoke()` wraps any synchronous throw from `register()` in `InvocationTargetException`, so the real cause is buried under a reflection frame, and getMethod failure produces a `NoSuchMethodException` with no hint about firebase-messaging versions.
Catch `InvocationTargetException` and rethrow `e.targetException`, and catch `NoSuchMethodException` with a message naming the required version.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -80,3 +100,74 @@ internal class PushRegistratorFCM(
...
+
+        if (registration.senderId != senderId) {
+            throw IllegalStateException(
+                "Firebase Installation ID registration is enabled but the default FirebaseApp uses " +
+                    "sender id ${registration.senderId}, while OneSignal is configured with sender " +
+                    "id $senderId. Point both at the same Firebase project, or remove " +
+                    "firebase_messaging_installation_id_enabled from your manifest to keep using " +
+                    "the legacy FCM token API.",
+            )
+        }
```
sherwinski, line 168: Consider adding more signal to this message, it could both help the user and us when debugging. For example, we could read and display the value for `firebase_messaging_installation_id_enabled` here since we're already calling it out.

fadi-george:
The range uses `require`, not `strictly`, so the upper bound doesn't cap anything. Per Gradle's [rich version docs](https://docs.gradle.org/8.10/userguide/rich_versions.html), a required version can go higher through conflict resolution even past an exclusive upper bound, so we compile against the preferred 24.0.0 but an app that asks for 25.1.x gets it. That's how I reproduced this: the demo branch declares `firebase-messaging:25.1.1`, the merged manifest confirms 25.1.1 resolved, and registration fails on device with `IllegalStateException: API disabled` once the flag is on. Widening the range wouldn't change anything since `prefer '24.0.0'` still wins by default, and bumping `prefer` to 25.1.x would raise our minSdk from 21 to 23, so that's a separate call.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: Good catch, and it was worse than theoretical. I reproduced it with the demo release build, which has `isMinifyEnabled = true` specifically to exercise R8.

Dumping the dex of `app-gms-release.apk` before the rule, every `FirebaseMessaging` method was obfuscated and `register` was gone:

```
names kept: [<clinit>, <init>, getInstance]
```

So `getMethod("register")` would have thrown `NoSuchMethodException` in any minified release build, exactly as you described. `getToken` gets renamed too, but our call site is a direct invocation that R8 rewrites, so only the reflective lookup breaks.

Also confirmed nothing upstream protects it: the 25.1.1 AAR ships no `proguard.txt`, and `register()` carries no `@Keep`.

Added your rule in 4d83eff7e, using `keepclassmembers` so Huawei apps that exclude firebase-messaging are unaffected. Same dump after:

```
names kept: [<clinit>, <init>, getInstance, register]
```

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: Done in bb1028762. `InvocationTargetException` now rethrows `targetException`, and `NoSuchMethodException` becomes an `IllegalStateException` naming 25.1.0 and pointing at the ProGuard rules, since after your other comment that is the likelier cause.

Moved the reflection into `FCMTokenProvider.invokeRegister` so it takes `Any` and is testable with a plain fake. Three tests added, and changed-line coverage went to 100%.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: Added in cb5fcd867. Both messages now include the resolved value, for example `firebase_messaging_installation_id_enabled=not set`.

Read as a raw meta-data value rather than a boolean, since a string `"true"` reads as false via `getBoolean`. Reading it at runtime also catches the case where the flag was merged in from a dependency instead of the app own manifest, which is the confusing one to debug.

For that same reason I changed the advice from "remove it from your manifest" to "set it to false", which works even when the app never declared it.

`FCMTokenProvider` stays free of Android deps by taking the value as a lambda, so it is only read on the fallback path.

fadi-george:
Yes, both, on an Android 16 emulator using `fadi/SDK-4946-firebase-25-demo` at the current tip. That branch lets me flip the manifest flag with `-PfidRegistrationEnabled`, so I ran each case in debug and in minified release.

Legacy FCM token API not disabled (flag `false`): normal FCM token, for example `fgHF9nQnSxWWBnuug0FM1y:APA91bE...`. The new code path is never entered.

Legacy FCM token API disabled (flag `true`): FID token, for example `eS8k9kXpSzuegDMN6S5PkF`. The Push ID shows up in the app and a notification sent from the demo is received, so the token works end to end.

Minified release behaves the same in both cases, which is the part that exercises the new keep rule. No `NoSuchMethodException`.

I also ran firebase-messaging 24.0.0 with the flag both ways, since that is where most apps sit today and `register()` does not exist in the class there. Legacy token both times, which is expected because the flag does nothing before 25.1.0.

One thing I noticed while testing: with the flag on, firebase-messaging logs `AsyncOperation: operation=Register, opStatusCode=8` and a topic sync failure just after we register. That is its background sync for OneSignal's own `FirebaseApp`, which cannot do FID registration. Registration and delivery are unaffected. It happens because we instantiate `FirebaseMessaging` on that app to try the legacy API first, so it is not new here, but I can short circuit on the manifest flag if you would rather avoid the noise.

fadi-george:
Also tested both conversions on an existing install, without clearing app data.

Legacy token to FID, which is the direction an app takes when it adopts the flag: the subscription id is preserved, the token swaps from `dN8Ya5MTR2OO-0BXdqimCI:APA91bFw...` to the FID `cXp9fY-TRFCLSy9kVTqorZ`, and a push sent to that subscription still arrives. Ran it twice from clean state to be sure.

FID back to legacy token: works when the rollback ships as an app version bump, which is how it would actually reach users. FCM issues a fresh token, the subscription id is unchanged, and delivery works.

One heads up if you test the rollback locally. If you flip the flag off by rebuilding at the same `versionCode`, `getToken()` hands back Firebase's cached pre-FID token, which FCM has already invalidated. The backend then gets an unregistered response and parks the subscription at `notification_types: -10`, which per our own enum is "server side detection only from FCM that the app is no longer installed", so push stops silently and the SDK keeps re-asserting the same dead token. Bumping `versionCode` or clearing app data fixes it. This is not something the PR causes, the flag-off path is unchanged and we return whatever Firebase gives us, but it looks like a bug in the fix if you hit it.

sherwinski (approved):

abdulraqeeb33:
Reverting this PR to conduct more testing - discuss with the group.

Issue #17, opened 2016-02-19
One Signal Android SDK getTags can't get all tags from server


pikachuShy:
When using the getTags function, the return json only return one value for every first time i open the app(uninstall and install). For example in the dashboard, i saw that i already consists three tags, after i called the getTags, it only return me the first tag. 
Is it a bug?

jkasten2:
Currently the getTags method only returns the device's local tags. We will look into changing this logic to fetch the tags the first time from the server.

kwiky:
This is an important feature for us.
How can we delete all server tags without knowing it on the device ?

jkasten2:
@kwiky The work around for now would be to use the OneSignal [View device](https://documentation.onesignal.com/docs/playersid) REST API call to get the current tags. Then you can use the [Edit device](https://documentation.onesignal.com/docs/playersid-1) REST API call to delete the list of tags (See the note on the bottom of the page of how to do this).

I'll shoot for getting a fix for this out in the next minor release in the next few weeks.

kwiky:
thanks @jkasten2, using the REST API is what we have done for now.
We will checkout this next minor release to use the fix instead.

jkasten2:
Fixed this issue in commit 21658df24d041fcdda11256964a5dfc17fdbf96d

PR #2713, opened 2026-08-10
feat: [SDK-4999] emit Kotlin version on remote log resource attrs

abdulraqeeb33:
## Summary
Android host follow-up for [OneSignal-KMP-SDK#17](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/17) ([SDK-4999](https://linear.app/onesignal/issue/SDK-4999)): supply `KotlinVersion.CURRENT` so remote logs include `ossdk.kotlin_version` for dashboard filtering.

## Changes
- **Submodule** — pins `OneSignal-KMP-SDK` to KMP #17 (`267b460`, `ar/sdk-4999`).
- **`OtelPlatformProvider`** — `kotlinVersion = KotlinVersion.CURRENT.toString()`.
- **`IOtelPlatformProvider` / `LoggerPlatformProviderAdapter`** — add abstract `kotlinVersion` / `swiftVersion` / `additionalVersionAttributes` (adapter forwards to the logger path).
- **`OtelFieldsTopLevel`** — mirrors KMP `LogFieldsTopLevel` (extras first, blank skip, `ossdk.` prefix normalize) so the legacy otel path stays at parity while both modules coexist.
- **Tests** — provider returns `KotlinVersion.CURRENT`; top-level attrs cover emit / blank omit / extras merge / reserved-key protection.
- **Docs** — mark `:otel` / `IOtelPlatformProvider` / `OtelFactory` as SDK-internal (not for public use).

## Compatibility note
Adding the new abstract members on `IOtelPlatformProvider` is a **deliberate source-compatibility break** for any external implementor of that interface. `:otel` is published and `OtelFactory` accepts the type, but this is an SDK-internal injection seam — the only production implementor is `OtelPlatformProvider` in `:core`, and there are no known external consumers. We are not adding default methods or a capability interface for this reason.

## Merge order
1. Land [OneSignal-KMP-SDK#17](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/17) first.
2. Merge this PR (re-pin submodule to the merged KMP SHA / release tag if the tip moves).

## Test plan
- [x] Code review vs KMP #17 contract
- [ ] `:OneSignal:core:testDebugUnitTest --tests "*OtelPlatformProviderTest*"`
- [ ] `:OneSignal:otel:testDebugUnitTest --tests "*OtelFieldsTopLevelTest*"`
- [ ] `:OneSignal:core:compileDebugKotlin` against pinned submodule
- [ ] After KMP merge: re-pin to merged SHA / tagged release via bump workflow when available

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/android/OtelPlatformProvider.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/android/OtelPlatformProvider.kt
@@ -68,18 +68,18 @@ internal class OtelPlatformProvider(
     // KotlinVersion.CURRENT reflects the stdlib the SDK was compiled against.
     override val kotlinVersion: String? = KotlinVersion.CURRENT.toString()
 
+    // Android has no Swift toolchain.
+    override val swiftVersion: String? = null
+
     // Extra toolchain labels under ossdk.* for dashboard filtering. Prefer values that are
-    // stable for the process lifetime and cheap to read:
-    // - java_version: ART's reported Java language level (spec version, e.g. "17")
+    // stable for the process lifetime and cheap to read.
     // - android_api_level: device API level (SDK_INT) — complementary to os.version (RELEASE)
-    // Build-time-only values (AGP, compileSdk, NDK) are not available here at runtime.
+    // Do NOT emit java_version from System.getProperty("java.specification.version"):
+    // on-device ART hardcodes that property to "0.9" (see AndroidHardcodedSystemProperties).
+    // Robolectric returns the host JDK version and would mask that in unit tests.
+    // Build-time-only values (AGP, compileSdk, NDK, real javac target) are unavailable at runtime.
     override val additionalVersionAttributes: Map<String, String> =
-        buildMap {
-            System.getProperty("java.specification.version")
-                ?.takeIf { it.isNotBlank() }
-                ?.let { put("java_version", it) }
-            put("android_api_level", Build.VERSION.SDK_INT.toString())
-        }
+        mapOf("android_api_level" to Build.VERSION.SDK_INT.toString())
 
     // Read through the supplier on every access so per-event attributes always reflect the
     // current featureStates snapshot (including IMMEDIATE-mode flag changes). The supplier is
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/android/OtelPlatformProvider.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/android/OtelPlatformProvider.kt
@@ -68,6 +68,19 @@ internal class OtelPlatformProvider(
     // KotlinVersion.CURRENT reflects the stdlib the SDK was compiled against.
     override val kotlinVersion: String? = KotlinVersion.CURRENT.toString()
 
+    // Extra toolchain labels under ossdk.* for dashboard filtering. Prefer values that are
+    // stable for the process lifetime and cheap to read:
+    // - java_version: ART's reported Java language level (spec version, e.g. "17")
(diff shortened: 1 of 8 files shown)


fadi-george (commented):
Two issues to address before merge:

- `java.specification.version` is hardcoded to `"0.9"` on Android, so `ossdk.java_version` would be misleading. Robolectric returns the host JDK version and masks this behavior.
- Default interface getters preserve source compatibility, but older compiled `IOtelPlatformProvider` implementations may still throw `AbstractMethodError`.

Also, please trim attribute keys before removing `ossdk.` to avoid emitting `ossdk.ossdk.*`.

fadi-george:
One compatibility concern remains: making these properties abstract removes source compatibility but does not prevent older compiled implementations from throwing `AbstractMethodError`.

The KDoc also says the interface and Android implementation share an artifact, but `IOtelPlatformProvider` is in `:otel` while `OtelPlatformProvider` is in `:core`. Please use an optional capability interface or explicitly treat this as a breaking API change.

abdulraqeeb33:
Agreed the KDoc was wrong on the same-artifact claim (`IOtelPlatformProvider` is in `:otel`, `OtelPlatformProvider` in `:core`).

We're deliberately **not** adding a capability interface or default methods here though: this isn't a public implementor surface. The only production implementor is our own `OtelPlatformProvider`, and app/wrapper authors never implement `IOtelPlatformProvider`. AbstractMethodError for "older compiled external implementors" isn't a real consumer risk for this type.

Updated the KDoc to say SDK-internal only and dropped the misleading binary-compat framing.

fadi-george:
I understand this is intended as SDK-internal, but it is still publicly exposed from the published :otel module, and public OtelFactory methods accept it. KDoc alone does not make it internal.

Adding abstract properties is therefore a source-breaking API change for any external implementor. If we accept that risk because there are no known consumers, please document it as a deliberate compatibility break and update the PR body, which still says defaults preserve compatibility.

abdulraqeeb33:
@fadi-george Good catch — agreed. Updated to document `:otel` / `IOtelPlatformProvider` / `OtelFactory` as SDK-internal (not for public implementors), explicitly called out the deliberate source-compat break for the new abstract members, and fixed the PR body so it no longer claims defaults preserve compatibility.

abdulraqeeb33:
> I understand this is intended as SDK-internal, but it is still publicly exposed from the published :otel module, and public OtelFactory methods accept it. KDoc alone does not make it internal.
> 
> Adding abstract properties is therefore a source-breaking API change for any external implementor. If we accept that risk because there are no known consumers, please document it as a deliberate compatibility break and update the PR body, which still says defaults preserve compatibility.

Discussed with @fadi-george  - the otel module is something that will always be consumed internally. We should deter external users to use it as we won't be supporting external usage of this module.

fadi-george:
Waiting on KMP branch merge?

fadi-george:
update kmp submodule to use latest main

fadi-george (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/debug/internal/logging/otel/android/OtelPlatformProvider.kt
```diff
@@ -64,6 +64,23 @@ internal class OtelPlatformProvider(
...
     override val sdkWrapperVersion: String? = OneSignalWrapper.sdkVersion
 
+    // Host language version for remote-log dashboard filtering (ossdk.kotlin_version).
+    // KotlinVersion.CURRENT reflects the stdlib the SDK was compiled against.
+    override val kotlinVersion: String? = KotlinVersion.CURRENT.toString()
+
+    // Android has no Swift toolchain.
+    override val swiftVersion: String? = null
+
+    // Extra toolchain labels under ossdk.* for dashboard filtering. Prefer values that are
```
fadi-george: lets have smaller comments

PR #2725, opened 2026-08-25
fix: [SDK-4946] prefer google-services.json over the shared FCM project

abdulraqeeb33:
# Description
## One Line Summary
Prefer the host app's `google-services.json` for FCM registration so we stop mixing the dashboard sender id with OneSignal's shared Firebase project (`onesignal-shared-public` / `754795614042`).

### Motivation
Google now requires the FCM sender id, project id, application id, and API key to belong to the same Firebase project, and Firebase Installation IDs are issued per project. The Android SDK has historically initialized its own `FirebaseApp` with OneSignal's shared public credentials and the customer's sender id at runtime. That mixed configuration is rejected for Installation ID registration (`ApiException: 8`) and is the dead end behind SDK-4946.

The SDK can get off the shared project by reading the file Google already asks every app to ship: `google-services.json` (compiled into string resources by the google-services Gradle plugin, then loaded via `FirebaseApp.initializeApp(context)`).

### Scope
FCM registration now picks **one** consistent project, in this order:

1. **Host `google-services.json`** — reuse the default `FirebaseApp` when its sender id matches the OneSignal dashboard.
2. **Backend `fcm` params** from `android_params.js`, when complete. These are now read at registration time, not snapshotted in the `PushRegistratorFCM` constructor (so they were previously ignored if the registrator was created before params hydration).
3. **OneSignal shared project** (`onesignal-shared-public` / `754795614042`) as a last-resort fallback for apps that never added `google-services.json`. This still will not work with Installation ID registration.

When the legacy `getToken()` API is disabled (`firebase_messaging_installation_id_enabled`), registration falls back to `FirebaseMessaging.register()` against that same host project and returns the Firebase Installation ID. Apps that opt in still need their own `google-services.json` whose sender id matches the dashboard.

If `google-services.json` is present but its sender id does not match the dashboard, the SDK logs a warning and does not use it (so an Analytics project can still coexist with a different OneSignal FCM project via backend/shared credentials).

No public API changes. Apps without `google-services.json` keep the legacy shared-project path.

# Testing
## Unit testing
- `FcmFirebaseConfigResolverTests` covers host match, host mismatch, incomplete host/backend credentials, backend params, and the shared fallback.
- `FCMTokenProviderTests` covers legacy tokens, Installation ID registration, missing default app / sender mismatch diagnostics, reflection, and error propagation.
- `PushRegistratorFCMTests` covers reusing the host app, backend params (including reading them after construction), shared fallback, sender mismatch, Installation ID wiring, and `FirebaseOptions` mapping.

`:OneSignal:notifications:testDebugUnitTest`, Spotless, and Detekt pass locally.

## Manual testing
Not run on a device in this environment. The registration branches are covered with unit tests. To verify on a device:

1. Drop a `google-services.json` whose sender id matches the OneSignal dashboard into the example app, apply the google-services plugin, and confirm logs say registration is using that project and that a push arrives.
2. Enable `firebase_messaging_installation_id_enabled` and confirm the subscription token is a Firebase Installation ID and delivery still works.
3. Remove `google-services.json` and confirm the shared-project fallback still returns a legacy FCM token when the Installation ID flag is off.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing — FCM registration source of truth

## Testing
   - [ ] I have personally tested this on my device, or explained why that is not possible — see Manual testing

## Final pass

<div><a href="https://cursor.com/agents/bc-e18cf81a-3682-43a3-8445-8b5b3827f536?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e18cf81a-3682-43a3-8445-8b5b3827f536&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/OneSignalSDK/coverage/jacoco.gradle
+++ b/OneSignalSDK/coverage/jacoco.gradle
@@ -16,6 +16,16 @@ subprojects {
                     testCoverageEnabled = true
                 }
             }
+
+            testOptions {
+                // Robolectric loads classes through its own classloader, which leaves them without
+                //   a code source location. JaCoCo skips those by default, so Robolectric tests
+                //   would contribute no coverage at all.
+                unitTests.all { test ->
+                    test.jacoco.includeNoLocationClasses = true
+                    test.jacoco.excludes = ['jdk.internal.*']
+                }
+            }
         }
         
         def coverageExcludes = [
--- a/OneSignalSDK/onesignal/notifications/build.gradle
+++ b/OneSignalSDK/onesignal/notifications/build.gradle
@@ -76,6 +76,8 @@ dependencies {
 
     // NOTE: firebase-messaging:24.0.0 requires customer's project to use
     //   compileSdkVersion 34 or higher.
+    // `require` is intentionally non-strict: this module compiles against the preferred 24.0.0,
+    //   while an app can select a newer version through Gradle conflict resolution.
     api('com.google.firebase:firebase-messaging') {
         version {
             require '[23.0.8, 24.0.99]'
--- a/OneSignalSDK/onesignal/notifications/consumer-rules.pro
+++ b/OneSignalSDK/onesignal/notifications/consumer-rules.pro
@@ -25,6 +25,16 @@
 -dontwarn com.google.firebase.**
 -dontwarn com.google.android.gms.**
 
+# PushRegistratorFCM looks up FirebaseMessaging.register() by name, because this module compiles
+# against firebase-messaging 24.x where the method does not exist yet. Nothing references it
+# symbolically, it carries no @Keep, and firebase-messaging ships no consumer rules, so R8 full mode
+# (AGP 8+) renames it along with the rest of the class, causing:
+# java.lang.NoSuchMethodException: com.google.firebase.messaging.FirebaseMessaging.register []
+# keepclassmembers rather than keep, so Huawei apps that exclude firebase-messaging are unaffected.
+-keepclassmembers class com.google.firebase.messaging.FirebaseMessaging {
+    public *** register();
+}
+
 # ADM handlers are instantiated by name from the app manifest AND their on* lifecycle callbacks
 # (onMessage/onRegistered/onRegistrationError/onUnregistered) are invoked by the ADM framework, not
 # the SDK, so keep both constructors and those methods. (Amazon-device-only path, untestable in CI.)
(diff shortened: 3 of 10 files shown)


fadi-george:
Host `google-services.json` match path looks right. Three things that can still pick the wrong project:

1. **First `getToken` pins `firebaseApp` for the process.** Constructor snapshot is gone, but resolution is still snapshotted one layer later. `start()` can register against cached/incomplete `fcmParams` before `android_params.js` hydrates, and the later retry cannot switch off `SHARED_DEFAULT`. Tests only cover params arriving after construction and *before* the first `getToken`.

2. **Installation ID can use a default app the resolver already rejected.** On `SHARED_DEFAULT`, `firebaseAppForInstallationId()` falls back to `defaultFirebaseApp` and only rechecks sender id. An incomplete host app (matching sender, missing `projectId`) is dropped for push, then reused for FID.

3. **`BACKEND` can still mix projects, then run FID on that mix.** `backendCredentials()` always stamps the dashboard sender onto `fcmParams`. Completeness is four non-blank strings. Any non-`SHARED_DEFAULT` source is treated as safe for FID. If `android_params.js` still ships the shared public project as a complete `fcm` object, this is classified as `BACKEND` and hits the same mixed config the ticket is about.

Restrict FID to `GOOGLE_SERVICES` (or require backend app id / project number to match the dashboard sender and not the shared defaults). Don't pin the first resolution. Drive II selection with the same predicate as the resolver.

fadi-george:
Id make this a separate linear ticket to not mix approaches for SDK-4946.
So a new branch and remake the pr.

PR #2726, opened 2026-08-25
chore: [SDK-5083] add an inert-by-default notification service extension to the demo

nan-li:
# Description
## One Line Summary

Adds a notification service extension to the Android demo, hidden unless you build with `-PSHOW_NSE_SECTION=true`.

### Motivation

Add NSE example to debug NSE issues.

### Scope

`examples/demo` only. No SDK source changes, no public API changes.

The demo screen is unchanged in a default build. `MainScreen` wraps the section in `if (BuildConfig.SHOW_NSE_SECTION)`, resolved through the existing `demoOverride()` helper (`-P` → `local.properties` → `false`). Only the UI is gated: the extension stays registered, so the class still compiles in `demo-build` and still exercises the `-keep` rule in `onesignal/notifications/consumer-rules.pro`. The flag is a compile-time constant, so R8 drops the section from a default release build.

Build with `-PSHOW_NSE_SECTION=true` to get the section. Every switch defaults to off.
image

| Toggle | Behavior |
| --- | --- |
| Enable Extension | Master switch. Off returns immediately; on logs id, sent time, and the resolved channel under `[Demo]NSE`. |
| Apply Extender | Prefixes the title through a `NotificationCompat.Extender`. |
| Force High Importance Channel | Moves the notification onto an app-owned `IMPORTANCE_HIGH` channel. |
| Delay Display | `preventDefault()`, then `display()` five seconds later. |
| Discard | `preventDefault(true)`. Takes precedence over the others. |

Switches persist through `SharedPreferenceUtil`, which the extension reads directly since it runs whether or not the app is open.

### Design notes

The channel readout uses `NotificationCompat.getChannelId(builder.build())` inside the extender, the only place an extension sees the SDK's choice. Reading `chnl` off `rawPayload` would report what the payload asked for, not what the SDK picked, and the two diverge on a restore. That readout is why the extender is set unconditionally, which costs nothing: an extender cannot change what displays, because `processHandlerResponse` gates on `canDisplay` before `shouldDisplayNotification` ever reads `hasExtender()`.

`event.restoring` is not on `INotificationReceivedEvent` yet, so the log line omits it. #2723 adds it, and a TODO marks the spot.

A separate commit routes the demo's 127 log calls through `DemoLog`, which marks both tag and message so `logcat -s` still filters and a line stays recognizable in the message column. The five SDK-forwarding calls in `MainApplication` stay unmarked.

```
D/[Demo]MainViewModel: [Demo] Sending notification: Simple
```

# Testing

## Unit testing

None added. Demo code with no test target, and the compile is the guard: `ci.yml`'s `demo-build` builds `:app` from inside `OneSignalSDK/`, where `settings.gradle` substitutes the published dependency with local source.

On a default release build I confirmed `DemoNotificationServiceExtension` survives R8 with its class-level mapping entry while `NotificationExtensionSection` is stripped to zero entries.

## Manual testing

Pixel 7 emulator, API 34, Google Play image, `gms` debug against local SDK source, section shown.

Each switch changes behavior: extender prefixes the title, forced channel posts on `demo_nse_high_importance` at `mImportance=4`, delay displays five seconds later, discard posts nothing. With the master switch off nothing is logged and notifications post normally.

Two beyond the matrix. Killing the process with `adb shell am kill` and sending a push showed the extension firing in a fresh pid, reading its switches from SharedPreferences with no `MainViewModel` alive. And after `adb reboot` the same notifications came back logging `channel=restored_OS_notifications` at `importance=2`, against `fcm_fallback_notification_channel` at `importance=3` on first delivery, which is why the channel is read from the builder.

The shared Appium suite passes on a default build, 32 passing with the 2 iOS-only Live Activities skips. It does not pass with `-PSHOW_NSE_SECTION=true`: the taller screen pushes `iam_info_icon` past the `height * 0.82` line that triggers `nudgeAboveBottomOverlay` in `sdk-shared`'s `appium/tests/helpers/app.ts`, and that path throws `getElementRect` with an undefined `elementId` past its own `catch`. The demo is fine, tapping the icon opens the right tooltip. Placement is not the cause; moving the section to the bottom of the screen fails identically. Noted in `build.md`. CI and `e2e.yml` both build the default configuration.

`spotlessCheck`, `detekt`, `assembleGmsRelease`, and `compileHuaweiDebugKotlin` all pass.

# Affected code checklist
   - [x] Notifications
      - [x] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

Ticked because the demo extension exercises those paths. No SDK behavior changes.

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
+++ b/examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
@@ -19,9 +19,11 @@ import com.onesignal.notifications.INotificationServiceExtension
  * runtime instead of at build time. That reflection is also why the class needs the -keep rule
  * in `onesignal/notifications/consumer-rules.pro` to survive R8.
  *
- * Nothing here runs until you turn it on in the demo's Notification Service Extension section.
- * An always-on extension would change the baseline for every notification the demo sends, and
- * anyone chasing a grouping or channel bug would end up debugging this file without knowing it.
+ * Nothing here runs until Enable Extension is turned on in the demo's Notification Service
+ * Extension section. An always-on extension would change the baseline for every notification the
+ * demo sends, and anyone chasing a grouping or channel bug would end up debugging this file
+ * without knowing it. That master toggle is the only UI; the behavior switches are flipped in
+ * code through [NotificationExtensionOptions] and SharedPreferenceUtil.
  *
  * The switches come from SharedPreferences, not MainViewModel. This is called whether or not the
  * app is open, so there may be no ViewModel yet.
--- a/examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
+++ b/examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
@@ -4,9 +4,9 @@ import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.content.Context
 import android.os.Build
-import android.util.Log
 import androidx.core.app.NotificationCompat
 import com.onesignal.example.data.model.NotificationExtensionOptions
+import com.onesignal.example.util.DemoLog
 import com.onesignal.example.util.SharedPreferenceUtil
 import com.onesignal.notifications.IDisplayableMutableNotification
 import com.onesignal.notifications.INotificationReceivedEvent
@@ -36,7 +36,7 @@ class DemoNotificationServiceExtension : INotificationServiceExtension {
         if (options.logDetails) {
             // TODO: [SDK-5011] log `event.restoring` here once it ships. Reading it next to the
             // channel below is the whole diagnosis for a notification that re-alerts on reboot.
-            Log.d(
+            DemoLog.d(
                 TAG,
                 "received androidNotificationId=${notification.androidNotificationId}" +
                     " notificationId=${notification.notificationId}" +
@@ -47,7 +47,7 @@ class DemoNotificationServiceExtension : INotificationServiceExtension {
 
         if (options.discard) {
             if (options.logDetails) {
-                Log.d(TAG, "discarding androidNotificationId=${notification.androidNotificationId}")
+                DemoLog.d(TAG, "discarding androidNotificationId=${notification.androidNotificationId}")
             }
             event.preventDefault(true)
             return
@@ -79,7 +79,7 @@ class DemoNotificationServiceExtension : INotificationServiceExtension {
             // extension can see what the SDK picked. A restored notification lands on
             // `restored_OS_notifications` no matter what the payload asked for, and the payload
             // by itself never shows that.
-            Log.d(
+            DemoLog.d(
                 TAG,
                 "building androidNotificationId=${notification.androidNotificationId}" +
                     " channel=${NotificationCompat.getChannelId(builder.build())}",
@@ -122,7 +122,7 @@ class DemoNotificationServiceExtension : INotificationServiceExtension {
     }
 
     private companion object {
-        const val TAG = "DemoNSE"
+        const val TAG = "NSE"
         const val HIGH_IMPORTANCE_CHANNEL_ID = "demo_nse_high_importance"
 
         // Well under the SDK's 30 second wait for the extension, and long enough to watch the
--- a/examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
+++ b/examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
@@ -1,0 +1,132 @@
+package com.onesignal.example.notification
+
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.content.Context
+import android.os.Build
+import android.util.Log
+import androidx.core.app.NotificationCompat
+import com.onesignal.example.data.model.NotificationExtensionOptions
+import com.onesignal.example.util.SharedPreferenceUtil
+import com.onesignal.notifications.IDisplayableMutableNotification
+import com.onesignal.notifications.INotificationReceivedEvent
+import com.onesignal.notifications.INotificationServiceExtension
+
+/**
+ * The demo's notification service extension, registered in AndroidManifest.xml under the
+ * `com.onesignal.NotificationServiceExtension` meta-data key. The SDK resolves that string with
+ * Class.forName and calls the no-arg constructor, so a typo in the manifest fails silently at
+ * runtime instead of at build time. That reflection is also why the class needs the -keep rule
+ * in `onesignal/notifications/consumer-rules.pro` to survive R8.
+ *
+ * Nothing here runs until you turn it on in the demo's Notification Service Extension section.
+ * An always-on extension would change the baseline for every notification the demo sends, and
+ * anyone chasing a grouping or channel bug would end up debugging this file without knowing it.
+ *
+ * The switches come from SharedPreferences, not MainViewModel. This is called whether or not the
+ * app is open, so there may be no ViewModel yet.
+ */
+class DemoNotificationServiceExtension : INotificationServiceExtension {
+    override fun onNotificationReceived(event: INotificationReceivedEvent) {
+        val options = SharedPreferenceUtil.getNotificationExtensionOptions(event.context)
+        if (!options.enabled) return
+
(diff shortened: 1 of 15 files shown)


examples/demo/app/src/main/java/com/onesignal/example/util/DemoLog.kt
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * SDK output that MainApplication's log listener forwards does not come through here. Those
+ * lines belong to the SDK, and marking them would bury the demo's own output when you grep.
+ */
+object DemoLog {
+    private const val PREFIX = "[Demo]"
+
+    fun v(tag: String, message: String) = Log.v(PREFIX + tag, "$PREFIX $message")
+
+    fun d(tag: String, message: String) = Log.d(PREFIX + tag, "$PREFIX $message")
```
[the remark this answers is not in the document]
nan-li: the 23-char limit only throws from Log.isLoggable, which nothing in this path calls. Log.d doesn't validate and doesn't truncate, so logcat -s matches fine.

examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
```diff
@@ -0,0 +1,132 @@
...
+            notification.setExtender(buildExtender(event.context, notification, options))
+        }
+
+        if (options.delayDisplay) {
+            event.preventDefault()
+            Thread {
+                Thread.sleep(DISPLAY_DELAY_MS)
+                notification.display()
+            }.start()
+        }
```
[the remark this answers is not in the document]
nan-li: risk is near zero and this is just a demo app

fadi-george:
I would check this doesnt break the e2e tests suit, you can test with sdk-shared run-local

fadi-george (changes requested):
Ideally we dont modify the current demo and just maybe make another example e.g. examples/demo-nse

nan-li:
> Ideally we dont modify the current demo and just maybe make another example e.g. examples/demo-nse

The NSE is disabled by default, do you think we really need another demo app just for the NSE?

image
image

examples/demo/app/src/main/java/com/onesignal/example/notification/DemoNotificationServiceExtension.kt
```diff
@@ -0,0 +1,132 @@
...
+            event.preventDefault(true)
+            return
+        }
+
+        // Set an extender only when a switch needs one. An extender makes the SDK display a
+        // data-only push that carries no `alert` (NotificationGenerationProcessor
+        // .shouldDisplayNotification), so an always-installed no-op extender is not free.
+        if (options.logDetails || options.applyExtender || options.forceHighImportanceChannel) {
+            notification.setExtender(buildExtender(event.context, notification, options))
+        }
```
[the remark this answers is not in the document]
nan-li: done

nan-li:
> I would check this doesnt break the e2e tests suit, you can test with sdk-shared run-local

Done locally, all passed. Note that I don't think the e2e workflow itself added in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2652 has ever ran

fadi-george:
> Ideally we dont modify the current demo and just maybe make another example e.g. examples/demo-nse
> The NSE is disabled by default, do you think we really need another demo app just for the NSE?  

The demo app is supposed to almost the same for all the wrappers. Ideally we don't add stuff to it willy-nilly. I still think this could be a separate example i.e. examples/demo-nse it can just have the section you've made and maybe some test notification if it needs it.

nan-li:
> > Ideally we dont modify the current demo and just maybe make another example e.g. examples/demo-nse
> > The NSE is disabled by default, do you think we really need another demo app just for the NSE?
> 
> The demo app is supposed to almost the same for all the wrappers. Ideally we don't add stuff to it willy-nilly. I still think this could be a separate example i.e. examples/demo-nse it can just have the section you've made and maybe some test notification if it needs it.

We have the NSE in our ios demo ( granted it is more crucial for ios for core features like received receipts), and it used to be included in the prev Android example ([see old example](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.6.0/Examples/OneSignalDemo/app/src/main/java/com/onesignal/sdktest/notification/NotificationServiceExtension.java)) that never ported over to the new sample. I would argue this keeps parity with previous version of the sample app and our current iOS sample rather than an extra feature that warrants to be in a separate example. But if you still feel strongly I can make a separate demo,  it's just when you start making more demos people are rarely going to run the other ones in their day-to-day, compared to other demos we have that exclude location or don't use pods. Those are demos that rarely run and seem to be for build verification.

nan-li:
Removed all the option toggles from the UI per JonF's feedback today, since the NSE is quite customizable, we don't want to lead users into only considering limited options, or taking our options as best practices.

abdulraqeeb33:
yeah we definitely shouldnt have another demo app. 
and we if we do add a new field/feature in one SDK, all SDK's need to be updated accordingly including their appium tests.

examples/demo/app/src/main/java/com/onesignal/example/ui/main/MainScreen.kt
```diff
@@ -161,6 +163,11 @@ fun MainScreen(viewModel: MainViewModel) {
                     onInfoClick = { showTooltipDialog = "sendPushNotification" }
                 )
 
+                NotificationExtensionSection(
```
fadi-george: Maybe render this section last as to not interfere with appium tests.
Potentially could also move to secondary screen.

fadi-george:
Otherwise maybe we can gate rendering the whole section by a gradle property e.g. onesignal.showNSESection ?

examples/demo/app/src/main/java/com/onesignal/example/ui/main/MainScreen.kt
abdulraqeeb33: yeah i think this is a better solution, lets move functionality that are meant for one platform into another screen. - Secondary screen. That way we dont break appium tests.

nan-li: It's at the bottom of the push section where it makes sense to be: after push ID, enabled toggle, "send push notification" section.

> lets move functionality that are meant for one platform into another screen

Can you clarify the Appium issue that is breaking? Because they passed locally for me. I did notice that the e2e workflow added in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2652 doesn't look like it has ever ran? I may be wrong and looking in the wrong places on this.

nan-li: Ok, since we are omitting mention of Android NSE for wrappers, I'll remove the UI and just make it code only (for when testing this)

fadi-george: what are you planning for the code-only change?

nan-li: added the gradle property to show it for testing

fadi-george:
dont need the collapse toggle (hide/show options)

fadi-george:
do we need a toggle for log details, maybe it should be controlled by enable extension toggle?

fadi-george:
Not blocking as we could do this some other time buy maybe Delay display and discard toggles could be moved to send push notification (and the wrappers)? doesnt need the NSE i believe unless its for background testing

fadi-george (approved):

examples/demo/app/src/main/java/com/onesignal/example/ui/main/MainScreen.kt
abdulraqeeb33: this works @nan-li  - thank oyu

abdulraqeeb33 (approved):

nan-li:
> Not blocking as we could do this some other time buy maybe Delay display and discard toggles could be moved to send push notification (and the wrappers)? doesnt need the NSE i believe unless its for background testing

Yeah, these are for the methods on the NSE, but we can follow up and add the foreground listener API to all the demo apps for testing, since you have been doing a bit of the foreground stuff on wrappers.

Issue #2696, opened 2026-07-24
[Bug]: FCM registration fails with "API disabled" when Firebase Installation ID registration is enabled (firebase-messaging 25.1.x)


hjdealba96:
### Description

OneSignal fails to retrieve an FCM push token when the app has opted into Firebase's new Firebase Installation ID (FID)-based FCM registration. `PushRegistratorFCM` calls the legacy `FirebaseMessaging.getToken()`, which `firebase-messaging` 25.1.x disables when FID-based registration is enabled, throwing `IllegalStateException: API disabled`. As a result no push token is obtained and the device never completes push registration.

### Environment

- OneSignal Android SDK: **5.9.7**
- firebase-messaging: **25.1.1** (Firebase Android BoM 34.16.x)
- Also present with Braze Android SDK 43.0.0 in the same app (see root cause)

### Stack trace

```
Unknown error getting FCM Token
java.lang.IllegalStateException: API disabled. Please use register() instead or enable this API by removing <meta-data android:name="firebase_messaging_installation_id_enabled" android:value="true" /> from your app's manifest.
    at com.google.firebase.messaging.FirebaseMessaging.getToken(FirebaseMessaging.java:469)
    at com.onesignal.notifications.internal.registration.impl.PushRegistratorFCM.getTokenWithClassFirebaseMessaging(PushRegistratorFCM.kt:62)
    at com.onesignal.notifications.internal.registration.impl.PushRegistratorFCM.getToken(PushRegistratorFCM.kt:52)
    at com.onesignal.notifications.internal.registration.impl.PushRegistratorAbstractGoogle.attemptRegistration(PushRegistratorAbstractGoogle.kt:128)
    at com.onesignal.notifications.internal.pushtoken.PushTokenManager.retrievePushToken(PushTokenManager.kt:31)
```

### Root cause

`firebase-messaging` 25.1.0 added FCM registration via the Firebase Installation ID and **deprecated `getToken()`, `deleteToken()`, and `onNewToken()`**. When any component opts the app into FID-based registration, Firebase disables the legacy `getToken()` API app-wide.

A common trigger is the **Braze Android SDK 43.0.0**, which automatically enables this by injecting `<meta-data android:name="firebase_messaging_installation_id_enabled" android:value="true" />` into the merged manifest. OneSignal's `PushRegistratorFCM.getTokenWithClassFirebaseMessaging()` still calls the now-disabled `getToken()`, so registration fails.

[...]

PR #2729, opened 2026-09-01
feat: [SDK-5146] send host FID-readiness signals on API request headers

abdulraqeeb33:
# Description
## One Line Summary
Add a `OneSignal-Fid-Env` header on OneSignal API requests so we can measure how many host apps already meet Firebase Installation ID requirements.

### Motivation
[SDK-5146](https://linear.app/onesignal/issue/SDK-5146/add-logging-to-sdk-for-tracking-fid-requirements) needs a census of the install base before FID migration. Remote logging is opt-in per app via `logging_config`, so it cannot answer "30% of app IDs already have the plugin." A request header on `android_params` (and other API calls) can.

`google-services.json` is not packaged in the APK. `gs=1` means the Google Services plugin wrote `google_app_id`. Gradle wrapper is not a FID gate and is not in the APK. `firebase-messaging` version is not in the header: R8 strips `BuildConfig`, and cohort is already `gs` / `snd` / `def` / `flag` plus the subscription identifier shape (legacy token vs FID).

### Scope
Header on every `HttpClient` request:

```
OneSignal-Fid-Env: gs=1;agp=8.8.2;flag=0;def=1;prov=1;min=24;tgt=35;snd=1
```

- `-` means unknown.
- Probe failures omit the header and never fail the request.
- A rejected header value omits `OneSignal-Fid-Env` instead of failing the call.
- No public API change. No change to FCM registration behavior.
- `CoreModule` only adds `FidEnvService` next to `DeviceService`. The dummy notifications/IAM/location managers are unchanged.
- Backend still needs to ingest the header to compute distinct-app-ID percentages. Aggregate last `snd` per install. `snd` is `-` until params hydrate for the current app ID (`dashboardSenderAppId` must match `appId`). An app update that changes the OneSignal app ID is uncommon; the stamp keeps that first request from using the previous sender. FCM is unchanged (`isInitializedWithRemote`).
- `tgt` is `ApplicationInfo.targetSdkVersion`. A PackageManager miss is not encoded as 21.

Cohorts (combine the header with the identifier already on the subscription):
- `flag=1`: already on the FID path (Braze or host). Missing identifier with `flag=1` is the #2696 `getToken()` break.
- `gs=1;snd=1;def=1;flag=0`: ready now, still on legacy `getToken()`. First migrate cohort.
- `gs=1;snd=1;prov=0`: ready, consent-gated. Do not call `initializeApp()`.
- `gs=1;snd=0`: plugin present, dashboard sender mismatch.
- `gs=0`: still on shared `onesignal-shared-public`.

Field notes: [SDK-5146](https://linear.app/onesignal/issue/SDK-5146/add-logging-to-sdk-for-tracking-fid-requirements).

# Testing
## Unit testing
- `FidEnvTests`: header encoding, ASCII sanitization, AGP parser, sender match.
- `FidEnvReaderTests`: Robolectric resource/flag/metadata probes.
- `HttpClientTests`: header present on all methods; omitted when the probe returns empty.

## Manual testing
Not run on a device. Unit tests cover encoding and HttpClient header attachment.

# Affected code checklist
   - [x] REST API requests
   - [ ] Notifications
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

Device testing is not done for this telemetry-only header. Core unit tests for FidEnv and HttpClient passed locally.

## Final pass

### DIFF
--- a/OneSignalSDK/detekt/detekt-baseline-core.xml
+++ b/OneSignalSDK/detekt/detekt-baseline-core.xml
@@ -25,8 +25,11 @@
     <ID>ConstructorParameterNaming:DatabaseCursor.kt$DatabaseCursor$private val _cursor: Cursor</ID>
     <ID>ConstructorParameterNaming:DatabaseProvider.kt$DatabaseProvider$private val _application: IApplicationService</ID>
     <ID>ConstructorParameterNaming:DeviceService.kt$DeviceService$private val _applicationService: IApplicationService</ID>
+    <ID>ConstructorParameterNaming:FidEnv.kt$FidEnvService$private val _applicationService: IApplicationService</ID>
+    <ID>ConstructorParameterNaming:FidEnv.kt$FidEnvService$private val _configModelStore: ConfigModelStore</ID>
     <ID>ConstructorParameterNaming:HttpClient.kt$HttpClient$private val _configModelStore: ConfigModelStore</ID>
     <ID>ConstructorParameterNaming:HttpClient.kt$HttpClient$private val _connectionFactory: IHttpConnectionFactory</ID>
+    <ID>ConstructorParameterNaming:HttpClient.kt$HttpClient$private val _fidEnv: IFidEnv</ID>
     <ID>ConstructorParameterNaming:HttpClient.kt$HttpClient$private val _installIdService: IInstallIdService</ID>
     <ID>ConstructorParameterNaming:HttpClient.kt$HttpClient$private val _prefs: IPreferencesService</ID>
     <ID>ConstructorParameterNaming:HttpClient.kt$HttpClient$private val _time: ITime</ID>
--- a/OneSignalSDK/detekt/detekt-baseline-core.xml
+++ b/OneSignalSDK/detekt/detekt-baseline-core.xml
@@ -190,6 +190,7 @@
     <ID>InstanceOfCheckForException:HttpClient.kt$HttpClient$t is UnknownHostException</ID>
     <ID>LongMethod:ApplicationService.kt$ApplicationService$override suspend fun waitUntilSystemConditionsAvailable(): Boolean</ID>
     <ID>LongMethod:ConfigModelStoreListener.kt$ConfigModelStoreListener$private fun fetchParams()</ID>
+    <ID>LongMethod:CoreModule.kt$CoreModule$override fun register(builder: ServiceBuilder)</ID>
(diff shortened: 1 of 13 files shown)


fadi-george:
Potential issues:
- `snd` can use the previous app ID’s sender after an app ID switch. Clear or scope `googleProjectNumber` until the new params load.
- `prov=1` includes disabled `FirebaseInitProvider`s, misclassifying consent-gated apps. Check `ProviderInfo.enabled`.
- A target SDK lookup failure reports `tgt=21` instead of unknown (`-`).

abdulraqeeb33:
@fadi-george thanks, replies below.

**snd after an app ID switch.** We do not switch app IDs at runtime (`initWithContext` takes the ID once, there is no `setAppId`). `snd` stays `-` until params hydrate (`isInitializedWithRemote`), then uses `googleProjectNumber`. A host rebuild onto a new OneSignal app ID can send one stale `snd` until the next params GET returns.

**prov and disabled providers.** Left as-is. The documented consent path is `tools:node="remove"`, which drops `FirebaseInitProvider` from the merged manifest, so we send `prov=0`. `GET_PROVIDERS` without `MATCH_DISABLED_COMPONENTS` typically omits `android:enabled="false"` already. 

**tgt=21 on lookup failure.** Fixed. `tgt` is `context.applicationInfo.targetSdkVersion`, not `AndroidUtils.getTargetSdkVersion` (that helper fell back to 21 / LOLLIPOP).

fadi-george:
snd: still valid for an app update changing the OneSignal app ID. The code checks only persisted isInitializedWithRemote, not which app ID hydrated the sender, so the first request can send stale data.
Avoid impacting FCM by storing the app ID associated with googleProjectNumber and only using snd when it matches the current app ID.

abdulraqeeb33:
@fadi-george good catch on the app-update path.

That is uncommon (a Play update that ships a different OneSignal app ID), but the first request after that cutover could encode `snd` against the previous sender. Better to address it than leave a stale bit.

`fetchParams` now persists `dashboardSenderAppId` next to `googleProjectNumber`. `snd` is encoded only when params have hydrated **and** that stamp matches the current `appId`; otherwise `snd=-`. FCM still waits on `isInitializedWithRemote`, so registration is unchanged.

fadi-george (approved):

PR #2732, opened 2026-09-03
feat: [SDK-5156] wire named events into the log pipeline

nan-li:
# Description
## One Line Summary
Registers the shared KMP event recorder as a service and attaches it to the remote logging telemetry, so named events can ship on the log pipeline.

### Motivation
Remote logging cannot carry a structured fact, and INFO lines never ship at the production default of ERROR. Named events use the same gate as crashes (a remote log level that is present and not `NONE`) plus their own catalog flag, which is off by default. The first event, `sdk.device_gesture`, is the only usage signal for the clipboard gesture (SDK-5088), which makes no request.

Ticket SDK-5156. Design doc: "Named events on the log pipeline". Depends on [OneSignal-KMP-SDK#24](https://github.com/OneSignal/OneSignal-KMP-SDK/pull/24).

### Scope
- `CoreModule` registers the recorder as `IObservabilityEventRecorder`, answering the recorder's flag lookups (`IFeatureFlagReader`) from `IFeatureManager`.
- `OneSignalImp` hands the recorder to the observability lifecycle manager right after bootstrap, fail-open like the calls around it.
- `LoggerLifecycleManager` attaches the recorder to every remote telemetry it installs, detaches it before shutting one down, and logs those transitions at INFO so a log capture can say whether the recorder was ever attached.
- The KMP submodule moves to the tip of the KMP PR branch, currently its merge of KMP main, so the pin also carries the API 21 minSdk change that #2736 brought to `main`. It must be re-pointed to the release tag before this merges.

Not in this PR: the `record` call in the gesture detector. It lands on the SDK-5088 branch once the clipboard gesture is in. No event ships until `sdk_event_device_gesture` is turned on for an app.

# Testing
## Unit testing
`LoggerLifecycleManagerTest` and `LoggerLifecycleManagerFaultTest` (58 tests, 13 new). They cover: the recorder waits for telemetry and attaches to the one HYDRATE installs, attaches at once when the cached-config telemetry is already live, follows a level change to the replacement telemetry, is detached before shutdown on disable, is not re-attached on a repeated config, never attaches below API 26, and a recorder handed over later takes over from the first. One test wires the real KMP recorder the way `CoreModule` does and proves an event recorded before HYDRATE ships while an INFO log line does not. The fault tests prove a throwing recorder cannot fail a level change, block teardown, or reach init.

Full `:core` suite, `detekt` and `spotlessCheck` pass locally.

## Manual testing
Not run on a device. Nothing calls `record` yet, so there is no user-visible behaviour to exercise.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing
   - [ ] I have personally tested this on my device, or explained why that is not possible (see manual testing)

## Final pass


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/LoggerLifecycleManager.kt
```diff
@@ -310,6 +352,16 @@ internal class LoggerLifecycleManager(
...
+        // Isolated like the shutdown below: the telemetry is already live, so a recorder fault
+        // must not fail the level change.
+        try {
+            eventRecorder?.let {
+                it.attach(telemetry)
+                Logging.info("OneSignal: event recorder attached to the remote telemetry at level $logLevel")
+            }
+        } catch (t: Throwable) {
+            Logging.warn("OneSignal: Failed to attach the event recorder: ${t.message}", t)
+        }
```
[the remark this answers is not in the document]
nan-li: The real recorder's `attach` never throws; it swallows its own faults. Keeping it out of the liveness check on purpose, so a recorder problem can never hold up the pipeline.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -419,6 +420,14 @@ internal class OneSignalImp : IOneSignal,
...
             // manager to config store events so it reacts to fresh remote config.
             observabilityManager?.subscribeToConfigStore(services.getService<ConfigModelStore>())
+            // The event recorder is a container service, so it can only be handed over now. Fail-open
+            // like the observability calls around it: telemetry must not be able to fail init.
+            val eventRecorder = services.getServiceOrNull<IObservabilityEventRecorder>()
+            if (eventRecorder != null) {
+                observabilityManager?.attachEventRecorder(eventRecorder)
+            } else {
+                Logging.warn("OneSignal: event recorder unavailable, observability events will not ship")
+            }
```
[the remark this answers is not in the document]
nan-li: The resolve cannot fail in practice, since bootstrap has already built the feature manager it needs and the factory only constructs an object, so the comment now says that instead of claiming fail-open

fadi-george:
rebase or fix submodule conflict

fadi-george (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
```diff
@@ -419,6 +420,15 @@ internal class OneSignalImp : IOneSignal,
             // Now that the IoC container is ready, subscribe the observability lifecycle
             // manager to config store events so it reacts to fresh remote config.
             observabilityManager?.subscribeToConfigStore(services.getService<ConfigModelStore>())
+            // The event recorder is a container service, so it can only be handed over now. The resolve
+            // cannot fail in practice (bootstrap already built the feature manager it needs), and the
+            // hand-over is fail-open inside the manager.
+            val eventRecorder = services.getServiceOrNull<IObservabilityEventRecorder>()
+            if (eventRecorder != null) {
+                observabilityManager?.attachEventRecorder(eventRecorder)
```
abdulraqeeb33, line 428: Nit: retry-after-FAILED builds a second manager without unsubscribing the first, and this attach already ran if `resolveAppId` fails. Both then drive the same IoC recorder. Reuse the manager, or tear the old one down first.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/LoggerLifecycleManager.kt
```diff
@@ -310,6 +352,16 @@ internal class LoggerLifecycleManager(
         remoteTelemetry = telemetry
         activeLogLevel = logLevel
         Logging.setLoggerTelemetry(telemetry, shouldSend)
+        // Isolated like the shutdown below: the telemetry is already live, so a recorder fault
+        // must not fail the level change.
+        try {
+            eventRecorder?.let {
```
abdulraqeeb33, line 358: Nit: disable detaches then shuts down. A level change attaches T2 then shuts T1 with no `detach(previous)`. Harmless while KMP `attach` overwrites the pointer. Adding the detach would match disable.

abdulraqeeb33 (approved):
Nits only. Host wiring is in the right place; gate, queue, and cap already live in KMP. One follow-up before anything calls `record`: retry-after-FAILED currently leaves two managers on one recorder.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt
nan-li: Good call but **not addressing** for the following reason. This needs a first init that reaches initEssentials and then fails, which after that point is only resolveAppId (no app id passed, none in storage) or an unexpected throw, followed by a second init in the same process. The internal context-only callers all need a prior successful init, so in practice it is a dev device initializing with a blank app id on a fresh install and then again with the real one. Empty storage also means no cached logging config, so the first manager is idle at retry time. Both then configure from the same HYDRATE, and the last attach wins on two equivalent telemetries. The second crash handler and ANR detector on that path predate this PR. Leaving it.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/LoggerLifecycleManager.kt
nan-li: Right that disable detaches and a level change does not. It is a no-op either way since attach overwrites the pointer, and adding the detach after the attach would only be safe if detach checks which telemetry it holds. Not changing it for symmetry alone.

PR #2733, opened 2026-09-03
fix: [SDK-5167] split FID census into flag= and reg=

abdulraqeeb33:
# Description
## One Line Summary
Split the FID census so we can count apps that have the FID manifest flag on but are still on the old `getToken()` path.

### Motivation
[SDK-5167](https://linear.app/onesignal/issue/SDK-5167/read-fid-flag-from-firebase-messaging-runtime-not-manifest-only), follow-up to [SDK-5146](https://linear.app/onesignal/issue/SDK-5146/add-logging-to-sdk-for-tracking-fid-requirements) / #2729.

#2729 sent a single `flag=` bit from the merged manifest (`firebase_messaging_installation_id_enabled`). That says "someone opted in via XML." It does not say whether Firebase Messaging actually honors it.

Firebase only honors that metadata in `firebase-messaging` 25.1+ (`FirebaseMessaging.register()`). Older libraries ignore it and keep `getToken()`. Braze (and host copy-paste) can still merge the XML onto those older libraries. Combining the two into one bit hid the group we need: **flag on, still off FID**.

### Scope
Header:

```
OneSignal-Fid-Env: gs=1;agp=8.8.2;flag=1;reg=1;def=1;prov=1;min=24;tgt=35;snd=1
```

| Key | What it answers |
| --- | --- |
| `flag` | Is `firebase_messaging_installation_id_enabled` true in the merged manifest? |
| `reg` | Is `FirebaseMessaging.register()` on the classpath (25.1+)? |

Read them together. Do not AND them in the SDK. Backend can count each slice:

| Header | Meaning |
| --- | --- |
| `flag=1;reg=1` | Flag on **and** on FID. `getToken()` throws (#2696). Already migrated. |
| `flag=1;reg=0` | Flag on, **still off** FID. Old FCM, still `getToken()`. This is the "how many have the flag turned on but are still off" number. |
| `flag=0;reg=1` | 25.1+ present, flag not set. Can opt in, not on FID yet. |
| `flag=0;reg=0` | No opt-in, no 25.1+. |

`flag` is read the same way Firebase does (`containsKey` + `getBoolean`), via a silent `PackageManager` probe. `AndroidUtils.getManifestMetaBoolean` is not used; it `Logging.error`s on `NameNotFoundException`.

`reg` is a `Class.forName` / `getMethod("register")` probe. We do not instantiate `FirebaseMessaging` (that can start token sync). `core` does not compile against `firebase-messaging`.

`core/consumer-rules.pro` keeps `FirebaseMessaging.register()`, `FirebaseApp.getApps(Context)`, and `FirebaseApp.getName()`. Core never compile-calls them, so R8 would strip the names and minified apps would send `reg=0` / `def=0`. Missing class (Huawei exclude) is an R8 info, not a fail.

No public API change. No FCM registration change.

# Testing
## Unit testing
- Header encoding includes `reg=`.
- Robolectric: metadata true without `register()` encodes `flag=1;reg=0`.
- `hasPublicNoArgMethod` covers the true branch (public no-arg `register()`) and false branches (missing class, missing method, method with args).

## Manual testing
Not run on a device. Unit tests cover encoding, the split, and the register lookup.

# Affected code checklist
   - [x] REST API requests
   - [ ] Notifications
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

Device testing is not done for this telemetry-only header change. Core unit tests for FidEnv passed locally.

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/device/impl/FidEnv.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/device/impl/FidEnv.kt
@@ -71,6 +71,16 @@ internal fun dashboardSenderForCensus(
 ): String? =
     if (hydrated && senderAppId == appId) sender else null
 
+@Suppress("TooGenericExceptionCaught")
+internal fun hasPublicNoArgMethod(className: String, methodName: String): Boolean {
+    return try {
+        Class.forName(className).getMethod(methodName)
+        true
+    } catch (_: Throwable) {
+        false
+    }
+}
+
 internal class AndroidFidEnvReader(
     private val context: Context,
 ) {
@@ -131,15 +141,8 @@ internal class AndroidFidEnvReader(
     }
 
     // 25.1+ exposes register(); older libraries ignore the metadata and keep getToken().
-    @Suppress("TooGenericExceptionCaught")
-    private fun hasFirebaseMessagingRegisterApi(): Boolean {
-        return try {
-            Class.forName(FIREBASE_MESSAGING).getMethod("register")
-            true
-        } catch (_: Throwable) {
-            false
-        }
-    }
+    private fun hasFirebaseMessagingRegisterApi(): Boolean =
+        hasPublicNoArgMethod(FIREBASE_MESSAGING, "register")
 
     @Suppress("TooGenericExceptionCaught")
     private fun fidFlagFromMetadata(): Boolean {
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/device/impl/FidEnv.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/device/impl/FidEnv.kt
@@ -18,6 +18,7 @@ internal data class FidEnvSnapshot(
     val googleServices: Boolean,
     val agpVersion: String?,
     val fidFlag: Boolean,
+    val fidRegisterApi: Boolean,
     val defaultFirebaseApp: Boolean,
     val firebaseInitProvider: Boolean,
     val minSdk: Int?,
@@ -29,6 +30,7 @@ internal data class FidEnvSnapshot(
             "gs=${googleServices.toBit()}",
             "agp=${agpVersion.sanitized()}",
             "flag=${fidFlag.toBit()}",
+            "reg=${fidRegisterApi.toBit()}",
             "def=${defaultFirebaseApp.toBit()}",
             "prov=${firebaseInitProvider.toBit()}",
             "min=${minSdk?.toString() ?: "-"}",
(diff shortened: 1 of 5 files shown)


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/device/impl/FidEnv.kt
```diff
@@ -127,6 +130,31 @@ internal class AndroidFidEnvReader(
         }
     }
 
+    // 25.1+ exposes register(); older libraries ignore the metadata and keep getToken().
+    @Suppress("TooGenericExceptionCaught")
+    private fun hasFirebaseMessagingRegisterApi(): Boolean {
+        return try {
+            Class.forName(FIREBASE_MESSAGING).getMethod("register")
```
nan-li: This returns false in any minified build unless the app keeps `FirebaseMessaging` itself. With R8, ours comes from `Class.forName`, the lookup fails and we send `reg=0`. 

One line in core/consumer-rules.pro fixes it. Checked that it keeps `register()` with no caller and that R8 only logs an info when the class is missing:

```
-keepclassmembers class com.google.firebase.messaging.FirebaseMessaging { *** register(); }
```

Add a comment saying why, like location/consumer-rules.pro does for `GoogleApiClient`.

nan-li (changes requested):
#### Main blocking issue:
`reg=` can be 0 in minified builds, so the new bit may not work in production. Details inline on FidEnv.kt:137.

`def=` breaks the same way (FidEnv.kt:163). Same fix, one more rule in core/consumer-rules.pro:

```
-keepclassmembers class com.google.firebase.FirebaseApp { *** getApps(android.content.Context); *** getName(); }
```

#### Worth considering for this PR:

- The `reg=1` path is untested (FidEnvReaderTests.kt:54). Core tests have no `FirebaseMessaging` class, so both false assertions pass by accident. Take the class name as a constructor parameter and test both branches.
- `fidFlagFromMetadata` duplicates `AndroidUtils.getManifestMetaBoolean` (FidEnv.kt:145). Use the helper.

fadi-george:
Potential issues:
- `register()` is looked up by name, but no consumer ProGuard rule keeps it. R8 may rename or remove it, causing minified apps on FCM 25.1+ to incorrectly report `reg=0`.
- Tests only cover `reg=0`. Please add a fixture with a public zero-argument `register()` and verify `reg=1`.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/device/impl/FidEnv.kt
abdulraqeeb33: In d7cb811. Also kept `FirebaseApp.getApps`/`getName` for `def=`.

abdulraqeeb33:
In d7cb811.

1. Keep rules for `register()`, `getApps`, and `getName`.
2. True/false lookup tests added. Left the flag helper silent (`getManifestMetaBoolean` logs).

nan-li (approved):
LGTM, only an ask to update the PR description to reflect the final changes

PR #2734, opened 2026-09-03
fix: [SDK-4946] register with Firebase Installation ID when the legacy FCM token API is disabled

fadi-george:
# Description
## One Line Summary
Restore Firebase Installation ID registration when an app disables the legacy FCM token API.

### Motivation
Firebase Messaging 25.1.0 allows apps and dependencies to enable `firebase_messaging_installation_id_enabled`, which disables the legacy token API process-wide. OneSignal's `getToken()` then throws `IllegalStateException: API disabled`, preventing push registration.

This remakes #2700 after its temporary revert in #2719. It retains the original fallback and diagnostics, and now mirrors Firebase Messaging's sender-ID fallback for programmatically configured default Firebase apps.

### Scope
Only the FCM registration path changes when Firebase reports that the legacy API is disabled. Existing integrations continue using `getToken()`.

FID registration uses the host app's default `FirebaseApp`, because its sender ID, app ID, and API key belong to one Firebase project. Apps enabling FID registration must provide Firebase configuration whose sender ID matches the OneSignal dashboard.

`FirebaseMessaging.register()` remains reflective so the SDK can compile against Firebase Messaging 24.0.0 while consumers resolve 25.1.0 or newer. The consumer rule preserves this method through minification.

# Testing
## Unit testing
- Covers legacy-token and FID paths, diagnostics, registration failures, reflection behavior, and Firebase's application-ID sender fallback.
- Verifies successful wiring invokes registration and Firebase Installations through the host default `FirebaseApp`.
- `./gradlew spotlessCheck detekt :OneSignal:notifications:testDebugUnitTest :OneSignal:notifications:assembleRelease --console=plain`
- Focused FCM tests rerun after review fixes.

## Manual testing
The original implementation was manually verified on an Android 16 emulator with Firebase Messaging 25.1.1 in both debug and minified release builds, including delivery and legacy-token/FID conversion. This remake restores that implementation; the added sender fallback is covered by unit tests and was not re-tested on-device.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
@@ -88,9 +88,14 @@ internal class PushRegistratorFCM(
             FirebaseApp
                 .getApps(_applicationService.appContext)
                 .firstOrNull { it.name == FirebaseApp.DEFAULT_APP_NAME } ?: return null
+        val defaultSenderId =
+            FCMTokenProvider.defaultSenderId(
+                defaultApp.options.gcmSenderId,
+                defaultApp.options.applicationId,
+            )
 
         return FCMTokenProvider.InstallationIdRegistration(
-            senderId = defaultApp.options.gcmSenderId,
+            senderId = defaultSenderId,
             register = { FCMTokenProvider.invokeRegister(defaultApp.get(FirebaseMessaging::class.java)) },
             installationId = { FirebaseInstallations.getInstance(defaultApp).id },
         )
@@ -111,6 +116,16 @@ internal class PushRegistratorFCM(
 }
 
 internal object FCMTokenProvider {
+    fun defaultSenderId(
+        senderId: String?,
+        applicationId: String,
+    ): String? =
+        when {
+            senderId != null -> senderId
+            !applicationId.startsWith("1:") -> applicationId
+            else -> applicationId.split(":").getOrNull(1)?.ifEmpty { null }
+        }
+
     /**
      * The Firebase Installation ID registration that replaces the legacy token API, along with the
      * sender id of the Firebase project it would register against.
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
@@ -64,11 +64,11 @@ internal class PushRegistratorFCM(
     }
 
     private fun getLegacyToken(senderId: String): Task<String> {
-        initFirebaseApp(senderId)
-        // We use firebaseApp.get(FirebaseMessaging.class) instead of FirebaseMessaging.getInstance()
+        val app = initFirebaseApp(senderId)
+        // We use the named app's FirebaseMessaging instance instead of FirebaseMessaging.getInstance()
         //   as the latter uses the default Firebase app. We need to use a custom Firebase app as
         //   the senderId is provided at runtime.
-        return firebaseApp!!.get(FirebaseMessaging::class.java).token
+        return app.get(FirebaseMessaging::class.java).token
     }
 
     // Manifest merging means the flag can arrive from a dependency instead of the app's own
@@ -100,8 +100,8 @@ internal class PushRegistratorFCM(
         )
     }
 
-    private fun initFirebaseApp(senderId: String) {
-        if (firebaseApp != null) return
+    private fun initFirebaseApp(senderId: String): FirebaseApp {
+        firebaseApp?.let { return it }
         val firebaseOptions =
             FirebaseOptions
                 .Builder()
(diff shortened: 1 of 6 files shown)


OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -80,3 +114,115 @@ internal class PushRegistratorFCM(
...
+            register.invoke(target) as Task<*>
+        } catch (e: InvocationTargetException) {
+            throw e.targetException ?: e
+        }
+    }
+
+    private fun isLegacyTokenApiDisabled(exception: IllegalStateException): Boolean {
+        val message = exception.message ?: return false
+        return message.contains("API disabled") && message.contains("register()")
+    }
```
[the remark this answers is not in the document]
fadi-george: Fixed in a044b0d3e. The manifest flag plus register availability is now the primary signal. Exception-message matching remains only as a compatibility fallback.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -49,22 +55,50 @@ internal class PushRegistratorFCM(
...
-            return Tasks.await(tokenTask)
-        } catch (e: ExecutionException) {
-            throw tokenTask.exception ?: e
-        }
+        return FCMTokenProvider.getToken(
+            senderId,
+            ::installationIdEnabled,
+            { fcmInstance.token },
+            ::defaultAppRegistration,
+        )
```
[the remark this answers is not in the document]
fadi-george: Fixed in a044b0d3e. FID mode now checks the merged flag and register availability before evaluating the legacy-token path, so the named Firebase app is not initialized. The test verifies initializeApp is not called.

abdulraqeeb33 (commented):
Looks right. Flag off stays on the named app. FID only uses the host default FirebaseApp, and only when the sender matches.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -48,23 +54,50 @@ internal class PushRegistratorFCM(
...
         //   the senderId is provided at runtime.
-        val fcmInstance = firebaseApp!!.get(FirebaseMessaging::class.java)
-        // FirebaseMessaging.getToken API was introduced in firebase-messaging:21.0.0
-        val tokenTask = fcmInstance.token
-        try {
-            return Tasks.await(tokenTask)
-        } catch (e: ExecutionException) {
-            throw tokenTask.exception ?: e
-        }
+        return firebaseApp!!.get(FirebaseMessaging::class.java).token
```
abdulraqeeb33: can we avoid the !!?

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -80,3 +113,129 @@ internal class PushRegistratorFCM(
...
 }
+
+internal object FCMTokenProvider {
+    fun defaultSenderId(
+        senderId: String?,
+        applicationId: String,
+    ): String? =
+        when {
+            senderId != null -> senderId
+            !applicationId.startsWith("1:") -> applicationId
```
abdulraqeeb33, line 125: how sure are we that it always starts with 1:, do we have google documentation that proves it.

abdulraqeeb33 (commented):
nit and a question

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: Done in b862679de. initFirebaseApp now returns the non-null app, so getLegacyToken no longer needs `!!`.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: We do not assume it always starts with `1:`. This mirrors Firebase Messaging’s own `Metadata.getDefaultSenderId`: explicit `gcmSenderId` wins, non-v1 application IDs are returned unchanged, and only `1:<sender>:...` IDs are parsed. Source: https://github.com/firebase/firebase-android-sdk/blob/main/firebase-messaging/src/main/java/com/google/firebase/messaging/Metadata.java#L140-L161

abdulraqeeb33 (approved):

PR #2735, opened 2026-09-04
feat: [SDK-5170] register push before FCM credentials upload

fadi-george:
# Description
## One Line Summary
Register Android push subscriptions from the host Firebase configuration before dashboard FCM credentials are uploaded.

### Motivation
A newly configured Android app currently receives no push token or Firebase Installation ID until its FCM service account is uploaded to OneSignal and the app launches again. This change derives the sender ID from the host default Firebase app when dashboard parameters do not provide one, allowing the identifier to be uploaded during the first app session.

### Scope
- Preserves the dashboard sender ID as the primary source when configured.
- Falls back to the host default Firebase app sender ID when dashboard parameters omit it.
- Supports both legacy FCM token and Firebase Installation ID registration.
- Detects sender-ID mismatches between the host Firebase app and OneSignal configuration.
- Keeps legacy dashboard-provided sender registration available when no default Firebase app exists.
- Makes local demo Firebase configuration optional and ignored by Git.

### Review focus
- Whether legacy token registration should reject a host/dashboard sender mismatch even though it uses OneSignal's isolated Firebase app.
- Correct behavior if the effective sender ID changes during the same process after remote parameter hydration.

# Testing
## Unit testing
Added coverage for sender-ID derivation, registration before a dashboard sender exists, legacy-token and FID paths, missing default Firebase apps, mismatches, reflective registration, and failures.

Ran:
- Focused FCM registration unit tests
- Full `testDebugUnitTest`
- Spotless
- Detekt
- Release assembly

## Manual testing
Built the GMS demo with a local `google-services.json` and confirmed the generated APK packages `google_app_id` and `gcm_defaultSenderId`. Final end-to-end validation of credential upload activating an existing subscription remains pending because the emulator disconnected before the rebuilt APK could be installed.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [ ] Open
      - [x] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
@@ -51,6 +51,8 @@ internal abstract class PushRegistratorAbstractGoogle(
     @Throws(ExecutionException::class, InterruptedException::class, IOException::class)
     abstract suspend fun getToken(senderId: String): String
 
+    protected open fun resolveSenderId(configuredSenderId: String?): String? = configuredSenderId
+
     override suspend fun registerForPush(): IPushRegistrator.RegisterResult {
         if (!_configModelStore.model.isInitializedWithRemote) {
             return IPushRegistrator.RegisterResult(null, SubscriptionStatus.FIREBASE_FCM_INIT_ERROR)
@@ -61,7 +63,8 @@ internal abstract class PushRegistratorAbstractGoogle(
             return IPushRegistrator.RegisterResult(null, SubscriptionStatus.MISSING_FIREBASE_FCM_LIBRARY)
         }
 
-        return if (!isValidProjectNumber(_configModelStore.model.googleProjectNumber)) {
+        val senderId = resolveSenderId(_configModelStore.model.googleProjectNumber)
+        return if (!isValidProjectNumber(senderId)) {
             Logging.warn(
                 "Missing Google Project number!\nPlease enter a Google Project number / Sender ID on under App Settings > Android > Configuration on the OneSignal dashboard.",
             )
@@ -70,7 +73,7 @@ internal abstract class PushRegistratorAbstractGoogle(
                 SubscriptionStatus.INVALID_FCM_SENDER_ID,
             )
         } else {
-            internalRegisterForPush(_configModelStore.model.googleProjectNumber!!)
+            internalRegisterForPush(senderId!!)
         }
     }
 
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
@@ -134,6 +134,8 @@ internal abstract class PushRegistratorAbstractGoogle(
                 registrationId,
                 SubscriptionStatus.SUBSCRIBED,
             )
+        } catch (e: FCMSenderIdMismatchException) {
+            return invalidSenderIdResult(e)
         } catch (e: IOException) {
             val pushStatus: SubscriptionStatus = pushStatusFromThrowable(e)
             val exceptionMessage: String? = AndroidUtils.getRootCauseMessage(e)
@@ -168,6 +170,14 @@ internal abstract class PushRegistratorAbstractGoogle(
         }
 
         return null
+    }
+
+    private fun invalidSenderIdResult(exception: FCMSenderIdMismatchException): IPushRegistrator.RegisterResult {
+        Logging.warn("FCM sender ID mismatch", exception)
+        return IPushRegistrator.RegisterResult(
+            null,
+            SubscriptionStatus.INVALID_FCM_SENDER_ID,
+        )
     }
 
     private fun pushStatusFromThrowable(throwable: Throwable): SubscriptionStatus {
--- a/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
+++ b/OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
@@ -64,7 +64,7 @@ internal abstract class PushRegistratorAbstractGoogle(
         }
 
         val senderId = resolveSenderId(_configModelStore.model.googleProjectNumber)
-        return if (!isValidProjectNumber(senderId)) {
+        return if (senderId == null || !isValidProjectNumber(senderId)) {
             Logging.warn(
                 "Missing Google Project number!\nPlease enter a Google Project number / Sender ID on under App Settings > Android > Configuration on the OneSignal dashboard.",
             )
@@ -73,7 +73,7 @@ internal abstract class PushRegistratorAbstractGoogle(
                 SubscriptionStatus.INVALID_FCM_SENDER_ID,
             )
         } else {
-            internalRegisterForPush(senderId!!)
+            internalRegisterForPush(senderId)
         }
     }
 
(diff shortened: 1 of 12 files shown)


abdulraqeeb33 (changes requested):
Answer to the review focus: do not reject a host vs dashboard sender mismatch on the legacy path. That path still mints from `ONESIGNAL_SDK_FCM_APP_NAME`. Mixed-project apps are supposed to keep working there. 2734 left that alone.

1. Drop `validateSenderId` from `getLegacyToken`. Keep it on the FID path only. Invert the mixed-project legacy test.
2. Same change: rebuild the named app when `senderId` changes. Host fallback plus HYDRATE can change sender in-process. Cache-by-nullness then uploads the old sender's token.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -48,25 +54,65 @@ internal class PushRegistratorFCM(
...
+        configuredSenderId ?: defaultFirebaseApp()?.let(::defaultSenderId)
+
+    private fun getLegacyToken(
+        senderId: String,
+        defaultApp: FirebaseApp?,
+    ): Task<String> {
+        val defaultSenderId = defaultApp?.let(::defaultSenderId)
+        if (defaultSenderId != null) {
+            FCMTokenProvider.validateSenderId(senderId, defaultSenderId)
+        }
```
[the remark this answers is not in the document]
fadi-george: Fixed in 73b05afe5. Legacy registration no longer validates the host sender because it mints through the isolated OneSignal Firebase app. The FID path retains sender validation.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -48,25 +54,65 @@ internal class PushRegistratorFCM(
...
+            .firstOrNull { it.name == FirebaseApp.DEFAULT_APP_NAME }
+
+    private fun defaultSenderId(defaultApp: FirebaseApp): String? =
+        FCMTokenProvider.defaultSenderId(
+            defaultApp.options.gcmSenderId,
+            defaultApp.options.applicationId,
+        )
+
     private fun initFirebaseApp(senderId: String) {
         if (firebaseApp != null) return
```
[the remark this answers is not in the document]
fadi-george: Fixed in 73b05afe5. The named Firebase app now tracks its sender ID, deletes the old instance when the sender changes, and initializes a new app. Added a two-sender regression test.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
```diff
@@ -70,7 +73,7 @@ internal abstract class PushRegistratorAbstractGoogle(
                 SubscriptionStatus.INVALID_FCM_SENDER_ID,
             )
         } else {
-            internalRegisterForPush(_configModelStore.model.googleProjectNumber!!)
+            internalRegisterForPush(senderId!!)
```
abdulraqeeb33: can we avoid the !!?

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -82,26 +87,32 @@ internal class PushRegistratorFCM(
...
             installationId = { FirebaseInstallations.getInstance(defaultApp).id },
         )
     }
 
+    private fun defaultFirebaseApp(): FirebaseApp? =
+        FirebaseApp
+            .getApps(_applicationService.appContext)
+            .firstOrNull { it.name == FirebaseApp.DEFAULT_APP_NAME }
+
+    private fun defaultSenderId(defaultApp: FirebaseApp): String? =
```
abdulraqeeb33: can we rename these more accurately? defaultSenderId - is this the one that the backend is sending?

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -178,20 +192,24 @@ internal object FCMTokenProvider {
...
     }
 
+    fun validateSenderId(
+        senderId: String,
+        defaultSenderId: String?,
+    ) {
+        if (defaultSenderId == senderId) return
+
+        throw IllegalStateException(
+            "The default FirebaseApp uses sender id $defaultSenderId, while OneSignal is " +
```
abdulraqeeb33: this is great. this handles the condition we discussed earlier today

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
```diff
@@ -82,26 +87,32 @@ internal class PushRegistratorFCM(
...
+            .getApps(_applicationService.appContext)
+            .firstOrNull { it.name == FirebaseApp.DEFAULT_APP_NAME }
+
+    private fun defaultSenderId(defaultApp: FirebaseApp): String? =
+        FCMTokenProvider.defaultSenderId(
+            defaultApp.options.gcmSenderId,
+            defaultApp.options.applicationId,
+        )
+
     private fun initFirebaseApp(senderId: String): FirebaseApp {
```
abdulraqeeb33: can we update the variable name here? is this the remoteSenderId or clientSenderId?

abdulraqeeb33 (commented):
a few nits but otherwise looking good

abdulraqeeb33:
Quick question on the path after the first session.

No `android_sender_id` on setup is expected, host fallback is what we want there. Once the service account is uploaded and params come back with a sender that does not match the host, `validateSenderId` throws and we surface that as `MISC_EXCEPTION`. That status is retryable, so the host FID stays `SUBSCRIBED`.

Should that mismatch be `INVALID_FCM_SENDER_ID` instead, so we do not keep the host identifier as healthy?

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
fadi-george: Done. The null case is checked explicitly, so Kotlin smart-casts `senderId` before registration. I also simplified the validator to use `toFloatOrNull()`.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: renamed

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorFCM.kt
fadi-george: It can come from either the dashboard or the host Firebase app fallback, so I renamed it to `resolvedSenderId`.

fadi-george:
Agreed. An FID sender mismatch now maps to `INVALID_FCM_SENDER_ID`. Permanent configuration errors can also replace a previously successful in-memory token status, so the host FID is not kept healthy. The legacy mixed-project path remains unchanged.

OneSignalSDK/onesignal/notifications/src/main/java/com/onesignal/notifications/internal/registration/impl/PushRegistratorAbstractGoogle.kt
```diff
@@ -180,21 +193,7 @@ internal abstract class PushRegistratorAbstractGoogle(
...
-                false
-            }
-
-        if (!isProjectNumberValidFormat) {
-            return false
-        }
-
-        return true
-    }
+    private fun isValidProjectNumber(senderId: String): Boolean = senderId.toFloatOrNull() != null
```
abdulraqeeb33, line 196: 👍

abdulraqeeb33 (approved):
Thank you for addressing the comments @fadi-george . Looks good.

PR #2742, opened 2026-09-08
feat: [SDK-5180] expose notification click event JSON

fadi-george:
# Description
## One Line Summary
Expose notification click events as JSON through the public Android SDK API.

### Motivation
`NotificationClickEvent` already had internal JSON serialization, but `INotificationClickEvent` did not expose it. Applications therefore received the default object identity when logging the event and had to duplicate serialization logic.

### Scope
- Adds `toJSONObject()` to `INotificationClickEvent`.
- Returns the wrapper-compatible `notification` and `result` event shape.
- Serializes notification `rawPayload` as a JSON object, including grouped notifications.
- Updates the demo to use the base SDK serializer and removes its duplicate helper.
- Changes the prior internal `action` key to `result` and the serialized `rawPayload` type from string to object.
- Adds an abstract member to the public event interface; external custom implementations must implement it.

# Testing
## Unit testing
Added notification click serialization tests covering notification fields, grouped notifications, action ID, URL, nullable fields, and nested raw payload objects.

## Manual testing
No device testing was performed. The targeted unit tests, Spotless checks, IDE lint diagnostics, and notifications release build passed.

# Affected code checklist
   - [x] Notifications
      - [ ] Display
      - [x] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


abdulraqeeb33 (approved):
Approved.

Stay native for KMP. This serializer sits on Android `org.json` and Android-only notification fields (`androidNotificationId`, icons, LED, lock-screen visibility). `:kmp` is still logger-only, and there is no iOS sibling. Revisit when the notifications model itself moves.

One non-blocking follow-up: `JSONObject(rawPayload)` throws when `rawPayload` is empty (`initPayloadData` returns early if `custom` is missing). The existing catch then omits the key. Happy-path OneSignal clicks always have a valid payload string.

PR #2743, opened 2026-09-09
feat: [SDK-5024] add composite login(externalId, profile) as one Create User POST

abdulraqeeb33:
# Description
## One Line Summary
Add `login(externalId, profile)` so email, SMS, tags, and aliases go out on a single Create User upsert, and return HTTP 4xx on the waiting `OneSignalResult`.

### Motivation
Identity-only `login(externalId)` cannot attach profile fields. Composite login needs those fields on one Create User POST (upsert by `external_id`) so a bad SMS/email cannot freeze the op repo, and the suspend caller gets the backend body instead of a generic "did not complete".

Stacked on #2710 (`ar/sdk-4988`). Parent: [SDK-5023](https://linear.app/onesignal/issue/SDK-5023). Types/signature: [SDK-5024](https://linear.app/onesignal/issue/SDK-5024). One POST: [SDK-5025](https://linear.app/onesignal/issue/SDK-5025).

### Scope
- New `OneSignalUserProfile` (`email`, `phoneNumber` in E.164, `tags`, `aliases`) plus Java `Builder`. JWT stays a separate argument. `externalId` stays the login key.
- Identity-only `login` / `loginSuspend` are unchanged.
- Same `externalId` with no profile fields is still a no-op. Same `externalId` with profile fields upserts.
- Composite 400/409 drop the op (`FAIL_NORETRY`) instead of pausing the repo. Identity-only 4xx still pause.
- `enqueueAndAwaitResult` carries HTTP status/body so composite `login` returns `OneSignalResult.failure(BACKEND_ERROR, message=body, backendCode=HTTP status)`. Catalog codes are not parsed yet.
- Demo: LOGIN WITH PROFILE dialog.

Not in this PR: persistence ([SDK-5026](https://linear.app/onesignal/issue/SDK-5026)), telemetry ([SDK-5027](https://linear.app/onesignal/issue/SDK-5027)), a login callback, catalog `backendCode`.

# Testing
## Unit testing
Executor tests cover identity-only payload unchanged, profile aliases/email/SMS/tags on Create User, reserved alias skip, 400/409 `FAIL_NORETRY` with HTTP fields, 401, and hydration. OperationRepo wakes `enqueueAndAwaitResult` with status/body. LoginHelper, UserBackendService nested tags, and profile builder tests added.

## Manual testing
Demo app on a physical device. Composite login with a non-E.164 SMS (`4129089471`) returns HTTP 400 `Invalid token format for device type SMS`, logs the body, returns `OneSignalResult.failure`, and does not pause the op repo. Retry with `+14129089471` succeeds.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [x] REST API requests
   - [x] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignalUserProfile.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignalUserProfile.kt
@@ -1,0 +1,41 @@
+package com.onesignal
+
+/**
+ * Profile fields applied at composite login. Email and SMS are additive subscriptions, not replaced traits.
+ */
+data class OneSignalUserProfile(
+    val email: String? = null,
+    /** SMS number in [E.164](https://documentation.onesignal.com/docs/sms-faq#what-is-the-e164-format) format. */
+    val phoneNumber: String? = null,
+    val tags: Map<String, String> = emptyMap(),
+    val aliases: Map<String, String> = emptyMap(),
+) {
+    /** Java builder. Use this instead of the constructor when only some fields are set. */
+    class Builder {
+        private var email: String? = null
+        private var phoneNumber: String? = null
+        private var tags: Map<String, String> = emptyMap()
+        private var aliases: Map<String, String> = emptyMap()
+
+        /** Sets the email subscription to create at login. */
+        fun setEmail(email: String?) = apply { this.email = email }
+
+        /** Sets the SMS subscription to create at login. Must be E.164 (e.g. +14155552671). */
+        fun setPhoneNumber(phoneNumber: String?) = apply { this.phoneNumber = phoneNumber }
+
+        /** Sets tags to apply at login. The map is copied. */
+        fun setTags(tags: Map<String, String>) = apply { this.tags = tags.toMap() }
+
+        /** Sets aliases to apply at login. The map is copied. */
+        fun setAliases(aliases: Map<String, String>) = apply { this.aliases = aliases.toMap() }
+
+        /** Builds the profile. */
+        fun build() = OneSignalUserProfile(email, phoneNumber, tags, aliases)
+    }
+
+    companion object {
+        /** Returns a new Java builder. */
+        @JvmStatic
+        fun builder() = Builder()
+    }
+}
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignalUserProfile.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignalUserProfile.kt
@@ -10,6 +10,13 @@ data class OneSignalUserProfile(
     val tags: Map<String, String> = emptyMap(),
     val aliases: Map<String, String> = emptyMap(),
 ) {
+    internal val hasFields: Boolean
+        get() =
+            !email.isNullOrBlank() ||
+                !phoneNumber.isNullOrBlank() ||
+                tags.isNotEmpty() ||
+                aliases.isNotEmpty()
+
     /** Java builder. Use this instead of the constructor when only some fields are set. */
     class Builder {
         private var email: String? = null
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
@@ -393,7 +393,7 @@ internal class OperationRepo(
                 ExecutionResult.SUCCESS -> {
                     // on success we remove the operation from the store and wake any waiters
                     ops.forEach { _operationModelStore.remove(it.operation.id) }
-                    ops.forEach { it.waiter?.wake(waitResult(true, response)) }
+                    ops.forEach { it.waiter?.wake(response.toWaitResult(true)) }
                 }
                 ExecutionResult.FAIL_UNAUTHORIZED -> {
                     // Outer gate: dispatch to IV extension only on new code paths.
@@ -422,7 +422,7 @@ internal class OperationRepo(
                     // remove the starting operation from the store and wake any waiters, then
                     // add back all but the starting op to the front of the queue to be re-executed
                     _operationModelStore.remove(startingOp.operation.id)
-                    startingOp.waiter?.wake(waitResult(true, response))
+                    startingOp.waiter?.wake(response.toWaitResult(true))
                     synchronized(queue) {
                         ops.filter { it != startingOp }.reversed().forEach { queue.add(0, it) }
                     }
@@ -444,7 +444,7 @@ internal class OperationRepo(
                     // keep the failed operation and pause the operation repo from executing
                     paused = true
                     // Unblock any enqueueAndWait callers so loginSuspend doesn't hang.
-                    ops.forEach { it.waiter?.wake(waitResult(false, response)) }
+                    ops.forEach { it.waiter?.wake(response.toWaitResult(false)) }
                     // Re-queue with waiter = null: the operation is preserved for retry
                     // on next cold start, but the original waiter is detached since it
                     // was already woken above.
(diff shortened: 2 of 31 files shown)


examples/demo/app/src/main/java/com/onesignal/example/ui/main/Sections.kt
```diff
@@ -176,6 +179,13 @@ fun UserSection(
             testTag = "login_user_button",
         )
 
+        OutlineButton(
+            text = "LOGIN WITH PROFILE",
```
abdulraqeeb33, line 183: this needs to be added to iOS and all the other wrappers

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/operations/impl/OperationRepo.kt
```diff
@@ -222,11 +224,12 @@ internal class OperationRepo(
                     } else {
                         Logging.debug("OperationRepo: internalEnqueue - LoginUserOperation for onesignalId: ${op.onesignalId} already exists in the queue.")
                     }
+                    existingOp.mergeProfileFrom(op)
                     // Transfer the waiter so enqueueAndWait callers see the queued op's real execution result.
                     if (queueItem.waiter != null && existing.waiter == null) {
                         existing.waiter = queueItem.waiter
                     } else {
-                        queueItem.waiter?.wake(true)
+                        queueItem.waiter?.wake(OperationWaitResult(true))
```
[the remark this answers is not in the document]
abdulraqeeb33: Fixed in 7e8eb2f75.

Incoming waiters attach to the queued op. Both wake with the real execution result.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/LoginUserOperation.kt
```diff
@@ -43,17 +47,63 @@ class LoginUserOperation() : Operation(LoginUserOperationExecutor.LOGIN_USER) {
...
+
+    internal fun hasProfileFields(): Boolean =
+        !email.isNullOrBlank() || !phoneNumber.isNullOrBlank() || tags.isNotEmpty() || aliases.isNotEmpty()
+
+    internal fun mergeProfileFrom(other: LoginUserOperation) {
+        if (email.isNullOrEmpty()) email = other.email
+        if (phoneNumber.isNullOrEmpty()) phoneNumber = other.phoneNumber
+        if (other.tags.isNotEmpty()) tags = tags + other.tags
+        if (other.aliases.isNotEmpty()) aliases = aliases + other.aliases
     }
```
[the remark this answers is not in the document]
abdulraqeeb33: Fixed in 7e8eb2f75.

mergeProfileFrom last-write-wins for email and phone.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/LoginHelper.kt
```diff
@@ -77,19 +79,32 @@ internal class LoginHelper(
...
+        return result
     }
+
+    internal fun contextForCurrentUser(externalId: String): LoginEnqueueContext =
+        LoginEnqueueContext(
+            appId = configModel.appId,
+            newIdentityOneSignalId = identityModelStore.model.onesignalId,
+            externalId = externalId,
+            existingOneSignalId = null,
+        )
```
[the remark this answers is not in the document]
abdulraqeeb33: Fixed in 7e8eb2f75.

switchUser snapshots the enqueue context under the lock. LoginData uses wait.metadata, not live stores after await.

fadi-george:
Potential issues:
- Composite 4xx failures drop grouped push subscription operations, so retries may create users without device subscriptions.
- Login can hang after the operation repo pauses because new waiters attach to an operation that will not execute.
- Concurrent composite logins merge conflicting profiles and return unrelated subscription IDs.
- Repeated identity-only login can issue a duplicate Create User request.
- `core.api` is stale and needs regeneration.
- Spotless currently fails on import ordering.

examples/demo/app/src/main/java/com/onesignal/example/data/repository/OneSignalRepository.kt
```diff
@@ -25,6 +26,19 @@ class OneSignalRepository {
         Log.d(TAG, "Logged in user with onesignalId: ${OneSignal.User.onesignalId}")
     }
 
+    suspend fun loginUserWithProfile(
```
fadi-george, line 29: maybe should be loginWithProfile? maybe could vote on it
not blocking

abdulraqeeb33:
On 3: leaving waiters on the shared result.

We want composite login to match the existing login methods (one Create User per onesignalId, attached waiters get that result) and not add a per-caller profile path.

fadi-george:
still a potential issue:
inFlightLogins remains populated during the 5-second post-create delay, but waiters are woken before that delay. A duplicate login during this window attaches too late and can hang forever. The new test only covers duplicates before execution completes.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/OneSignalUserProfile.kt
```diff
@@ -0,0 +1,74 @@
+package com.onesignal
+
+/**
+ * Profile fields applied at composite login. Email and SMS are additive subscriptions, not replaced traits.
+ */
+class OneSignalUserProfile @JvmOverloads constructor(
+    email: String? = null,
+    phoneNumber: String? = null,
```
nan-li, line 8: Since our API is currently `OneSignal.User.addSms(sms: String)`, we should be consistent and maybe do `sms` or `smsNumber` instead of phoneNumber?

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/impl/executors/LoginUserOperationExecutor.kt
```diff
@@ -222,14 +236,33 @@ internal class LoginUserOperationExecutor(
...
+                        when (LoginProfileApplier.subscriptionType(backendSubscription.type)) {
+                            SubscriptionType.EMAIL -> emailSubscriptionId = backendSubscription.id
+                            SubscriptionType.SMS -> smsSubscriptionId = backendSubscription.id
+                            else -> {}
+                        }
+                    } else {
+                        Logging.error("LoginUserOperationExecutor.createUser response is missing a local subscription model for ${pair.first}")
+                    }
                 } else {
                     Logging.error("LoginUserOperationExecutor.createUser response is missing subscription data for ${pair.first}")
```
nan-li, line 267: normally in the past this would fire for a genuinely missing push sub but now with the profile including email and/or sms regularly, this will log an error with __profile_email or __profile_sms if that email/sms already belonged on this user. Similar to SDK-5181

nan-li (commented):
Fadi's post-create point is still open, and I reproduced the 4xx case on a device.

1. Second login hangs (Fadi's point).

2. Profile silently dropped. A `login(id, profile)` that arrives while a plain `login(id)` for the same user is in flight reuses that result, so the email and phone are never sent.

3. User stuck after a 400. The login is dropped, but the push subscription and anything added afterwards stay queued under a local id that is never resolved, so nothing for this user is sent again. Reproduced by calling login("1", profile).

4. Anonymous user no longer merged. With a profile attached, a new external id from an anonymous user creates a second user and leaves the anonymous one behind with its tags. May need to be called in our docs and PR as behavior change.

5. Repeat logins with the same email and phone. The backend omits giving ID for unchanged subscriptions, so a repeat login logs an error. It is common for a client to always login with the profile they have on hand for the user, which will trigger this error log often. Reproduced by calling `login(user, existing email or sms)`

PR #2744, opened 2026-09-10
feat: [SDK-5193] let an app opt out of subscription ID copy in its manifest

nan-li:
# Description
## One Line Summary
Lets an app turn off subscription ID copy with a manifest key.

### Motivation
The gesture in #2727 is on by default, and the only switch is OneSignal's remote kill switch. An app that doesn't want a clipboard write in production should be able to say so itself, in its own manifest, without a support ticket. 

Ticket SDK-5193. Stacked on #2727, which this PR targets.

### Scope
`com.onesignal.subscriptionIdCopyDisabled` set to `true` in the app's manifest meta-data

- read in `DeviceGestureDetector.start()` through `AndroidUtils.getManifestMetaBoolean`, the same way `com.onesignal.preferHMS` is read. 
- With the key set the detector never registers its lifecycle handler, so nothing is counted, copied or recorded. 
- That is stricter than the remote kill switch, which still records `disabled` so we can count attempts; an app that opted out has said no. 
- One INFO line at start says the key was honoured, so a developer can confirm it took.

The key names the outcome, not the gesture, so the docs can call the feature one thing. No public API, so the wrapper SDKs expose it with no change.

### Other
The docs page and release notes need to say the feature is on by default and name the key.

# Testing
## Unit testing
One new test in `DeviceGestureDetectorTests` mocks the manifest read the way `DeviceServiceTests` does and asserts no lifecycle handler, no clip, no event. 22 tests, Spotless and detekt pass.

## Manual testing
Not run on a device. The read goes through the helper every other manifest switch uses.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass

### DIFF
--- a/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/gesture/DeviceGestureDetector.kt
+++ b/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/gesture/DeviceGestureDetector.kt
@@ -4,6 +4,7 @@ import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
 import android.os.SystemClock
+import com.onesignal.common.AndroidUtils
 import com.onesignal.common.IDManager
 import com.onesignal.core.internal.application.IApplicationLifecycleHandler
 import com.onesignal.core.internal.application.IApplicationService
@@ -28,7 +29,9 @@ import com.onesignal.logger.ObservabilityEvent
  * only rate rule; six cycles fit inside it at round trips of five seconds or faster.
  *
  * [FeatureFlag.SDK_DEVICE_GESTURE_DISABLED] turns the gesture off. Absent means enabled, so a
- * device that has never fetched flags still has it.
+ * device that has never fetched flags still has it. An app can also opt out for good with the
+ * manifest meta-data [MANIFEST_DISABLED_KEY]; then the detector never starts, so nothing is
+ * counted or recorded.
  *
  * Every recognised gesture also records [ObservabilityEvent.DEVICE_GESTURE], with its outcome
  * and the copied ID, so the gesture's usage can be measured.
@@ -53,6 +56,11 @@ internal class DeviceGestureDetector(
     private val cycleTimestamps = mutableListOf<Long>()
 
     override fun start() {
+        // The app owner's opt-out. Logged so a developer can confirm the key took.
+        if (AndroidUtils.getManifestMetaBoolean(applicationService.appContext, MANIFEST_DISABLED_KEY)) {
+            Logging.info("DeviceGestureDetector: disabled by $MANIFEST_DISABLED_KEY in the manifest, not starting")
+            return
+        }
         applicationService.addApplicationLifecycleHandler(this)
     }
 
@@ -180,6 +188,7 @@ internal class DeviceGestureDetector(
         /** Shortest background phase a human can produce; anything faster is synthetic. */
         internal const val MIN_BACKGROUND_DWELL_MS = 250L
 
+        internal const val MANIFEST_DISABLED_KEY = "com.onesignal.subscriptionIdCopyDisabled"
         private const val CLIP_LABEL = "OneSignal subscription ID"
         private const val CLIP_PREFIX = "os: "
         internal const val NO_SUBSCRIPTION_CLIP_TEXT = CLIP_PREFIX + "no subscription ID yet"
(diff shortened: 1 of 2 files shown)


abdulraqeeb33 (commented):
The opt-out matches the ticket: `com.onesignal.subscriptionIdCopyDisabled=true` returns from `start()` before the lifecycle handler, so nothing is counted, copied, or recorded. That is stricter than the remote kill switch, as intended.

Two nits on how we pin and confirm the key (inline).

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/gesture/DeviceGestureDetectorTests.kt
```diff
@@ -0,0 +1,379 @@
...
+
+    // ===== Boundaries and failure paths =====
+
+    test("the manifest opt-out keeps the detector from starting") {
+        // The app owner said no: no lifecycle handler is registered, so nothing is counted,
+        // copied or recorded, and the only trace is one INFO line at start.
+        mockkObject(AndroidUtils)
+        try {
+            every {
+                AndroidUtils.getManifestMetaBoolean(any(), DeviceGestureDetector.MANIFEST_DISABLED_KEY)
```
abdulraqeeb33: The stub uses `MANIFEST_DISABLED_KEY`, the same constant `start()` reads, so a typo in `"com.onesignal.subscriptionIdCopyDisabled"` would still leave this test green. That string is the whole customer contract (no public API, docs not in this PR). `DeviceServiceTests` hardcodes `"com.onesignal.preferHMS"` for the same reason. Worth pinning the literal here.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/gesture/DeviceGestureDetector.kt
```diff
@@ -0,0 +1,204 @@
...
+     */
+    internal var monotonicMillis: () -> Long = { SystemClock.elapsedRealtime() }
+
+    private var lastUnfocusedAt: Long? = null
+    private val cycleTimestamps = mutableListOf<Long>()
+
+    override fun start() {
+        // The app owner's opt-out. Logged so a developer can confirm the key took.
+        if (AndroidUtils.getManifestMetaBoolean(applicationService.appContext, MANIFEST_DISABLED_KEY)) {
+            Logging.info("DeviceGestureDetector: disabled by $MANIFEST_DISABLED_KEY in the manifest, not starting")
```
abdulraqeeb33, line 61: Default `Logging.logLevel` is WARN, and `logToLogcat` drops INFO. At default logcat both "key honored" and "key typo'd" are silent, so this line cannot be the confirmation the commit describes. `ILogListener`s still get it. The successful copy already uses INFO, and iOS uses `LL_INFO`, so this may be intentional. If the confirmation is load-bearing, WARN would show once per process at the default level.

abdulraqeeb33 (approved):
Approving with the two nits already on the diff. Neither blocks the opt-out.

fadi-george (approved):
Same as AR's nits

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/gesture/DeviceGestureDetectorTests.kt
nan-li: Done in 7471112e9. The stub spells out com.onesignal.subscriptionIdCopyDisabled and the constant is private now, since nothing else reads it. I typo'd the constant to check. The stub misses, the real read finds no meta-data, the handler gets added, and only this test fails.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/gesture/DeviceGestureDetector.kt
nan-li: Right, at the default level it never prints, it is while someone is testing their app etc. I kept INFO, this is consistent with other informational logs in the SDK. An honoured opt-out is a healthy state, not a warning, and the other manifest switches log nothing at all when they take. Just updated the comment no longer claims the line is a confirmation, and I dropped that sentence from the description.

nan-li:
> Same as AR's nits

thanks addressed

PR #2746, opened 2026-09-14
fix: [SDK-5204] expose FID registration statuses

fadi-george:
# Description
## One Line Summary
Expose three backend-visible Firebase Installation ID registration failure statuses.

### Motivation
Support and dashboard surfaces need to distinguish a missing default Firebase app, an unavailable `FirebaseMessaging.register()` API, and a generic FID runtime registration failure.

### Scope
- Allocates `-32`, `-33`, and `-34` in the Android subscription status enum.
- Maps the three typed FID failures without changing sender mismatch or FCM `IOException` handling.
- Preserves healthy cached identifiers for all three failures.
- Safely hydrates unknown non-positive backend statuses as a disabled generic registration error instead of subscribed.

The SDK remains subject to the backend-first rollout constraint: these emitting statuses must not ship before schema allocation and backend acceptance/storage support are deployed.

# Testing
## Unit testing
Adds status allocation/hydration coverage and registration/retention coverage for all three FID failures. PR-caused CI failures—a generic task type-inference error and outdated exhaustive retryable-status expectations—are fixed in `44310a446` and `a9faaa79f`.

Targeted Gradle tests cannot start in the Cloud Agent environment because no Android SDK is installed or configured (`ANDROID_HOME` and `sdk.dir` are absent). CI runs them in the repository's Android environment.

## Manual testing
Not run; these error paths are covered with deterministic unit tests.

# Affected code checklist
- [x] Notifications
  - [x] Push Processing
- [x] REST API requests
- [ ] Public API changes

# Checklist
## Overview
- [x] I have filled out all **REQUIRED** sections above
- [x] PR does one thing
- [x] Any Public API changes are explained in the PR details and conform to existing APIs

## Testing
- [x] I have included test coverage for these changes
- [ ] All automated tests pass — local run blocked by missing Android SDK; CI rerun pending
- [x] I have personally tested this on my device, or explained why that is not possible

## Final pass
- [x] Code is as readable as possible
- [x] I have reviewed this PR myself

[...]

PR #2750, opened 2026-09-15
fix: [SDK-5206] expose FID registration statuses

fadi-george:
# Description
## One Line Summary
Expose three backend-visible Firebase Installation ID registration failure statuses.

### Motivation
Support and dashboard surfaces need to distinguish a missing default Firebase app, an unavailable `FirebaseMessaging.register()` API, and a generic FID runtime registration failure.

### Scope
- Allocates `-32`, `-33`, and `-34` in the Android subscription status enum.
- Maps the three typed FID failures without changing sender mismatch or FCM `IOException` handling.
- Preserves healthy cached identifiers for all three failures.
- Safely hydrates unknown non-positive backend statuses as a disabled generic registration error instead of subscribed.

The SDK remains subject to the backend-first rollout constraint: these emitting statuses must not ship before schema allocation and backend acceptance/storage support are deployed.

# Testing
## Unit testing
Adds status allocation/hydration coverage and registration/retention coverage for all three FID failures. The replacement branch has the same tree as #2746, where code quality and test checks passed.

## Manual testing
Not run; these error paths are covered with deterministic unit tests.

# Affected code checklist
- [x] Notifications
  - [x] Push Processing
- [x] REST API requests
- [ ] Public API changes

# Checklist
## Overview
- [x] PR does one thing

## Testing

## Final pass


sherwinski (commented):
Three should-fix items, all inline on `SubscriptionModel.kt`. Nothing blocking.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModel.kt
```diff
@@ -135,6 +145,19 @@ enum class SubscriptionStatus(val value: Int) {
...
+
+        /**
+         * Converts the backend notification-types wire value without allowing an unknown disabled
+         * state to become [SUBSCRIBED]. Older SDKs can receive statuses introduced by a newer
+         * backend, so unknown non-positive values use the existing generic registration failure.
+         */
+        fun fromNotificationTypes(value: Int): SubscriptionStatus =
+            fromInt(value)
+                ?: if (value <= NO_PERMISSION.value) {
+                    FIREBASE_FCM_ERROR_MISC_EXCEPTION
```
sherwinski: The fallback for an unknown non-positive code is `FIREBASE_FCM_ERROR_MISC_EXCEPTION` (-12). This helper also runs for email and SMS rows in `RefreshUserOperationExecutor`. If the server returns an unknown negative code on an email subscription, the model gets an FCM exception status. The next update op then echoes `-12` back to the server through `getSubscriptionEnabledAndStatus`. That misreports the cause on a non-push channel.

`ERROR(9999)` is documented as "some other (unknown locally) error" and matches the intent here. If the backend rejects 9999 on the wire, keep -12, but say why in the doc comment. The current comment does not explain why an FCM code was chosen.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModel.kt
```diff
@@ -102,6 +111,7 @@ enum class SubscriptionStatus(val value: Int) {
                 HMS_TOKEN_TIMEOUT, // -25
                 HMS_API_EXCEPTION_OTHER, // -27
                 FIREBASE_FCM_ERROR_IOEXCEPTION_AUTHENTICATION_FAILED, // -29
+                FIREBASE_FCM_FID_REGISTRATION_FAILED, // -34
```
sherwinski, line 114: `FIREBASE_FCM_FID_REGISTRATION_FAILED` is retryable here, but the exception it maps from is raised for every non-IO `Exception` out of `register()` or `installationId()` (`retrieveInstallationId` in `PushRegistratorFCM.kt`). Its message tells the developer to check `google-services.json` and the Gradle plugin, which describes a permanent configuration error. Firebase Installations also throws `BAD_CONFIG` for a wrong API key, which will not recover on retry. The classification and the message contradict each other.

Two options:

1. Split the reason by cause. Keep `REGISTRATION_FAILED` retryable for `UNAVAILABLE` and `TOO_MANY_REQUESTS`, and add a non-retryable config variant.
2. Keep one code, but update the message and the doc comment on line 82 to say "this can be retried". Every other retryable status says so in its doc comment.

Note that `attemptRegistration` returns on `FCMInstallationIdException` at once, with no in-process backoff. "Retryable" here only means "retry at the next session start".

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModel.kt
```diff
@@ -135,6 +145,19 @@ enum class SubscriptionStatus(val value: Int) {
...
         }
+
+        /**
+         * Converts the backend notification-types wire value without allowing an unknown disabled
+         * state to become [SUBSCRIBED]. Older SDKs can receive statuses introduced by a newer
+         * backend, so unknown non-positive values use the existing generic registration failure.
+         */
+        fun fromNotificationTypes(value: Int): SubscriptionStatus =
+            fromInt(value)
+                ?: if (value <= NO_PERMISSION.value) {
```
sherwinski: `value <= NO_PERMISSION.value` reads as "less than or equal to no permission". `PushTokenManager.retrievePushToken` already partitions on `value < SUBSCRIBED.value`. Use the same form so both sites read the same way.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModel.kt
fadi-george: Updated the fallback to the channel-neutral ERROR(9999), so unknown server statuses are not reported as FCM failures on email or SMS subscriptions.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModel.kt
fadi-george: Kept the allocated -34 status and clarified in both the enum documentation and diagnostic message that it is retryable on the next session. Persistent failures still direct developers to verify Firebase configuration.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/subscriptions/SubscriptionModel.kt
fadi-george: Changed the partition to value < SUBSCRIBED.value to match the existing push-token logic.

sherwinski (approved):

PR #2753, opened 2026-09-17
fix: [SDK-5202] canonicalize Chinese locales and make setLanguage("") use device default

fadi-george:
# Description
## One Line Summary
Canonicalize Chinese device locales, restore device-default language reset behavior, and add complete language selection to the Android demo.

Screenshot 2026-09-17 at 11 51 19 AM
Screenshot 2026-09-17 at 12 22 45 PM

### Motivation
Android previously reduced Chinese locales to language plus region, such as `zh-CN` or `zh-HK`, discarding an explicit `Hans` or `Hant` script. IAM language variants use script-based codes, so script and region conflicts could select the wrong translation through server-side regional aliases.

The documented `setLanguage("")` device-default behavior also retained an empty override instead of clearing it. The demo had no direct way to select and exercise supported language codes.

### Scope
- Resolve automatic Chinese device languages to the supported `zh-Hans` and `zh-Hant` codes.
- Prefer explicit script subtags, with a region fallback for locales that do not provide a script.
- Treat an empty language value as clearing the explicit override so device detection is used again.
- Add a demo language dropdown containing Device Default and every supported OneSignal language.

No public API signatures change.

# Testing
## Unit testing
Added coverage for explicit Chinese script precedence, region fallback, ambiguous Chinese locales, explicit language overrides, and clearing an override with an empty value.

Validated with:
- `./gradlew :OneSignal:core:testDebugUnitTest --tests 'com.onesignal.core.internal.language.impl.LanguageContextTests' --tests 'com.onesignal.core.internal.language.impl.LanguageProviderDeviceTests'`
- `./gradlew :OneSignal:core:spotlessKotlinCheck`

## Manual testing
Built and installed the local SDK demo on an Android emulator. Verified OneSignal initializes successfully, the app remains running without a crash, and Chinese segment-based IAM targeting works with the emulator device language.

A production IAM-service compatibility shim aliases regional Chinese codes, so an end-to-end branch difference for true multi-language IAM variants cannot be isolated against production.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [x] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/language/impl/LanguageContext.kt
```diff
@@ -6,11 +6,11 @@ import com.onesignal.user.internal.properties.PropertiesModelStore
...
     private val _propertiesModelStore: PropertiesModelStore,
 ) : ILanguageContext {
-    private var deviceLanguageProvider = LanguageProviderDevice()
+    private val deviceLanguageProvider = LanguageProviderDevice()
 
     override var language: String
         get() = _propertiesModelStore.model.language ?: deviceLanguageProvider.language
         set(value) {
-            _propertiesModelStore.model.language = value
+            _propertiesModelStore.model.language = value.ifEmpty { null }
```
nan-li: Was there a reason this change came in this PR? a `null` doesn't propagate up to the server since it is dropped from the request body. We have a kdoc over `setLanguage` that says this, so maybe if we touch this, it should be `_propertiesModelStore.model.language = value.ifEmpty { deviceLanguageProvider.language }`?

```
/**
 * Set the 2-character language either as a detected language or explicitly set for this user. See
 * See [Supported Languages | OneSignal](https://documentation.onesignal.com/docs/language-localization#what-languages-are-supported)
 *
 * @param value The 2-character language string, or an empty string to use the device default.
 */
```

nan-li (commented):
A couple of followups to this:

1. The KDoc on `IUserManager.setLanguage` still says "2-character language", which this PR's own zh-Hans output allows so we can update that comment, and maybe it should say what the empty string does.

2. Note that iOS rejects an empty string passed to `setLanguage` API with an error log, so the wrappers behave differently per platform. We can use the fallback in the iOS fix.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/language/impl/LanguageContextTests.kt
```diff
@@ -0,0 +1,29 @@
...
+        val languageContext = LanguageContext(propertiesModelStore)
+
+        languageContext.language = "de"
+
+        propertiesModelStore.model.language shouldBe "de"
+        languageContext.language shouldBe "de"
+    }
+
+    test("setting an empty language clears the override and uses the device language") {
+        val propertiesModelStore =
```
sherwinski, line 19: The test name says "uses the device language", but the test only checks that the model value is null. Add one assertion on the getter, for example:

```kotlin
languageContext.language shouldBe LanguageProviderDevice().language
```

Without it, a regression in the `?:` fallback passes this suite.

examples/demo/app/src/main/java/com/onesignal/example/ui/main/MainViewModel.kt
```diff
@@ -329,6 +333,13 @@ class MainViewModel(application: Application) : AndroidViewModel(application), I
         DemoLog.i(if (enabled) "Identity verification enabled" else "Identity verification disabled")
     }
 
+    fun setLanguage(language: String) {
```
sherwinski, line 337: The demo caches the language choice in SharedPreferences, separate from the SDK. On logout, `UserSwitcher` creates a fresh `PropertiesModel`, so the SDK drops the override. The dropdown still shows the cached value, so the UI and the SDK disagree.

Two options:
- Call `repository.setLanguage(language)` again in the logout path.
- Or clear the cached value on logout so the dropdown returns to "Device Default".

Demo-only, so not blocking.

sherwinski (commented):
Two small items below. Neither blocks the merge.

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/language/impl/LanguageContext.kt
fadi-george: Updated empty-string handling to store the resolved device language so it propagates to the server. I also updated the KDoc.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/core/internal/language/impl/LanguageContextTests.kt
fadi-george: Added the getter assertion and updated the test name to match the stored behavior.

examples/demo/app/src/main/java/com/onesignal/example/ui/main/MainViewModel.kt
fadi-george: Reset the language through setLanguage("") after logout so both the SDK and dropdown return to device default.

sherwinski (approved):
LGTM

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
```diff
@@ -40,10 +40,11 @@ interface IUserManager {
     val externalId: String
 
     /**
-     * Set the 2-character language either as a detected language or explicitly set for this user. See
-     * See [Supported Languages | OneSignal](https://documentation.onesignal.com/docs/language-localization#what-languages-are-supported)
+     * Sets the language for this user.
      *
-     * @param value The 2-character language string, or an empty string to use the device default.
+     * See [Supported Languages | OneSignal](https://documentation.onesignal.com/docs/language-localization#what-languages-are-supported).
```
nan-li: nit, this looks like a 404, since we're touching this, we can update to https://documentation.onesignal.com/docs/en/multi-language-messaging#supported-languages

nan-li (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/IUserManager.kt
fadi-george: Updated the KDoc to use the current supported-languages URL.

PR #2761, opened 2026-09-22
fix: [SDK-5328] reject blank identity strings

abdulraqeeb33:
# Description
## One Line Summary
Reject null and empty identity strings through one shared helper.

### Motivation
Fixes SDK-5328.

`login("")` was stored as an external id. The same blank-string problem exists on app id, language, alias labels and ids, tag keys, trigger keys, and custom event names. Null and empty are rejected together, matching the other client SDKs under SDK-5327.

### Scope
`rejectNullOrEmpty` in `com.onesignal.common` is used by login, initialize, setLanguage, aliases, tags, triggers, and trackEvent.

Tag values and trigger values may still be an empty string. A null tag or trigger value is rejected. Email and SMS still go through the existing validators, which already reject empty. Whitespace is still allowed.

Kotlin `String` parameters still throw at the Java null boundary before the method body. This change rejects empty strings on those APIs, and null where the parameter is already nullable (`initWithContextSuspend` app id).

# Testing
## Unit testing
`OneSignalImpTests` covers login("") after init (no user switch) and before init (still throws). `UserManagerTests` still pass. Core and in-app messages compile, and spotless is clean.

## Manual testing
Not run. This is a guard before native storage, covered by unit tests.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [x] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/UserManager.kt
```diff
@@ -54,6 +57,7 @@ internal open class UserManager(
         get() = _propertiesModelStore.model
 
     override fun setLanguage(value: String) {
+        if (rejectNullOrEmpty(value, "setLanguage: language")) return
```
fadi-george, line 60: This is a breaking change

abdulraqeeb33: Good catch @fadi-george . Resolved, removed it.

PR #2764, opened 2026-09-23
fix: [SDK-4995] always run suspendifyWithCompletion onComplete

fadi-george:
# Description
## One Line Summary
`suspendifyWithCompletion` always runs `onComplete`, including when the block throws or the job is cancelled before it starts.

### Motivation
Split out of #2712. Callers that release a resource in `onComplete` (for example a `PendingResult.finish()`) could leak it when the block threw, because `onComplete` only ran on success.

### Scope
- `onComplete` is registered with `invokeOnCompletion`, so it runs on success, failure, and pre-start cancellation.
- Exceptions thrown by `onComplete` are logged instead of escaping.
- No production caller passes `onComplete` today, so there is no behavior change for existing SDK paths.

# Testing
## Unit testing
Added a test that `onComplete` runs when the block throws.

## Manual testing
Not tested on device; this is a threading helper with no production caller of the changed parameter.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


abdulraqeeb33 (approved):

OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/ThreadUtils.kt
```diff
@@ -119,12 +119,20 @@ fun suspendifyWithCompletion(
...
+    val job =
+        launch {
+            try {
+                block()
+            } catch (e: Exception) {
+                Logging.error("Exception in suspendifyWithCompletion", e)
+            }
+        }
+    // invokeOnCompletion also fires when the job is cancelled before its body starts.
+    job.invokeOnCompletion {
```
abdulraqeeb33: `invokeOnCompletion` runs on the completing thread, or inline on the caller if the job is already finished. `suspendifyOnIO`'s KDoc still says `onComplete` runs on the background thread.

OneSignalSDK/onesignal/core/src/test/java/com/onesignal/common/threading/ThreadUtilsTests.kt
```diff
@@ -139,6 +139,25 @@ class ThreadUtilsTests : FunSpec({
...
+            block = {
+                throw RuntimeException("Test error")
+            },
+            onComplete = {
+                onCompleteCalled = true
+                latch.countDown()
+            },
+        )
+
+        latch.await()
```
abdulraqeeb33: `latch.await()` has no timeout, so a regression hangs the suite. This also does not cover cancel-before-start, which is the case the change is for.

abdulraqeeb33 (commented):
Two nits.

PR #2766, opened 2026-09-23
refactor: [SDK-4995] bootstrap dispatchers off the caller thread

fadi-george:
# Description
## One Line Summary
Build dispatcher executors on a background bootstrap thread so the first `launchOn*` call never constructs a pool on the caller (often main) thread.

### Motivation
Split out of #2712. `prewarm()` is best-effort: a cold-start entry point that dispatches right after it can still win the lazy-init race and pay executor construction on the main thread (SDK-4995 Play Vitals data showed no improvement after the 5.9.5 prewarm).

### Scope
- Each lane (IO, Default, SerialIO) is a `GateDispatcher` that queues work while cold and requests bootstrap from a single `BootstrapCoordinator` thread.
- Once the executor is built and core threads are prestarted, queued work drains in order and later dispatches go straight to the executor.
- If primary construction fails, the lane falls back to `Dispatchers.IO` / `Default` / `IO.limitedParallelism(1)`. If fallback also fails, queued jobs are cancelled and later lanes are not wedged.
- Rejected or cancelled work is completed on `Dispatchers.IO`, never inline on the caller thread.
- SerialIO keeps an unbounded queue, matching the previous `newSingleThreadExecutor`.
- `prewarm()` now just requests warmup for every lane. No public API changes.

# Testing
## Unit testing
- First launch returns while its lane is still being created, and creation happens off the caller thread.
- A cold queued launch completes when its generation is reset.
- Primary and fallback bootstrap failure does not wedge later lanes.

## Manual testing
Not tested on device separately; covered by the full core and notifications unit suites.

# Affected code checklist
   - [ ] Notifications
      - [ ] Display
      - [ ] Open
      - [ ] Push Processing
      - [ ] Confirm Deliveries
   - [ ] Outcomes
   - [ ] Sessions
   - [ ] In-App Messaging
   - [ ] REST API requests
   - [ ] Public API changes

# Checklist
## Overview
   - [x] PR does one thing

## Testing

## Final pass


OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/threading/OneSignalDispatchers.kt
```diff
@@ -63,129 +74,337 @@ object OneSignalDispatchers {
...
-                } catch (e: Exception) {
-                    Logging.error("OneSignalDispatchers: Failed to create SerialIO executor: ${e.message}")
-                    throw e
+                    createFallbackTarget(lane)
+                } catch (t: Throwable) {
+                    Logging.warn("OneSignalDispatchers: Fallback failed for $lane lane: ${t.message}", t)
+                    failPending("OneSignal $lane dispatcher fallback failed")
+                    null
                 }
+            } catch (t: Throwable) {
```
abdulraqeeb33: `prestartAllCoreThreads` (line 365) throws `OutOfMemoryError` after some core threads exist. That misses `catch (Exception)` and lands here, which cancels queued work and never calls `createFallbackTarget`. Shut the partial executor down in a `finally`, then fall back. Only `failPending` if the fallback also fails. `failPending` also sets the lane back to `COLD`, which reopens a `CLOSED` generation during `resetForTest`.

abdulraqeeb33 (changes requested):
Requesting changes. CI is still red (`NotificationGenerationProcessorTests > processNotificationData should not display notification when external callback indicates not to`).

`prestartAllCoreThreads` throwing `Error` skips the `Dispatchers` fallback and leaks the half-built executor. That is the failure the fallback is for.

